PR #25, opened 2014-08-20
Bugfix: flush cache if not using external cache or if filter is modified

johanee:
Currently it will ONLY flush caches if using an external cache and if the filter is not modified, the inverse of what we want.

### DIFF
--- a/classes/ActionScheduler_QueueRunner.php
+++ b/classes/ActionScheduler_QueueRunner.php
@@ -122,7 +122,7 @@ class ActionScheduler_QueueRunner {
 	 * @return void
 	 */
 	protected function clear_caches() {
-		if ( wp_using_ext_object_cache() && !apply_filters( 'action_scheduler_queue_runner_flush_cache', false ) ) {
+		if ( ! wp_using_ext_object_cache() || apply_filters( 'action_scheduler_queue_runner_flush_cache', false ) ) {
 			wp_cache_flush();
 		}
 	}


thenbrent:
@johanee thank you for this. We (Prospress) are now maintaining the repository and I only just saw this PR. You are right and this is a great patch. Thanks!

javorszky:
@jbma, this is probably the fix that affects caching on wp rocket. You could potentially use the `action_scheduler_queue_runner_flush_cache` filter to change whether completely flushed the cache or not, but be aware that it might affect other functionality

jbma:
@javorszky unfortunately not. wp_cache_flush clear the WP object cache, the issue we have is it clear our disk cache

javorszky:
Hm... so a hit to `wp-cron.php` clears all the disk cache you're setting?

jbma:
Yep. We are hooked into most of the content update to clear the cache when there is an update of a user, a post, etc. 
So I guess some operations you are doing into the cron implies content update.

javorszky:
Right. If you read the readme, every minute the `scheduled-action` post type will be involved. Some will be trashed, some will have new comments put on them. If any needs to be run, a claim is placed, which is a raw `UPDATE` query.

I would think that clearing the entire page cache when something updates one thing is unnecessary.

jbma:
OK got it :) 
Doest the cron directly update the user?

javorszky:
Depends. Usually it's like this: cron -> action scheduler -> process on a subscription -> update user if needed (change / update roles)

jbma:
Going to try to disable purge cache when editing a user to check if this solve the issue.
I'll let you know, thank's :)

jbma:
Here is our hooks to fire a purge, would you see something else than update user which could be triggered by your action scheduler?

````
add_action( 'switch_theme'				,     'rocket_clean_domain' );	// When user change theme
add_action( 'user_register'				,     'rocket_clean_domain' );	// When a user is added
add_action( 'profile_update'			,     'rocket_clean_domain' );	// When a user is updated
add_action( 'deleted_user'				,     'rocket_clean_domain' );	// When a user is deleted
add_action( 'wp_update_nav_menu'		,     'rocket_clean_domain' );	// When a custom menu is update
add_action( 'update_option_sidebars_widgets', 'rocket_clean_domain' );	// When you change the order of widgets
add_action( 'update_option_category_base',    'rocket_clean_domain' );	// When category permalink prefix is update
add_action( 'update_option_tag_base'	,     'rocket_clean_domain' ); 	// When tag permalink prefix is update
add_action( 'permalink_structure_changed',    'rocket_clean_domain' ); 	// When permalink structure is update
add_action( 'create_term'				,     'rocket_clean_domain' ); 	// When a term is created
add_action( 'edited_terms'				,     'rocket_clean_domain' );	// When a term is updated
add_action( 'delete_term'				,     'rocket_clean_domain' );	// When a term is deleted
add_action( 'add_link'					,     'rocket_clean_domain' );	// When a link is added
add_action( 'edit_link'					,     'rocket_clean_domain' );	// When a link is updated
add_action( 'delete_link'				,     'rocket_clean_domain' );	// When a link is deleted
add_action( 'customize_save'			,     'rocket_clean_domain' );	// When customizer is saved
add_action( 'avada_clear_dynamic_css_cache',  'rocket_clean_domain' );	// When Avada theme purge its own cache
add_action( 'update_option_theme_mods_' . get_option( 'stylesheet' ), 'rocket_clean_domain' ); // When location of a menu is updated
```

PR #32, opened 2015-04-23
fix fatal error process_action()

toddlahman:
Fatal error: Call to protected method
ActionScheduler_QueueRunner::process_action() from context. Changing
visibility to public allows action link to be used from a CPT dashboard.

### DIFF
--- a/classes/ActionScheduler_QueueRunner.php
+++ b/classes/ActionScheduler_QueueRunner.php
@@ -89,7 +89,7 @@ class ActionScheduler_QueueRunner {
 		return $processed_actions;
 	}
 
-	protected function process_action( $action_id ) {
+	public function process_action( $action_id ) {
 		try {
 			do_action( 'action_scheduler_before_execute', $action_id );
 			$action = $this->store->fetch_action( $action_id );
@@ -136,4 +136,4 @@ class ActionScheduler_QueueRunner {
 		return $schedules;
 	}
 }
- 
+


jconroy:
Hi @toddlahman is this a separate issue to the one @thenbrent addressed here https://github.com/Prospress/action-scheduler/commit/1d5b589759c2dd47477b4f7a7a641ebccdca7b46 earlier today (just checking quickly because I saw them both pop up in my stream for the same function) ?

toddlahman:
@jconroy Yes, this is a different issue. If ```process_action()``` runs in the background set to private visibilty, there is no issue, but if the ```process_action()``` method is called from an admin dashboard link to manually trigger the action, there is a fatal error unless ```process_action()``` is set to public visibility.

jconroy:
:+1: thanks

thenbrent:
@toddlahman thanks for the patch. Are you running this somewhere?

Subscriptions is actually using this branch atm: https://github.com/thenbrent/action-scheduler/tree/bleeding_branch which has the correct visibility on that method: https://github.com/thenbrent/action-scheduler/blob/bleeding_branch/classes/ActionScheduler_QueueRunner.php#L92

toddlahman:
@thenbrent I have been building a large project over the past few months, and recently started using the Action Scheduler as a stepping stone to the Amazon Simple Queue Service. It remains unclear at this point if Amazon SQS will be required to improve scalability.

Issue #35, opened 2015-08-17
Fatal error 


PrafullaKumarSahu:
Fatal error: Call to undefined function wc_next_scheduled_action() in  ....classes\class-wc-subscriptions-manager.php on line 2256.I am getting this error can you please give me some idea, how I can fix it.
```php
	add_action( 'plugins_loaded', 'action_scheduler_register_1_dot_2_dev', 0, 0 );

	function action_scheduler_register_1_dot_2_dev() {
		echo "test";
		exit;
		$versions = ActionScheduler_Versions::instance();
		$versions->register( '1.2-dev', 'action_scheduler_initialize_1_dot_2_dev' );
	}
```

This is where the problem is exist .

PrafullaKumarSahu:
In Class ActionScheduler_Versions changed 
```php
	public function register( $version_string, $initialization_callback ) {
		if ( isset($this->versions[$version_string]) ) {
			return FALSE;
		}
		$this->versions[$version_string] = $initialization_callback;
		return TRUE;
	}
```
   to

```php
	public function register( $version_string, $initialization_callback ) {
		if ( isset($this->versions[$version_string]) ) {
			return FALSE;
		}
		$this->versions[$version_string] = call_user_func ( $initialization_callback );
		return TRUE;
	}
```

thenbrent:
Hi @PrafullaKumarSahu please use the [`bleeding_branch`](https://github.com/Prospress/action-scheduler/tree/bleeding_branch) as that `action_scheduler_register_1_dot_4_dev()` is using version 1.4. 

It also looks like you have extra code in `action_scheduler_register_1_dot_2_dev()` which is causing the function to `exit` before the version of the library is registered. That will cause this issue.

PrafullaKumarSahu:
Hi @thenbrent  thank for your  reply , actually that was causing because my plugin was loaded with plugins_loaded hook,now I am using admin_init hook as woocommerce-subscriptions uses it.This plugin what I have developed is same as woocommerce-subscriptions having a few tweaks to know the woocommerce better .but I have noticed using call_user_function on that also not causing any problem,It is better to use call_user_function or there is some problem with it, could you please explain it.I will be thankful.

PR #37, opened 2015-08-18
Used call_user_func function to call action_scheduler_initialize_1_do…

PrafullaKumarSahu:
…t_2_dev function

It was unable to call the action_scheduler_initialize_1_dot_2_dev
function , the error was undefined function
action_scheduler_initialize_1_dot_2_dev in WC_Subscriptions_Manager
class of  woocommerce-subscriptions plugin

### DIFF
--- a/classes/ActionScheduler_Versions.php
+++ b/classes/ActionScheduler_Versions.php
@@ -15,7 +15,7 @@ class ActionScheduler_Versions {
 		if ( isset($this->versions[$version_string]) ) {
 			return FALSE;
 		}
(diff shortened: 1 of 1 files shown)


thenbrent:
I'm closing this as Subscriptions doesn't use v1.2-dev anymore, it uses v1.4-dev, so if you have v1.2-dev running, your version of Subscriptions is very out-of-date and needs to be updated. I'm also not sure this is the issue given the code pasted on #35 which I addressed here: https://github.com/Prospress/action-scheduler/issues/35#issuecomment-132780049

PR #40, opened 2015-08-30
Fix Fatal error: Call to undefined function wc_next_scheduled_action()

toddlahman:
On some systems dirname() is not creating the correct path. Using
plugin_dir_path() does find the correct path.


toddlahman:
This didn't fix the issue while testing with 5.3.29.

PR #45, opened 2015-12-12
Making sure new scheduled actions won't hog uniqid

### LINKED ISSUE #44: `wp_unique_post_slug()` overhead on large sites
Keeping this short and sweet, more as a note / for documentation.

`wp_unique_post_slug()` is called on some action scheduler related wp-cron.php calls when actions are marked complete etc.

On sites with a large number of items in the posts table this adds a substantial overhead  e.g.  

new relic trace

Looks like its bit of a known bottleneck for large sites http://wordpress.stackexchange.com/questions/119738/can-we-have-a-post-without-a-slug

As the actions aren't really public, calculating a slug/name is not really necessary.

Unfortunately there is no easy way to get around it as it is used within `wp_update_post()` e.g. https://github.com/Prospress/action-scheduler/blob/master/classes/ActionScheduler_wpPostStore.php#L563 and the only filters within in are at [the end](https://github.com/WordPress/WordPress/blob/5031ca5c23f68b20fd63fd8370c05e9b0e1af164/wp-includes/post.php#L3675)

An action or filter at the beginning around [here](https://github.com/WordPress/WordPress/blob/5031ca5c23f68b20fd63fd8370c05e9b0e1af164/wp-includes/post.php#L3553) would potentially allow us to hijack and ignore what happens after.

[This trak ticket](https://core.trac.wordpress.org/ticket/21112) may or may not be useful/interesting. Its certainly related.

For more discussion see https://github.com/Prospress/woocommerce-subscriptions/issues/1355

javorszky:
Fixes #44

Partially. Here's what's happening:

* when creating an action, it's set to draft or pending, and those don't need post_name, so they will be empty without this fix (see [here](https://github.com/WordPress/WordPress/blob/master/wp-includes/post.php#L3022-L3036))
* when updating them to publish status, they will need a valid post name, and because it's empty, it will sanitize the post_title, and roll it through the unique function
* because we're using the post_title to distinguish between actions, the first post_name is always going to be the same, so with every new action of the same type, the unique function needs to do n+1 checks
* setting the post name to something sort of random (this fix) eliminates that need to roll for unique

Part 2 of the fix:

Existing sites would need to run a similar query:

```sql
UPDATE wp_posts SET post_name = CONCAT_WS( '_', post_title, REPLACE( RAND(), '.', '-' ), ID ) WHERE post_type = 'scheduled-action' AND post_name = '';
```

When I tried this fix earlier, it didn't work, because AS was trying to update already existing empty post_named actions.

### DIFF
--- a/classes/ActionScheduler_wpPostStore.php
+++ b/classes/ActionScheduler_wpPostStore.php
@@ -29,6 +29,7 @@ class ActionScheduler_wpPostStore extends ActionScheduler_Store {
 		$post = array(
 			'post_type' => self::POST_TYPE,
 			'post_title' => $action->get_hook(),
+			'post_name' => uniqid( $action->get_hook() . '-', true ) . '-' . wp_generate_password( 34, false, false ),
 			'post_content' => json_encode($action->get_args()),
 			'post_status' => ( $action->is_finished() ? 'publish' : 'pending' ),
 			'post_date_gmt' => $this->get_timestamp($action, $date),
@@ -581,4 +582,4 @@ class ActionScheduler_wpPostStore extends ActionScheduler_Store {
 		$taxonomy_registrar->register();
 	}
 }
- 
+


javorszky:
Another interesting thing though is that if the site has another plugin that's using Action Scheduler as well, and is initialised earlier, even if it's not active but declares the class, Subs will end up using that plugin's copy of AS. So all other plugins will need to update themselves. Therefore, ping @franticpsyx :)

75nineteen:
@javorszky - thanks. We have a fix release planned for this week so we will get this added in.

75nineteen:
@javorszky - which branch should we use to update to? Should we be concerned your commit failed?

toddlahman:
It looks like this pull request will be merged into the master branch. Perhaps it might be time to create a new release, since this is a significant update, and add instructions to run the update query into the readme.

javorszky:
@75nineteen probably `master`, once this changeset gets merged. It needs to go through code review, and I'd like to have either @jconroy or @thenbrent give a thumbs up or down.

@toddlahman once there's a decision whether this will be in or not, I suspect there'll be a new version too :)

javorszky:
Heads up, this might not be sufficient change, investigating further...

75nineteen:
K. We are at least coming up to date with what you guys are running on 2.0.6 to start.

toddlahman:
@javorszky Take your time. Would be happy to test the almost final version locally when you think it's ready for the spotlight.

javorszky:
The problem is not with Action Scheduler, but WordPress core. If we pass in the `post_name`, it gets populated, but only if there's an active user present. If WordPress is doing it on its own through ajax, [this bit of code](https://github.com/WordPress/WordPress/blob/master/wp-includes/post.php#L3013-L3016) will be hit, because it's a new `pending`, and there's no authenticated user.

There's a trac ticket that could help us short circuit it: https://core.trac.wordpress.org/ticket/21112

Alternatively we can have a standalone action scheduler plugin loaded early (so directory name like `__aaa` so it comes first in alphabetical sorting), and then that will be used instead of the individual plugins' versions. This should be used for now I think for high traffic sites.

For this reason I'm closing this, as there's not much we can do with WP from the AS side.

toddlahman:
Yes, I've seen that ticket, and hope with all the new efforts at making WordPress more scalable, considering these use cases, a remedy will land in core at some point. Really appreciate all the investigative work on these issues by everyone at Prospress.

toddlahman:
@javorszky Have you explored creating a custom table, and custom SQL queries, to bypass the post limitations?

javorszky:
Yes, briefly :laughing: on 11th

The major issue with that is migrating existing schedules from the `wp_posts` table and the `comments` table. It shouldn't be too hard (famous last words), as the code that deals with storage is in 4 files, and is extending `ActionScheduler_Store`. Tracking on https://github.com/Prospress/action-scheduler/issues/46

PR #47, opened 2016-01-19
Merging Action Scheduler changes inside Subs

javorszky:
I'm not entirely sure why there are changes in Subs' included action scheduler version that's not here, but that's the case. This changeset contains all the changes between this repo's master (https://github.com/Prospress/action-scheduler/commit/bdee7d16d4830f4f5fc53023580585917e08f7a9) and Subs 2's master (https://github.com/Prospress/woocommerce-subscriptions/commit/2457cf61f51ec82cf20e36122141a56830e1c48f).

One reason to do this is I was trying to implement the changes for this issue: https://github.com/Prospress/action-scheduler/issues/38, but couldn't even find the file I would have needed to poke.


javorszky:
Also I'm unfamiliar with travis, but the `push` job runs on a legacy platform, and is failing on choosing a php version.

toddlahman:
Thank you for the update @javorszky :)

thenbrent:
> I'm not entirely sure why there are changes in Subs' included action scheduler version that's not here

Those changes are here, they're not `master`, they're in: https://github.com/Prospress/action-scheduler/tree/bleeding_branch 

> This changeset contains all the changes between this repo's master

This isn't the merge we want.

Part of the reason Subscriptions uses the `bleeding_branch` is because `master` has diverged. The `bleeding_branch` includes the important Admin UI while the `master` branch includes an improvement to the way actions are claimed/locked.

We want to take advantage of both of these improvements, but the new claiming system needs extremely thorough testing before it can be released into the wild (otherwise we may end up with duplicate payments in Subscriptions) which is why we still ship Subscriptions with the `bleeding_branch`, because it hasn't been thoroughly tested yet.

I did hope to merge this as part of Subscriptions v2.0 dev cycle, but after it went way over schedule, it was tentatively pushed back (most likely to be part of Subscriptions v2.1 dev cycle).

A merge closer to what we want can be found here: https://github.com/Prospress/action-scheduler/tree/bleeding_merge_master

That merges the new claiming system from `master` with the admin UI and other minor tweaks in `bleeding_branch`.

Again though, that branch isn't something we'll push out in a minor release though because it needs the thorough testing of a major release (which is why there is no PR for it yet). It's something that will likely be merged just before v2.1 hits beta, and then tested for weeks before it is released.

> I'm not entirely sure why

It doesn't hurt to ask before doing up a PR. Especially in the case of Action Scheduler, there's likely an undocumented reason that I can explain (there are a few gotchas like this still lingering that I've been wanting to get fixed up for over a year). :smile: 

I thought there was an issue for this already, but clearly not so I've just created one: https://github.com/Prospress/action-scheduler/issues/48

I'll close this PR and we can discuss the issue further if needed on https://github.com/Prospress/action-scheduler/issues/48

At least now, it's documented for others. :smile:

PR #60, opened 2016-10-03
Fix cron schedule's next date fetch method

javorszky:
No issue. The problem is that the third argument permits the cron schedule to return the same date as the "next" date, which results in duplicate scheduled actions. Take the following cron expression:

```
20-26/2 13 * * * *
```

Without this fix, the schedules will be the following:

```
3rd Oct 13:20
3rd Oct 13:20
3rd Oct 13:22
3rd Oct 13:22
3rd Oct 13:24
3rd Oct 13:24
3rd Oct 13:26
3rd Oct 13:26
4th Oct 13:20
```

With this fix, these are the schedules:

```
3rd Oct 13:20
3rd Oct 13:22
3rd Oct 13:24
3rd Oct 13:26
4th Oct 13:20
```

Current implementation would lead to bugs and race conditions and parallel execution.

Related to https://github.com/Prospress/woocommerce-subscriptions/pull/1690

### DIFF
--- a/classes/ActionScheduler_CronSchedule.php
+++ b/classes/ActionScheduler_CronSchedule.php
@@ -21,7 +21,7 @@ class ActionScheduler_CronSchedule implements ActionScheduler_Schedule {
 	 */
 	public function next( DateTime $after = NULL ) {
 		$after = empty($after) ? clone $this->start : clone $after;
-		return $this->cron->getNextRunDate($after, 0, TRUE);
+		return $this->cron->getNextRunDate($after, 0, false);
 	}
 
 
@@ -41,4 +41,4 @@ class ActionScheduler_CronSchedule implements ActionScheduler_Schedule {
 		$this->start = as_get_datetime_object($this->start_timestamp);
 	}
 }
- 
+


thenbrent:
> The problem is that the third argument permits the cron schedule to return the same date as the "next" date

I've reproduced this now. It's not so much that third argument permits the cron schedule to return the same date as the "next" date as that it permits the _same_ time as the time its run, which if AS is set to run more frequently than WP-Cron would allow (perhaps by using Linux cron job) would could be the same time as the time the schedule is run. So it's not actually duplicating the action, it's creating a new action for the same time as that's the "next" time that action should be run.

This appears to have been [by design](https://github.com/Prospress/action-scheduler/commit/30652837d650f1110728afdd51eb2ce27019f624#diff-d8eeec0c5c9e98b886c1bfb4a2a23680R23), which looking at the code may have been to better support recurring actions (which you may want running every minute or more). But I can definitely see why it can lead to a pretty bad gotcha so I'll merge this.

If we get complaints about recurring actions not running regularly enough, I'll review alternative solutions, like creating a new API function.

thenbrent:
Oh @javorszky this probably also won't go out in Subscriptions until v2.1, which will probably be the next version of Subscriptions we release (i.e. no more `2.0.n` versions). That's due for release early November.

javorszky:
:+1: thanks for the update.

In the meantime, I'll redo the cron class and use that instead.

PR #64, opened 2016-10-07
Cache comment counts in our own transient

### LINKED ISSUE #30: Action Scheduler comment filtering never caches
According to query debugging, the caching implemented for comments-0 never kicks in.

* https://github.com/Prospress/action-scheduler/blob/master/classes/ActionScheduler_wpCommentLogger.php#L123

This may be due to non-strict comparison of the cached value. I'll provide feedback if/when I look into this further.

thenbrent:
Rather than using object cache, which doesn't work.

Similar to https://github.com/woocommerce/woocommerce/pull/11449

Fixes #30

### DIFF
--- a/classes/ActionScheduler_wpCommentLogger.php
+++ b/classes/ActionScheduler_wpCommentLogger.php
@@ -126,11 +126,24 @@ class ActionScheduler_wpCommentLogger extends ActionScheduler_Logger {
 		global $wpdb;
 
 		if ( 0 === $post_id ) {
+			$stats = $this->get_comment_count();
+		}
 
-			$count = wp_cache_get( 'comments-0', 'counts' );
-			if ( false !== $count ) {
-				return $count;
-			}
+		return $stats;
+	}
+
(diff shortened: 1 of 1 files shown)


james-allan:
This goes beyond the initial issue this PR set out to address but something I noticed while reviewing this, that we might be able to sneak into this PR, is that the **All** count includes some comments. eg

screen shot 2016-10-11 at 12 20 40 pm

Looking at the WC equivalent function , `ActionScheduler_wpCommentLogger::get_comment_count()` is missing this line [(class-wc-comments.php#L292)](https://github.com/woocommerce/woocommerce/blob/dec792b731dbbea6d34aa3ddcd6a3bf05bf43077/includes/class-wc-comments.php#L292).

Other than that this LGTM 👍

thenbrent:
@james-allan nice catch! I've made sure `$stats['all']` is set now and rebased it into the existing commit. 👍

james-allan:
👍

PR #82, opened 2017-06-26
update cron expression lib, and use composer

jbrinley:
Since we have dropped support for PHP 5.2, we can use the
official lib from mtdowling and use Composer's autoloader.
Eliminates the necessity to maintain our own 5.2-compatible
back-port of cron expression.


thenbrent:
LGTM. 👍

PR #87, opened 2017-08-07
Update Action Scheduler to Handle Custom Table Data Store

thenbrent:
We now have a working custom table implementation as a data store available here: https://github.com/Prospress/action-scheduler-custom-tables

To achieve that required additional changes to Action Scheduler core. This PR introduces those changes. It also updates some housekeeping, like PHP versions for Travis tests etc.


toddlahman:
👍

thenbrent:
I had to add a few patches to this to:

* use _Precise_ on Travis so we can continue to test against PHP 5.3
* revert the cron expression lib patch, because WC _does_ still support PHP < 5.6, it's just that it _recommends_ 5.6+, same as WP 😭 

Otherwise, LGTM. 👍

PR #88, opened 2017-08-15
Add new Admin UI that uses data stores to be independent of WP post store

### LINKED ISSUE #85: Update admin view to use data stores
When introducing the [admin view](https://github.com/Prospress/action-scheduler/blob/1.5.3/classes/ActionScheduler_AdminView.php) with SHA: cbe65a2bd, I leveraged WordPress's code for creating a list table for a custom post type. That was the quickest way to make an admin view available. However, that meant the admin view was tightly coupled to the built-in data store. One of the great design decisions of Action Scheduler was to abstract the data store.

Now that [highly scalable custom tables](https://github.com/Prospress/action-scheduler-custom-tables/) are coming to Action Scheduler, the core admin view needs to be updated to use the data store.

To this will require:

- [x] Create new [list table](https://www.sitepoint.com/using-wp_list_table-to-create-wordpress-admin-tables/) for scheduled actions which sources data from the data store/public APIs
- [x] Implement feature parity with the existing list table, including:
    - [x] Display the same columns and content
    - [x] **Filter** actions by _status_
    - [x] View log entries (the UI for this could be changed to provide a better UX, specifically by avoiding an additional page load just for a few rows of data. A JS popup opening the log in a lightbox like dialog would be better.)
    - [x] Display claim ID for actions with a claim (like `in-progress` or `completed` actions
    - [x] **Search** for scheduled actions by: hook name or arguments (like a subscription ID)
    - [x] **Sort** scheduled actions by hook name, scheduled, start or completed date, claim ID or number of log entries
    - [x] Allow actions to be _Run_ or _Trash_ed via the inline actions, e.g. https://cl.ly/0k011i0N1f3H
- [x] Move table to new tab under **WooCommerce > System Status** page rather than **Tools** menu, this wasn't previously possible with the way WC requires tabs to be setup, but it should be possible once we have our own list table. S#29  for background.

crodas:
Fixes #85 

Things that I'm struggled with, that I'm working on a way around them:

 * [x] There is no way of telling if a Finished action failed or not.
 * [x] There is no information about `Trash`ed actions


thenbrent:
@crodas can you please rebase this on the latest version of `1-6-dev`. It now includes the `custom_tables` branch as I've merged #87.

classes/ActionScheduler_AdminView.php
```diff
@@ -58,12 +65,37 @@ public function init() {
 		add_filter( 'posts_search', array( self::instance(), 'search_post_password' ), 10, 2 );
 	}
 
-	public function action_scheduler_post_type_args( $args ) {
```
thenbrent: As a public function, this needs to be deprecated, not simply deleted. It could be being used by 3rd party code, and if it's deleted, that will trigger a fatal error.

See examples in Subscriptions for how to deprecate a function, but in short:

1. move it to the end of the file
2. trigger a deprecated function notice

classes/ActionScheduler_ListTable.php
```diff
@@ -0,0 +1,354 @@
...
+	 */
+	protected $columns = array(
+		'hook'   => 'Hook',
+		'status' => 'Status',
+		'args'   => 'Arguments',
+		'group'  => 'Group',
+		'recurrence' => 'Recurrence',
+		'scheduled'  => 'Scheduled Date',
+		'claim_id'   => 'Claim ID',
+		'comments'   => 'Log',
```
thenbrent: These strings should all be made available for translation and passed through i18n functions.

classes/ActionScheduler_ListTable.php
```diff
@@ -0,0 +1,354 @@
...
+	);
+
+	protected $row_actions = array(
+		'hook' => array(
+			'run' => array( 'Run', 'Process the action now as if it were run as part of a queue' ),
```
thenbrent: These strings should be made available for translation and passed through i18n functions.

classes/ActionScheduler_ListTable.php
```diff
@@ -0,0 +1,354 @@
...
+	protected static $should_include_js = false;
+
+
+	/**
+	 * Sets the current data store object into `store->action` and initialises the object.
+	 */
+	public function __construct() {
+		self::$should_include_js = true;
+
+		$this->store = (object) array(
```
thenbrent: As this property contains two stores, its name should be `$stores` not the singular `$store` to communicate that.

Also, what is the benefit of storing them in a property instead of just calling `ActionScheduler::store()` or  `ActionScheduler::logger()`?

Calling them explicitly makes it clear which store/object is being used when it is used instead of abstracting that information, and it will always be the canonical source of a store, so it's not necessary to future proof it against changes.

classes/ActionScheduler_ListTable.php
```diff
@@ -0,0 +1,354 @@
...
+
+		parent::__construct( array(
+			'singular' => $this->translate( 'action-scheduler' ),
+			'plural'   => $this->translate( 'action-scheduler' ),
+			'ajax'     => false,
+		) );
+	}
+
+	/**
+	 * Setups the libraries. It is executed if `is_admin()` is TRUE.
```
thenbrent: Instead of `Setups`, the correct English would be `Sets up`. Also this isn't a library, it's a class. :)

classes/ActionScheduler_ListTable.php
```diff
@@ -0,0 +1,354 @@
...
+	 */
+	public static function init() {
+		add_action( 'admin_footer', __CLASS__ . '::print_script' );
+		add_action( 'admin_enqueue_scripts', __CLASS__ . '::register_javascript' );
+	}
+
+	/**
+	 * Prints our javascript, if we need it.
```
thenbrent: This docblock is redundant, `print_script()` already makes it pretty clear the method "Prints our javascript". Where does it print this JS? Why does it print it? Why is the JS required? Those are things that aren't obvious from the function name and that would be useful to document here.

classes/ActionScheduler_ListTable.php
```diff
@@ -0,0 +1,354 @@
...
+		}
+		return __( 'Non-repeating', 'action-scheduler' );
+	}
+
+	/**
+	 * {@inheritDoc}
+	 */
+	protected function get_items_query_limit() {
+		return $this->get_items_per_page( $this->package . '_items_per_page', $this->items_per_page );
+	}
```
thenbrent: Would this method be better suited to the parent class?

classes/ActionScheduler_ListTable.php
```diff
@@ -0,0 +1,354 @@
...
+	protected function get_items_query_offset() {
+		global $wpdb;
+
+		$per_page = $this->get_items_query_limit();
+		$current_page = $this->get_pagenum();
+		if ( 1 < $current_page ) {
+			return $per_page * ( $current_page - 1 );
+		}
+
+		return 0;
```
thenbrent: Would this method be better suited to the parent class?

classes/ActionScheduler_ListTable.php
```diff
@@ -0,0 +1,354 @@
...
+	 * Prints the comments, which are the log entries. It needs to be named comments otherwise it won't pickup
+	 * the WordPress styles to make the number pretty.
+	 *
+	 * It will also render all the log entries, but it will be hidden and rendered on click in the number as a modal.
+	 *
+	 * @param array $row Action array.
+	 */
+	public function column_comments( array $row ) {
+		echo '<div id="log-' . $row['ID'] . '" class="log-modal hidden" style="max-width:800px">';
+		echo '<h3>Log entries for ' . $row['ID'] . '</h3>';
```
thenbrent: These strings should be passed through i18n functions so they are available for translation.

classes/ActionScheduler_ListTable.php
```diff
@@ -0,0 +1,354 @@
...
+		foreach ( $row['comments'] as $log ) {
+			echo '<p>' . $log->get_message() . '</p>';
+		}
+		echo '</div>';
```
thenbrent: The date/time/timezone of the comment would be very helpful to include here too.

classes/ActionScheduler_ListTable.php
```diff
@@ -0,0 +1,354 @@
...
+	 * Returns if the current action is finished or pending.
+	 *
+	 * @param ActionScheduler_Action $item
+	 */
+	protected function get_status( $item ) {
+		if ( $item->is_finished() ) {
+			return 'Completed';
+		}
+
+		return 'Pending';
```
thenbrent: These status strings should be passed through i18n functions so they are available for translation.

classes/ActionScheduler_AdminView.php
```diff
@@ -28,9 +28,16 @@ public static function instance() {
...
+			if ( class_exists( 'WooCommerce' ) ) {
+				add_action( 'woocommerce_admin_status_content_action-scheduler', array( $self, 'render_admin_ui' ) );
+				add_filter( 'woocommerce_admin_status_tabs', array( $self, 'register_menu_woo_tab' ) );
+			} else {
+				add_action( 'admin_menu', array( $self, 'register_menu' ) );
+			}
+			ActionScheduler_ListTable::init();
 		}
 
 		self::$admin_url = admin_url( 'edit.php?post_type=' . ActionScheduler_wpPostStore::POST_TYPE );
```
thenbrent: All of the code here responsible for setting up the old admin UI should be deleted when save to do so, or at least deprecated.

classes/ActionScheduler_ListTable.php
crodas: The library (PP_Table_List) would translate any string automatically.

crodas: https://github.com/Prospress/action-scheduler/blob/issue_85/vendor/pp-list-table/src/class-pp-list-table.php#L97-L99

Should I make the translation implicit in the code?

classes/ActionScheduler_ListTable.php
crodas: There is just a minor improvements. It's cheaper to read a property to call a method. But I'll change it.

classes/ActionScheduler_ListTable.php
crodas: True. I'll make it happen.

classes/ActionScheduler_ListTable.php
crodas: We need to override this. The parent class will return SQL code for the pagination, in our version we do not need SQL but the real number.

classes/ActionScheduler_ListTable.php
crodas: That would be ideal, but the class does not exposes it.

classes/ActionScheduler_ListTable.php
```diff
@@ -0,0 +1,385 @@
...
+	 * Returns the recurrence of an action or 'Non-repeating'. The output is human readable.
+	 *
+	 * @param ActionScheduler_Action $item
+	 *
+	 * @return string
+	 */
+	protected function get_recurrence( $item ) {
+		$recurrence = $item->get_schedule();
+		if ( method_exists( $recurrence, 'interval_in_seconds' ) ) {
+			return self::human_interval( $recurrence->interval_in_seconds() );
```
thenbrent: Where is `self::human_interval()` defined? I don't see it in `PP_List_Table` or `ActionScheduler_ListTable`.

classes/ActionScheduler_ListTable.php
```diff
@@ -0,0 +1,385 @@
...
+	 * Returns if the current action is finished or pending.
+	 *
+	 * @param ActionScheduler_Action $item
+	 */
+	protected function get_status( $item ) {
+		if ( $item->is_finished() ) {
+			return __( 'Completed', 'action-scheduler' );
+		}
+
+		return __( 'Pending', 'action-scheduler' );
```
thenbrent: Instead of creating this function, you should use `ActionScheduler::store()->get_status( $action_id )` above.

classes/ActionScheduler_ListTable.php
```diff
@@ -0,0 +1,385 @@
...
+		$this->items = array();
+
+		$total_items = $this->stores->action->query_actions_count( $query );
+
+		foreach ( $this->stores->action->query_actions( $query ) as $id ) {
+			$item = $this->stores->action->fetch_action( $id );
+			$this->items[ $id ] = array(
+				'ID'     => $id,
+				'hook'   => $item->get_hook(),
+				'status' => $this->get_status( $item ),
```
thenbrent: You might want to use `ActionScheduler::store()->get_status( $action_id )` here.

classes/ActionScheduler_ListTable.php
```diff
@@ -0,0 +1,385 @@
...
+	 * We filter by status because if we do not filter we have NullActions that are not rendered at the 
+	 * moment.
+	 *
+	 * @return string
+	 */
+	public function get_request_status() {
+		$statuses = array(
+			ActionScheduler_Store::STATUS_PENDING,
+			ActionScheduler_Store::STATUS_COMPLETE,
+			ActionScheduler_Store::STATUS_FAILED,
```
thenbrent: What about `ActionScheduler_Store::STATUS_CANCELED` and `ActionScheduler_Store::STATUS_RUNNING`?

classes/ActionScheduler_ListTable.php
```diff
@@ -0,0 +1,385 @@
...
+	/**
+	 * Prints the comments, which are the log entries. It needs to be named comments otherwise it won't pickup
+	 * the WordPress styles to make the number pretty.
+	 *
+	 * It will also render all the log entries, but it will be hidden and rendered on click in the number as a modal.
+	 *
+	 * @param array $row Action array.
+	 */
+	public function column_comments( array $row ) {
+		echo '<div id="log-' . $row['ID'] . '" class="log-modal hidden" style="max-width:800px">';
```
thenbrent: The `$row['ID']` should be escaped before being printed here (or forced to be `absint()`).

classes/ActionScheduler_ListTable.php
```diff
@@ -0,0 +1,385 @@
...
+		echo '<h3>' . esc_html( sprintf( __( 'Log entries for %d', 'action-scheduler' ),  $row['ID'] ) ) . '</h3>';
+		$timezone = new DateTimezone( 'UTC' );
+		foreach ( $row['comments'] as $log ) {
+			$date = $log->get_date();
+			$date->setTimezone( $timezone );
+			echo '<p><strong>' . esc_html( $date->format( 'Y-m-d H:i:s' ) ) . ' UTC</strong> ' . esc_html( $log->get_message() ) . '</p>';
```
thenbrent: Why not use `tz` or `tzcorrection` here instead of hardcoding `UTC`?

classes/ActionScheduler_ListTable.php
```diff
@@ -0,0 +1,385 @@
...
+		}
+		echo '</div>';
+
+		echo '<div class="post-com-count-wrapper">';
+		echo '<a href="#" data-log-id="' . $row['ID'] . '" class="post-com-count post-com-count-approved log-modal-open">';
```
thenbrent: `$row['ID']` should be escaped before being printed here (or forced to be `absint()`).

classes/ActionScheduler_ListTable.php
```diff
@@ -0,0 +1,385 @@
...
+	}
+
+	/**
+	 * Prints the scheduled date in a human friendly format.
+	 *
+	 * @param array $row The array representation of the current row of the table
+	 */
+	public function column_scheduled( $row ) {
+		$next_timestamp = $row['scheduled']->next()->format( 'U' );
+		echo get_date_from_gmt( gmdate( 'Y-m-d H:i:s', $next_timestamp ), 'Y-m-d H:i:s' );
```
thenbrent: Why use `get_date_from_gmt()` and `gmdate()` here instead of just `$row['scheduled']->next()->format( 'Y-m-d H:i:s' )`?

The value should probably also be escaped before being printed just to be safe.

classes/ActionScheduler_ListTable.php
```diff
@@ -0,0 +1,385 @@
...
+	/**
+	 * Prints the available statuses so the user can click to filter.
+	 *
+	 * @return void
+	 */
+	public function display_filter_by_status() {
+		$statuses = array(
+			'pending' => ActionScheduler_Store::STATUS_PENDING,
+			'complete' => ActionScheduler_Store::STATUS_COMPLETE,
+			'failed'  => ActionScheduler_Store::STATUS_FAILED,
```
thenbrent: What about `ActionScheduler_Store::STATUS_CANCELED` and `ActionScheduler_Store::STATUS_RUNNING`?

classes/ActionScheduler_ListTable.php
```diff
@@ -0,0 +1,385 @@
...
+			$total_items = $this->stores->action->query_actions_count( compact( 'status' ) );
+			if ( 0 === $total_items ) {
+				continue;
+			}
+
+			if ( $status === $this->get_request_status() ) {
+				$li[] =  '<li class="' . esc_attr( $name ) . '">'
+					. '<strong>'
+						. esc_html( ucfirst( $name ) )
+					. "</strong> ($total_items)"
```
thenbrent: `$total_items` needs to be escaped.

classes/ActionScheduler_ListTable.php
```diff
@@ -0,0 +1,385 @@
...
+				. '</li>';
+				continue;
+			}
+
+			$li[] =  '<li class="' . esc_attr( $name ) . '">'
+				. '<a href="' . esc_url( add_query_arg( 'status', $status ) )  . '">'
+				. esc_html( ucfirst( $name ) )
+				. "</a> ($total_items)"
```
thenbrent: `$total_items` needs to be escaped.

classes/ActionScheduler_AdminView.php
```diff
@@ -258,6 +288,7 @@ public static function row_actions( $actions, $post ) {
 	 * @return string The URL for running the action.
 	 */
 	private static function get_run_action_link( $action_id, $operation = 'process' ) {
```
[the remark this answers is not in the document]
cesar-prospress: I did not remove any private property/function because they are still referenced from some of the public (but deprecated method). Are we breaking those functions? (while keeping the function accesible?)

thenbrent:
One other important item for this PR that I just remembered - it should also update the `README` to use a screenshot of the new UI, and also accurately give the location of the new UI: https://github.com/Prospress/action-scheduler#managing-scheduled-actions

classes/ActionScheduler_ListTable.php
thenbrent: > Do you want me to fix it @thenbrent ? 

Yes. But make sure your fix is using the least possible code. Don't just add all those methods `ActionScheduler_NullAction`. That is most likely not required. Perhaps instead create a new `ActionScheduler_CancelledAction` which extends `ActionScheduler_Action` or similar.

cesar-prospress: Will do it now.

thenbrent: I should also say be sure to double and triple check all occasions where a cancelled action is instantiated to see what other implications the above suggestion will have. Changing the way cancelled actions are instantiated could have terrible consequences, like duplicating actions.

There are likely other possible solutions with lower impact than that suggest, like manually instantiating an `ActionScheduler_Action` instance for cancelled actions in this code to avoid impacting any other code. Please explore that and any other possible options you come up with too.

cesar-prospress: > There are likely other possible solutions with lower impact than that suggest, like manually instantiating an ActionScheduler_Action instance for cancelled actions in this code to avoid impacting any other code. Please explore that and any other possible options you come up with too.

This is a hack, but I agree, it may has with lower impact. However I'm think I should change https://github.com/Prospress/action-scheduler/blob/master/classes/ActionScheduler_wpPostStore.php#L96-L97 to instantiate a 'CancelledAction' object if the `post_status` is `'trash'`. It does not seem that error prone. I haven't look at the custom tables repo though.

thenbrent: > It does not seem that error prone. I haven't look at the custom tables repo though.

It's not the custom tables repo I'm worried about. It's the queue runner.

> I'm think I should change

I can't see why that wouldn't return a valid action for `ActionScheduler_Store::STATUS_CANCELED`?

I just had a look at `make_action_from_post()`, and I can't see why that would return an instance of `ActionScheduler_NullAction` either. It should return an instance of `ActionScheduler_FinishedAction`.

classes/ActionScheduler_ListTable.php
```diff
@@ -0,0 +1,385 @@
...
+			'status' => __( 'Status', 'action-scheduler' ),
+			'args'   => __( 'Arguments', 'action-scheduler' ),
+			'group'  => __( 'Group', 'action-scheduler' ),
+			'recurrence' => __( 'Recurrence', 'action-scheduler' ),
+			'scheduled'  => __( 'Scheduled Date', 'action-scheduler' ),
+			'claim_id'   => __( 'Claim ID', 'action-scheduler' ),
+			'comments'   => __( 'Log', 'action-scheduler' ),
+		);
+
+		$this->row_actions = array(
```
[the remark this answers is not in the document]
cesar-prospress: This was fixed 1c9db0a.

classes/ActionScheduler_ListTable.php
cesar-prospress: Fixed already

classes/ActionScheduler_ListTable.php
```diff
@@ -0,0 +1,385 @@
...
+	 *
+	 * @return string
+	 */
+	public function column_args( array $row ) {
+		return '<code>' . json_encode( $row['args'] ) . '</code>';
+	}
+
+	/**
+	 * Prints the comments, which are the log entries. It needs to be named comments otherwise it won't pickup
+	 * the WordPress styles to make the number pretty.
```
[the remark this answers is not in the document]
cesar-prospress: ❤️

classes/ActionScheduler_Store.php
```diff
@@ -45,6 +45,12 @@
 	abstract public function query_actions( $query = array() );
 
 	/**
+	 * @param array $query
+	 * @return array The IDs of actions matching the query
+	 */
+	abstract public function query_actions_count( $query = array() );
```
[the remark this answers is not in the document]
cesar-prospress: https://github.com/Prospress/action-scheduler-custom-tables/pull/4

cesar-prospress:
> One other important item for this PR that I just remembered - it should also update the README to use a screenshot of the new UI, and also accurately give the location of the new UI: https://github.com/Prospress/action-scheduler#managing-scheduled-actions

It is updated in 78482f6

thenbrent:
@cesar-prospress an issue found while reviewing #99 - `query_actions_count()` is duplicating too much code from `query_actions()`.

The problem with this duplication is demonstrated in the duplicate changes required to the post store: https://github.com/Prospress/action-scheduler/pull/99/commits/f33e65dd6d8e81eff3729db508f5e4f8c7a96bb7#diff-ffba07a4fcd3281ec1f31d812c761966

`query_actions()` should be refactored so that `query_actions_count()` can be implemented without duplicating any of the query code. That should be possible using a new method to build the query that accepts parameters sufficient to account for the differences between the two methods. A [quick diff](https://www.diffchecker.com/lYjmpAmr) shows that the only differences are the:

* `SELECT` value
* inclusion of an order by clause
* `$wpdb` getter used to run the query

Each of those can easily be accounted for via params on a new `get_query_actions_sql()` method or similar.

cesar-prospress:
@thenbrent that's a good feedback. `get_query_actions_sql()` would be a private/protected method though, specific to the default store.
I would have to go over and replicate the same thing on the custom-table plugin. But I insist that the `get_query_actions_sql` remains store specific, that's a store detail (and hopefully people will implement other non-sql storages).

cesar-prospress:
@thenbrent @JPry  I think 7c510f3 improves both functions. If you guys like it I will send a similar PR to https://github.com/Prospress/action-scheduler-custom-tables

thenbrent:
@crodas those patches LGTM 👍

thenbrent:
In line with this comment https://github.com/Prospress/action-scheduler/pull/98#issuecomment-360628261 I've just pushed up a bunch of commits to introduce status handling on the action itself. Specific commit range: b2fdc29...725313a.

thenbrent:
Travis is now giving this the green light, so it's ready for final review.

thenbrent:
@JPry when you get to this, something I'd really like feedback on here is how to set a stored action's ID, status and claim ID. 

Initially, @cesar-prospress attempted to solve this by adding `set_id()`, `set_status()` and `set_claim_id()`  methods.

I moved these to the new `ActionScheduler_StoredAction` class and made them `protected` in SHA: bec1879b157f76ba79319. That was to try to maintain existing behavior of `ActionScheduler_Action` and disambiguate an instance of an action that has been stored with one which has not been stored.

However, passing `$id` and `$status` as optional parameters to `ActionScheduler_StoredAction::__construct()` doesn't make sense. They should be required for instantiating a stored action. They should probably also be in the first 3 parameters, along with `$hook`. But we can't do that while extending `ActionScheduler_Action`, nor could we allow `ActionScheduler_FinishedAction` to extend `ActionScheduler_StoredAction` if we did that as it would break backward compatibility.

The only alternatives I really see is are:

* go back to having public `set_id()`, `set_status()` and `set_claim_id()`  methods, which is less than ideal as most other data for an action is set via its constructor, not public setters (e.g. `set_hook()` is `protected`, not `public`).
* deprecate the existing `ActionScheduler_Action` classes completely and introduce new class to more accurately reflect the data within (i.e. stored vs. unstored/pending storage). For example, the `ActionScheduler_Action` could be renamed `ActionScheduler_ActionPendingSave`. This would allow for more suitable parameters to be passed to each types constructors, while clearly distinguishing between an action that does have an ID and status, and those which do not. We still need to be able to instantiate unstored actions, for reasons why see `ActionScheduler_ActionFactory` and its methods, like `ActionScheduler_ActionFactory::single()`.

JPry:
@thenbrent I started looking into this code, but mostly I've started thinking about your [latest comment](https://github.com/Prospress/action-scheduler/pull/88#issuecomment-363229074). I agree that none of those options are really ideal. I have an idea for a different option that might help us simplify our constructors and maintain different classes for actions in different phases (stored, finished, unstated, etc.).

First I think we should create an `ActionInterface` that provides for all of the core functionality in a given action. Specifically, I think this should be the hook, args, and Schedule. It would also include the `execute()` method.

If we can't do an Interface for PHP version reasons, no problem. Whether we do or don't, we could have a BasicAction class that meets the above interface. Then we can have other versions of actions that take an Action object (which has the hook, args, and schedule), and then any additional data we need. This could include things like the ID, status, and claim. If needed, we could update the Factory class to create different types of actions for us, and make that the approved way to create new action objects. Using the factory in particular is where we would really have the win with not having to make the rest of the code overly complicated (I think).

What do you think? If this idea is confusing, I'd be willing to put together a proof of concept and also discuss it directly.

thenbrent:
@JPry sounds good to me. I'll follow up on Slack about proof of concept vs. discussing it.

thenbrent:
I've just triggered the following error when viewing a cancelled action. I'll fix this up tomorrow.

```
PHP Fatal error:  Uncaught Error: Call to a member function format() on null in plugins/action-scheduler/classes/ActionScheduler_ListTable.php:227
```

JPry:
@thenbrent Here's a POC: https://github.com/Prospress/action-scheduler/compare/issue_85...interface_poc

classes/ActionScheduler_wpPostStore.php
```diff
@@ -265,44 +317,87 @@ public function query_actions( $query = array() ) {
...
+					break;
+				case 'modified':
+					$orderby = 'p.post_modified';
+					break;
+				case 'date':
+				default:
+					$orderby = 'p.post_date_gmt';
+					break;
+			}
+			if ( strtoupper($query['order']) == 'ASC' ) {
```
JPry: Minor formatting adjustment needed within the `strtoupper()` call. Also, Yoda condition and `===` should be used here.

vendor/pp-list-table/src/class-pp-list-table.php
```diff
@@ -0,0 +1,511 @@
...
+	 * escaped and ready to use (with `IN`).
+	 */
+	protected $bulk_actions = array();
+
+	/**
+	 * Makes translation easier, it basically just wraps
+	 * `_x` with some default (the package name)
+	 */
+	protected function translate( $text, $context = '' ) {
+		return _x( $text, $context, $this->package );
```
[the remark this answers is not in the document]
JPry: Just mentioning that this still needs updated.

classes/ActionScheduler_ListTable.php
```diff
@@ -0,0 +1,378 @@
...
+
+	/**
+	 * Bulk delete
+	 *
+	 * Deletes actions based on their ID. This is the handler for the bulk delete. It assumes the data
+	 * properly validated by the callee and it will delete the actions without any extra validation.
+	 *
+	 * @param array $ids
+	 *
+	 * @return void
```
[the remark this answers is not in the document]
thenbrent: It sounds like it's actually the recommended practice, not just a personal preference: https://stackoverflow.com/a/2061559

As we have a lot of instances of this, I've created #107 to address them. I'd prefer this instance be removed as part of that rather than rebasing its removal into Cesar's commit.

classes/ActionScheduler_ListTable.php
```diff
@@ -0,0 +1,378 @@
...
+				'ID'     => $id,
+				'hook'   => $action->get_hook(),
+				'status' => $action->get_status(),
+				'args'   => $action->get_args(),
+				'group'  => $action->get_group(),
+				'log'    => $this->logger->get_logs( $id ),
+				'recurrence' => $this->get_recurrence( $action ),
+				'scheduled'  => $action->get_schedule(),
+			);
+		}
```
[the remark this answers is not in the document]
thenbrent: Agreed. I also considered `$action` being the "item" instead of looping over all the items here and determining half the data. I'm not sure if this should be blocking, but it could definitely use improvements.

classes/ActionScheduler_wpPostStore.php
```diff
@@ -119,12 +119,68 @@ protected function make_action_from_post( $post ) {
...
+		if ( self::STATUS_PENDING === $status ) {
+			$action_class = 'ActionScheduler_StoredAction';
+		} elseif ( self::STATUS_CANCELED === $status ) {
+			$action_class = 'ActionScheduler_CanceledAction';
 		} else {
-			$action = new ActionScheduler_FinishedAction( $hook, $args, $schedule, $group );
+			$action_class = 'ActionScheduler_FinishedAction';
+		}
+
+		return new $action_class( $hook, $args, $schedule, $group, $post->ID, $status, $post->post_password );
```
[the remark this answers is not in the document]
thenbrent: Good idea. I haven't addressed it yet, I'll review your prototype for other issues first and will come back to it.

classes/ActionScheduler_ListTable.php
```diff
@@ -0,0 +1,378 @@
...
+
+	/**
+	 * Prints the logs entries inline. We do so to avoid loading Javascript and other hacks to show it in a modal.
+	 *
+	 * @param array $row Action array.
+	 */
+	public function column_log( array $row ) {
+		echo '<ol>';
+		$timezone = new DateTimezone( 'UTC' );
+		foreach ( $row['log'] as $log ) {
```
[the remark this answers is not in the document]
thenbrent: I addressed this in a (hopefully) better way by adding a new `ActionScheduler_ListTable::print_log_entry()` methods which uses parameter type declaration to show that the `$log_entry` parameter is an instance of `ActionScheduler_LogEntry`. See SHA: 58ba267c7a68eb0d8ef369fbeba256d3ecbb49e3

classes/ActionScheduler_ListTable.php
```diff
@@ -0,0 +1,378 @@
...
+
+	/**
+	 * Serializes the argument of an action to render it in a human friendly format.
+	 *
+	 * @param array $row The array representation of the current row of the table
+	 *
+	 * @return string
+	 */
+	public function column_args( array $row ) {
+		return '<code>' . json_encode( $row['args'] ) . '</code>';
```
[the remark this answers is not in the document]
thenbrent: I've avoided that ambiguity by displaying it as a list now with SHA: b7e6fbc8d7340f9ec71d83e1adfe07c35170c51d

classes/ActionScheduler_ListTable.php
```diff
@@ -0,0 +1,378 @@
...
+	 * Based on Crontrol::interval() funciton by Edward Dale: https://wordpress.org/plugins/wp-crontrol/
+	 *
+	 * @param int $interval A interval in seconds.
+	 * @return string A human friendly string representation of the interval.
+	 */
+	private static function human_interval( $interval ) {
+
+		// array of time period chunks
+		$chunks = array(
+			array( 60 * 60 * 24 * 365 , _n_noop( '%s year', '%s years', 'action-scheduler' ) ),
```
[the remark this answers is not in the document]
thenbrent: Fixed both of those up with SHA: 32a68a12528e8ea9532ddd00db71ba9c2f89b74f and SHA: ce47a6d9a9f8ba826732350d104c485c0f18d1eb

I then rewrote `human_interval()` entirely with SHA: f7692983221e5aec5ff5e7c389a6568196613cb7 because it was gross, duplicating some code, and it seemed worth updating because it could be useful for more than just displaying the recurrence interval (so I used it for displaying the interval until a scheduled date too).

thenbrent:
Thanks a bunch for the review @JPry. 🙇 

I've fixed up most of the issues and responded to those not yet fixed up. The main one left to address relates to moving instantiation logic out of the store and into the factory. I'll come back to that after I have looked properly at https://github.com/Prospress/action-scheduler/pull/88#issuecomment-363942542

classes/ActionScheduler_ListTable.php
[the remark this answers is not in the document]
JPry: Sounds good to me 👍

JPry:
@thenbrent Just an FYI in case you hadn't looked back at this page, this PR currently has merge conflicts.

thenbrent:
Closing this now in favour of #113.

Issue #26, opened 2014-11-06
Invalid `post_content` causes all jobs to fail


ChuckMac:
I came across a situation where there was an invalid entry in the `post_content` of a scheduled action caused the entire `do_batch` process to quit.

I have not been able to recreate what caused the issue in the first place in my dev environment, but somehow the `post_content` field (which is a serialized array) ended up being improperly encoded and the trailing quote to close the array ended up as a `&quot;` instead of a `"`.  Ex:

`["This is an invalid value &quot;]`

Regardless, it would seem to be better if the scheduler reaches an invalid entry it is able to continue processing the remaining events rather than stopping all processing.

thenbrent:
Fixed with #27.

PR #89, opened 2017-08-21
Set up WP CLI command for running the actions scheduler

### LINKED ISSUE #84: Run scheduler via WP-CLI
Would be great to be able to run the scheduler via WP-CLI and get some better messaging as it processes each task, letting you go through all the tasks in the queue, and telling you how many tasks are left (using progress bar).

For the interim, I've been trying something like:
`wp eval "do_action( 'action_scheduler_run_queue' );"`

Though I'm not 100% sure if that works.

JPry:
Fixes #84.


classes/ActionScheduler_Store.php
```diff
@@ -124,5 +124,11 @@ public static function instance() {
 		}
 		return self::$store;
 	}
+
+	/**
+	 * @param string $claim_id
+	 *
+	 * @return array
+	 */
+	abstract public function find_actions_by_claim_id( $claim_id );
```
JPry: Just one thing to call out here: I added this to ensure IDE code completion works properly. I didn't see this class extended anywhere else. If this is a class that *is* often extended elsewhere, I can change this into a noop method instead of an abstract method to prevent errors for anyone updating.

classes/ActionScheduler.php
```diff
@@ -97,6 +97,10 @@ public static function init( $plugin_file ) {
 		add_action( 'init', array( $admin_view, 'init' ), 0, 0 ); // run before $store::init()
 
 		require_once( self::plugin_path('functions.php') );
+
+		if ( defined( 'WP_CLI' ) && WP_CLI ) {
+			\WP_CLI::add_command( 'action-scheduler', 'ActionScheduler_WPCLI_Scheduler_command' );
```
crodas: Would you mind removing `\`? `WP_CLI` is enough here.  Removing the slash has two cool things

1. Makes this 5.2 friendly
2. The slash is not needed, there is no namespace in this file.

classes/ActionScheduler_WPCLI_QueueRunner.php
```diff
@@ -0,0 +1,223 @@
...
+	}
+
+	/**
+	 * Set up the WP CLI progress bar.
+	 *
+	 * @author Jeremy Pry
+	 */
+	protected function setup_progress_bar() {
+		$count              = count( $this->actions );
+		$this->progress_bar = \WP_CLI\Utils\make_progress_bar(
```
crodas: I assume this file is autoloaded, right? If that is the case it's fine. Otherwise including it automatically will break 5.2 compatibility.
Also, I think you could use some `use WP_CLI\Utils` in beginning of the file.

classes/ActionScheduler_WPCLI_Scheduler_command.php
```diff
@@ -0,0 +1,59 @@
...
+			$store   = ActionScheduler_Store::instance();
+			$monitor = new ActionScheduler_FatalErrorMonitor( $store );
+			$cleaner = new ActionScheduler_QueueCleaner( $store );
+
+			// Get the queue runner instance
+			$runner = new ActionScheduler_WPCLI_QueueRunner( $store, $monitor, $cleaner );
+
+			// Determine how many tasks will be run.
+			$total = $runner->setup( $batch, $force );
+			\WP_CLI::line(
```
crodas: `WP_CLI` is enough

composer.json
```diff
@@ -5,5 +5,8 @@
   "type": "wordpress-plugin",
   "license": "GPL-3.0",
   "minimum-stability": "dev",
-  "require": {}
+  "require": {},
+  "require-dev": {
+    "wp-cli/wp-cli": "^1.3"
```
crodas: Any reason why it is loaded as `dev`?

classes/ActionScheduler_WPCLI_QueueRunner.php
JPry: Yes, the file relies on the existing autoloading.

I didn't do a `use` statement because it was a single function call. Plus, it is a function instead of a class, so even with `use WP_CLI\Utils;`, you would still have to call `Utils\make_progress_bar()`.

composer.json
JPry: I loaded it as `dev` because WP CLI will be installed separately. It's just helpful for IDEs to have code completion during development, but it's not necessary for production.

JPry:
@crodas @thenbrent This is ready for another look. I responded to questions and also made some changes.

In regards to Travis failing, it looks like it's unrelated to any changes I made... Is there something that needs to be done about that?

classes/ActionScheduler_Store.php
thenbrent: @JPry we actually had a patch which was adding this already in an outstanding PR - #87 (see [this commit specifically](https://github.com/Prospress/action-scheduler/pull/87/commits/2b0cebd3e6bc47baf1919baed60b007f71c1a8b4)).

Now that #87 has been merged, could you please:

1. remove the commit which introduced this
1. rebase the rest of the PRs commits on top of the `1-6-dev` branch
1. change the PR to be against the `1-6-dev` branch instead of master (we'll release this with 1.6)

> In regards to Travis failing, it looks like it's unrelated to any changes I made... Is there something that needs to be done about that?

That should also be fixed once rebasing on top of `1-6-dev` as it was fixed in #87 with https://github.com/Prospress/action-scheduler/pull/87/commits/4802f134c53815d37370d9e6edd9f5052b36f320 :)

JPry:
@thenbrent The updates you requested have been made.

classes/ActionScheduler_Abstract_QueueCleaner.php
```diff
@@ -0,0 +1,62 @@
+<?php
+
+/**
+ * Abstract class with common Queue Cleaner functionality.
+ */
+abstract class ActionScheduler_Abstract_QueueCleaner {
```
thenbrent: Was this accidentally given the `QueueCleaner` suffix?

It seems like `ActionScheduler_Abstract_QueueRunner` is a more suitable name as it includes the common `QueueRunner` functionality, not functionality from `ActionScheduler_QueueCleaner`.

classes/ActionScheduler_Abstract_QueueCleaner.php
```diff
@@ -0,0 +1,62 @@
...
+	/**
+	 * Process an individual action.
+	 *
+	 * @param int $action_id The action ID to process.
+	 */
+	protected function process_action( $action_id ) {
+		$action = $this->store->fetch_action( $action_id );
+		try {
+			do_action( 'action_scheduler_before_execute', $action_id );
+			$this->store->log_execution( $action_id );
```
thenbrent: Was this intentionally moved outside the `try/catch` block and before the `'action_scheduler_before_execute'` hook? If so, why?

We used to have them run like this back in 2014, but that was changed with SHA: 1da6a46f and SHA: 00c90a3. See #26 and #61 for more info. I think it's better to keep them as they are in v1.5.3 unless there is a compelling reason to go back to this order.

Between these changes and the one where `process_action()` has been made `protected`  (which was changed with SHA: e1a0ab9603d62), I'm wondering if this code was accidentally copied from a much older version of Action Scheduler?

classes/ActionScheduler_Abstract_QueueCleaner.php
```diff
@@ -0,0 +1,62 @@
...
+
+	/** @var ActionScheduler_Store */
+	protected $store;
+
+	/**
+	 * Process an individual action.
+	 *
+	 * @param int $action_id The action ID to process.
+	 */
+	protected function process_action( $action_id ) {
```
thenbrent: This function was `public` when it was in `ActionScheduler_QueueRunner`. Changing it to `protected` here will break any 3rd party code calling it directly.

classes/ActionScheduler_QueueCleaner.php
```diff
@@ -74,5 +74,15 @@ public function mark_failures() {
...
+	/**
+	 * Do all of the cleaning actions.
+	 *
+	 * @author Jeremy Pry
+	 */
+	public function clean() {
+		$this->delete_old_actions();
+		$this->reset_timeouts();
+		$this->mark_failures();
+	}
```
thenbrent: Nice change - much neater encapsulation of the cleanup logic. 👍

classes/ActionScheduler_WPCLI_QueueRunner.php
```diff
@@ -0,0 +1,232 @@
...
+	 * @param ActionScheduler_FatalErrorMonitor $monitor
+	 * @param ActionScheduler_QueueCleaner      $cleaner
+	 *
+	 * @throws Exception When this is not run within WP CLI
+	 */
+	public function __construct(
+		ActionScheduler_Store $store,
+		ActionScheduler_FatalErrorMonitor $monitor,
+		ActionScheduler_QueueCleaner $cleaner
+	) {
```
thenbrent: I'd prefer to keep the function declarations consistent with all the others, which means all parameters declared on one line.

classes/ActionScheduler_WPCLI_QueueRunner.php
```diff
@@ -0,0 +1,232 @@
...
+		if ( ! ( defined( 'WP_CLI' ) && WP_CLI ) ) {
+			throw new Exception( __( 'The ' . __CLASS__ . ' class can only be run within WP CLI.', 'action-scheduler' ) );
+		}
+
+		$this->store   = $store;
+		$this->monitor = $monitor;
+		$this->cleaner = $cleaner;
```
thenbrent: I like the use of dependency injection for the monitor and cleaner. Not essential, but if you're feeling generous, it's probably worth doing that consistently in the old `ActionScheduler_QueueRunner` now too. Though the 2nd and 3rd params on its `__construct()` will need to have default values set to be backward compatible.

classes/ActionScheduler_Abstract_QueueCleaner.php
JPry: Yep, that was a mistake in naming. I'll change that to the correct name.

classes/ActionScheduler_WPCLI_QueueRunner.php
JPry: I brought the arguments onto multiple lines since it was getting to be a bit long, but it's no big deal to put them back on one line.

classes/ActionScheduler_Abstract_QueueCleaner.php
JPry: It actually was an intentional change to move the `$action = $this->store->fetch_action( $action_id );` statement out of the try/catch block. The main reason for that was that my IDE was giving a warning at the end of the method on the `$this->schedule_next_instance( $action );` line. The warning was that `$action` might not have been defined, since the `try` block could exit earlier (even if that's not likely).

In looking at the flow of `ActionScheduler_wpPostStore::fetch_action()`, I didn't see any possibility of it throwing an exception, so it seemed safe to move that particular statement outside of the try/catch block.

I wasn't aware of the history here, so I can certainly move it back into the block if you think that it would be best to do so.

JPry:
For visibility (GitHub is hiding the conversation under outdated changes), here's a copy of [my response](https://github.com/Prospress/action-scheduler/pull/89#discussion_r135559010) to @thenbrent's [question](https://github.com/Prospress/action-scheduler/pull/89#discussion_r135371649).

@thenbrent  said:
> Was this [`$action = $this->store->fetch_action( $action_id );`] intentionally moved outside the `try/catch` block and before the `'action_scheduler_before_execute'` hook? If so, why?
>
> We used to have them run like this back in 2014, but that was changed with SHA: 1da6a46f and SHA: 00c90a3. See #26 and #61 for more info. I think it's better to keep them as they are in v1.5.3 unless there is a compelling reason to go back to this order.
> 
> Between these changes and the one where `process_action()` has been made `protected`  (which was changed with SHA: e1a0ab9603d62), I'm wondering if this code was accidentally copied from a much older version of Action Scheduler?

My reply:
> It actually was an intentional change to move the `$action = $this->store->fetch_action( $action_id );` statement out of the try/catch block. The main reason for that was that my IDE was giving a warning at the end of the method on the `$this->schedule_next_instance( $action );` line. The warning was that `$action` might not have been defined, since the `try` block could exit earlier (even if that's not likely).
> 
> In looking at the flow of `ActionScheduler_wpPostStore::fetch_action()`, I didn't see any possibility of it throwing an exception, so it seemed safe to move that particular statement outside of the try/catch block.
>
> I wasn't aware of the history here, so I can certainly move it back into the block if you think that it would be best to do so.

classes/ActionScheduler_WPCLI_QueueRunner.php
```diff
@@ -0,0 +1,220 @@
...
+
+	/**
+	 * Set up the WP CLI progress bar.
+	 *
+	 * @author Jeremy Pry
+	 */
+	protected function setup_progress_bar() {
+		$count              = count( $this->actions );
+		$this->progress_bar = \WP_CLI\Utils\make_progress_bar(
+			sprintf( _n( 'Running %d task', 'Running %d tasks', $count, 'action-scheduler' ), number_format_i18n( $count ) ),
```
thenbrent: Why use `task`/`tasks` here instead of `action`/`actions`?

classes/ActionScheduler_WPCLI_QueueRunner.php
```diff
@@ -0,0 +1,220 @@
...
+class ActionScheduler_WPCLI_QueueRunner extends ActionScheduler_Abstract_QueueRunner {
+
+	/** @var array */
+	protected $actions;
+
+	/** @var  ActionScheduler_ActionClaim */
+	protected $claim;
+
+	/** @var string */
+	protected $id;
```
thenbrent: Renaming this `$claim_id` would help disambiguate it. I thought for a second the _runner_ itself might have been getting an ID rather than this value being used for a batch's claim ID.

thenbrent: Actually on further investigation, the `$id` property seems unnecessary. It stores `$this->claim->get_id()`, so `$this->claim->get_id()` could be called directly instead on the one occasion where `$this->id` is used.

classes/ActionScheduler_WPCLI_QueueRunner.php
```diff
@@ -0,0 +1,220 @@
...
+	 *
+	 * @author Jeremy Pry
+	 * @return int The number of actions processed.
+	 */
+	public function run() {
+		$this->setup_progress_bar();
+		foreach ( $this->actions as $action_id ) {
+			// Error if we lost the claim.
+			$all_actions = array_flip( $this->store->find_actions_by_claim_id( $this->id ) );
+			if ( ! array_key_exists( $action_id, $all_actions ) ) {
```
thenbrent: The `array_flip()` usage seems unnecessary here. Couldn't `in_array()` be used instead of `array_key_exists()` to save having to use `array_flip()`?

classes/ActionScheduler_WPCLI_QueueRunner.php
```diff
@@ -0,0 +1,220 @@
...
+			if ( 0 === $this->progress_bar->current() % 50 ) {
+				$this->stop_the_insanity();
+			}
+		}
+
+		$completed = $this->progress_bar->current();
+		$this->finish_progress_bar();
+
+		return $completed;
+	}
```
thenbrent: Because this instance of `run()` isn't calling `'action_scheduler_before_process_queue'` and `'action_scheduler_after_process_queue'`, comment counting isn't being deferred by `ActionScheduler_wpCommentLogger`. Not sure if that makes much/any perf difference in the WP CLI context and not a major issue, but I thought it was something worth noting.

classes/ActionScheduler_Abstract_QueueRunner.php
```diff
@@ -0,0 +1,81 @@
...
+		$this->cleaner = $cleaner;
+	}
+
+	/**
+	 * Process an individual action.
+	 *
+	 * @param int $action_id The action ID to process.
+	 */
+	public function process_action( $action_id ) {
+		$action = $this->store->fetch_action( $action_id );
```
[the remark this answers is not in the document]
JPry: I had moved it out because I thought there might be a chance that the `$action` variable could go undefined, and lead to warnings outside of the try/catch block. I also noticed that the hook is named `action_scheduler_before_execute`, so I was assuming that the main thing was to ensure that the hook fired before calling `$action->execute();`.

If the best solution to maintain current functionality and compatibility is to keep things as-is here, that's no problem. But if we can limit what is inside the try/catch block, then we could move the `$action` initialization and the `action_scheduler_before_execute` hook outside of the block, and allow only catching the execution.

classes/ActionScheduler_Abstract_QueueRunner.php
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	/**
+	 * Run the queue cleaner.
+	 *
+	 * @author Jeremy Pry
+	 */
+	protected function run_cleanup() {
+		$cleaner = new ActionScheduler_QueueCleaner( $this->store );
+		$cleaner->clean();
```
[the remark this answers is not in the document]
JPry: Good catch!

classes/ActionScheduler_WPCLI_QueueRunner.php
JPry: For as many items as we're potentially doing, `array_flip` would be much better than `in_array`, but definitely not the way it is currently written. Ideally I would move `array_flip` before the `foreach` loop, but then we wouldn't be checking up-to-date data with each iteration.

I'll just switch this to use `in_array`.

classes/ActionScheduler_WPCLI_Scheduler_command.php
```diff
@@ -0,0 +1,59 @@
...
+		$batch = \WP_CLI\Utils\get_flag_value( $assoc_args, 'batch-size', 100 );
+		$force = \WP_CLI\Utils\get_flag_value( $assoc_args, 'force', false );
+
+		$completed = 0;
+
+		try {
+			// Set up the class instances we'll need
+			$store   = ActionScheduler_Store::instance();
+			$monitor = new ActionScheduler_FatalErrorMonitor( $store );
+			$cleaner = new ActionScheduler_QueueCleaner( $store );
```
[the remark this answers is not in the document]
JPry: The main reason for setting up the constructor of `ActionScheduler_WPCLI_QueueRunner` without defaults was to be a bit more OOP where possible. But I guess in this case it would make sense to stick with what is already there since the parent class will now handle things for us.

JPry:
@thenbrent I think I addressed everything you mentioned, with just one remaining question [here](https://github.com/Prospress/action-scheduler/pull/89#discussion_r143544269).

JPry:
@james-allan I added you here since Brent is likely to be out for a bit, in case you wanted to weigh in.

classes/ActionScheduler_Abstract_QueueRunner.php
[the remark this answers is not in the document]
thenbrent: > If the best solution to maintain current functionality and compatibility is to keep things as-is here

That's the best approach here. Please revert the changes. 🙂

thenbrent:
> @thenbrent I think I addressed everything you mentioned, with just one remaining question here.

@JPry I've just left an answer for that. I think once it's addressed, this should be ready to merge. 🙂

JPry:
@thenbrent updated!

thenbrent:
LGTM. Fantastic work @JPry. 🥇 💯

sc0ttkclark:
👏

PR #92, opened 2017-09-21
Add `get_date()` to the ActionScheduler_LogEntry

crodas:
Add `get_date()` which contains the date about when the current log happened.

### DIFF
--- a/classes/ActionScheduler_ListTable.php
+++ b/classes/ActionScheduler_ListTable.php
@@ -183,7 +183,7 @@ class ActionScheduler_ListTable extends PP_List_Table {
 		echo '<div id="log-' . $row['ID'] . '" class="log-modal hidden" style="max-width:800px">';
 		echo '<h3>' . esc_html( sprintf( __( 'Log entries for %d', 'action-scheduler' ),  $row['ID'] ) ) . '</h3>';
 		foreach ( $row['comments'] as $log ) {
-			echo '<p><strong>' . esc_html( $log->get_date() ) . '</strong> ' . esc_html( $log->get_message() ) . '</p>';
+			echo '<p><strong>' . esc_html( $log->get_date()->format( 'Y-m-d H:i:s' ) ) . '</strong> ' . esc_html( $log->get_message() ) . '</p>';
 		}
 		echo '</div>';
 
--- a/classes/ActionScheduler_ListTable.php
+++ b/classes/ActionScheduler_ListTable.php
@@ -181,9 +181,9 @@ class ActionScheduler_ListTable extends PP_List_Table {
 	 */
 	public function column_comments( array $row ) {
 		echo '<div id="log-' . $row['ID'] . '" class="log-modal hidden" style="max-width:800px">';
-		echo '<h3>' . sprintf( __( 'Log entries for %d', 'action-scheduler' ),  $row['ID'] ) . '</h3>';
+		echo '<h3>' . esc_html( sprintf( __( 'Log entries for %d', 'action-scheduler' ),  $row['ID'] ) ) . '</h3>';
 		foreach ( $row['comments'] as $log ) {
-			echo '<p>' . $log->get_message() . '</p>';
+			echo '<p><strong>' . esc_html( $log->get_date() ) . '</strong> ' . esc_html( $log->get_message() ) . '</p>';
 		}
 		echo '</div>';
 
@@ -274,6 +274,8 @@ class ActionScheduler_ListTable extends PP_List_Table {
 			'per_page' => $per_page,
 			'offset'   => $this->get_items_query_offset(),
 			'status'   => $this->get_request_status(),
+			'orderby'  => 'modified',
+			'order'    => 'ASC',
 		);
 
 		$this->items = array();
--- a/classes/ActionScheduler_LogEntry.php
+++ b/classes/ActionScheduler_LogEntry.php
@@ -13,28 +13,22 @@ class ActionScheduler_LogEntry {
 	 *
 	 * @param mixed  $action_id	Action ID
 	 * @param string $message   Message
-	 * @param string $date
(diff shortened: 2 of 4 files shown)


classes/ActionScheduler_LogEntry.php
```diff
@@ -4,12 +4,37 @@
...
+	protected $message =  '';
+	protected $date;
 
-	public function __construct( $action_id, $message ) {
+	/**
+	 * Constructor
+	 *
+	 * @param mixed  $action_id	Action ID
+	 * @param string $message   Message
+	 * @param string $date
```
thenbrent: What format does this expect the date string to be in? Also what timezone? Best to explicitly mention that to avoid bugs from inconsistent use.

MySQL formatted date/time string in GMT/UTC timezone will match Subscriptions. Alternatively, we could match the behaviour of `wc_string_to_datetime()` in WC 3.0+: https://github.com/woocommerce/woocommerce/blob/3.1.2/includes/wc-formatting-functions.php#L579:L597

This is a public API, we need to be clear to 3rd parties how to use it. 🙂

classes/ActionScheduler_LogEntry.php
```diff
@@ -4,12 +4,37 @@
...
+	 */
+	public function __construct( $action_id, $message, $date  = '' ) {
 		$this->action_id = $action_id;
-		$this->message = $message;
+		$this->message   = $message;
+		$this->date      = $date;
```
thenbrent: I think this should be stored as a `DateTime` object to be consistent regardless of the date format passed in. At the moment, its value could be any type of date/time string, e.g. `2017-09-20 20:00:00` or `January 1st, 2018` or something else entirely.

classes/ActionScheduler_LogEntry.php
```diff
@@ -4,12 +4,37 @@
...
+	 * through the `action_scheduler_date_format` filter.
+	 *
+	 * The date is the same timezone as the WordPress site.
+	 *
+	 * @return string
+	 */
+	public function get_date() {
+		$date = as_get_datetime_object( $this->date );
+		$date->setTimezone( ActionScheduler_TimezoneHelper::get_local_timezone() );
+		return $date->format( apply_filters( 'action_scheduler_date_format', 'Y-m-d H:i:s' ) );
```
thenbrent: If the above suggestion is followed, and `$this->date` stores a `DateTime` object, then this can simply return the `DateTime` without messing with timezones or formatting and let the calling code format it however it needs.

classes/ActionScheduler_ListTable.php
```diff
@@ -181,9 +181,9 @@ public function column_args( array $row ) {
 	 */
 	public function column_comments( array $row ) {
 		echo '<div id="log-' . $row['ID'] . '" class="log-modal hidden" style="max-width:800px">';
-		echo '<h3>' . sprintf( __( 'Log entries for %d', 'action-scheduler' ),  $row['ID'] ) . '</h3>';
+		echo '<h3>' . esc_html( sprintf( __( 'Log entries for %d', 'action-scheduler' ),  $row['ID'] ) ) . '</h3>';
 		foreach ( $row['comments'] as $log ) {
-			echo '<p>' . $log->get_message() . '</p>';
+			echo '<p><strong>' . esc_html( $log->get_date() ) . '</strong> ' . esc_html( $log->get_message() ) . '</p>';
```
thenbrent: Ideally this should display the date's timezone to avoid ambiguity. If the suggestion to make `ActionScheduler_LogEntry::get_date()` return a `DateTime` is taken onboard, the code here can be responsible for displaying that.

classes/ActionScheduler_LogEntry.php
thenbrent: If you do want to provide a way to get the formatted value of the date, then we should add a new function, like `get_date_to_display()` or similar. But I wouldn't bother with that. Keep `ActionScheduler_LogEntry` super simple for now. 🙂

cesar-prospress:
Does it look any better now @thenbrent ?

classes/ActionScheduler_wpCommentLogger.php
```diff
@@ -49,7 +49,10 @@ public function get_entry( $entry_id ) {
 		if ( empty($comment) || $comment->comment_type != self::TYPE ) {
 			return new ActionScheduler_NullLogEntry();
 		}
-		return new ActionScheduler_LogEntry( $comment->comment_post_ID, $comment->comment_content, $comment->comment_type );
+
+		$date = as_get_datetime_object( $comment->comment_date_gmt );
+		$date->setTimezone( ActionScheduler_TimezoneHelper::get_local_timezone() );
+		return new ActionScheduler_LogEntry( $comment->comment_post_ID, $comment->comment_content, $date );
```
thenbrent: `as_get_datetime_object()` accepts a `$timezone` as the 2nd parameter and once it is condensed into the `as_get_datetime_object()` call, the `$date` param is no longer used, so you can include all this code on the one line when constructing the `ActionScheduler_LogEntry` instance.

thenbrent:
@cesar-prospress just the one last comment to tighten it up a bit, then it looks good to merge. 👍

classes/ActionScheduler_wpCommentLogger.php
cesar-prospress: I don't think that's going to work. Please correct me if I'm wrong (that's most likely the case 😆 )
`comment_date_gmt` is *ALWAYS* GMT, so at construction time, the `Datetime` should have the UTC timezone.
After the datetime object is created, then we need to set the site's timezone (right before displaying).

thenbrent: @cesar-prospress ah yes, you're right. 👍 

And we don't want to use `comment_date`, because we can't guarantee that the site's timezone hasn't changed since that was set. We definitely need to use `comment_date_gmt`.

Issue #96, opened 2017-09-29
Add ActionScheduler_Action::get_claim_id()


thenbrent:
So that an action can share information about its claim status/ID.

Will likely also require a new method in `ActionScheduler_Store` like `get_claim_id_for_action()` instead of just the `find_actions_by_claim_id()` method currently available.

Came up as part of: #88

thenbrent:
Closing in favour of using a method on the data store, see SHA: 34119e78428a0 for more details.

Issue #95, opened 2017-09-29
Add ActionScheduler_Action::get_id()


thenbrent:
So that an action can identify itself.

Came up as part of: https://github.com/Prospress/action-scheduler/pull/88

thenbrent:
Closing in favour of using a method on the data store, see #116 for more details.

PR #98, opened 2017-10-20
[WIP] Added setters/getters to Action objects.

### LINKED ISSUE #94: Add ActionScheduler_Action::get_status()
Instead of requiring the use of `ActionScheduler::store()->get_status( $action_id )`.

Came up as part of: https://github.com/Prospress/action-scheduler/pull/88

cesar-prospress:
The idea behind this PR is to add setters/getters to Action objects. All
the properties that belongs to the actions should be accesible directly
from the object, without any external calls to the store.

This PR adds a fallback mechanism reading most unset properties from the
store, but ideally the store should create the action object property.

To be able to read the claim ID off an action id, this PR introduces a
new abstract method to the Store abstract class
(`get_claim_id_for_action`).

This PR wants to fix #94, #95 and #96 and it will be merged against #88  (and not master).

Beware to also implement `get_claim_id_for_action` in https://github.com/Prospress/action-scheduler-custom-tables

### DIFF
--- a/classes/ActionScheduler_Action.php
+++ b/classes/ActionScheduler_Action.php
@@ -10,12 +10,100 @@ class ActionScheduler_Action {
 	protected $schedule = NULL;
 	protected $group = '';
 
+	/** @var mixed */
+	protected $id;
+
+	/** @var int */
+	protected $status;
+
+	/** @var mixed */
+	protected $claim_id;
+
+	/**
+	 * This array keeps track of all the properties that were set through the setters.
+	 *
+	 * Why is it needed?
+	 *
+	 *  1. If a variable was not assigned it may be populated on demand.
+	 *  2. FALSE, NULL or 0 could be valid properties value, thus making it impossible to tell
+	 *     without an external aid if our properties were assigned or not.
+	 *
+	 * To make things efficent the value of this array is always TRUE and the key is the property
+	 * name.
+	 *
+	 * @type array
+	 */
+	protected $has = array();
+
 	public function __construct( $hook, array $args = array(), ActionScheduler_Schedule $schedule = NULL, $group = '' ) {
 		$schedule = empty( $schedule ) ? new ActionScheduler_NullSchedule() : $schedule;
 		$this->set_hook($hook);
 		$this->set_schedule($schedule);
 		$this->set_args($args);
 		$this->set_group($group);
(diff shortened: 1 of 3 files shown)


cesar-prospress:
Maybe we can force the caller to pass a valid `$id` in the constructor. But I preferred not to change the constructor signature to accept a new parameter without any discussions first.
This PR is functional but it relies on the constructor to call to `set_id`.

toddlahman:
Do the setters need to return?

cesar-prospress:
> Do the setters need to return?
Maybe not, but I think it's handy if it returns `$this`, so the caller can do `$action->set_id(..)->set_claim_id(...);`.

classes/ActionScheduler_Action.php
```diff
@@ -18,6 +43,69 @@ public function __construct( $hook, array $args = array(), ActionScheduler_Sched
...
 
+	/**
+	 * Sets the ID for this current action.
+	 *
+	 * @param mixed $id Action ID
+	 *
+	 * @return self
+	 */
+	public function set_id( $id ) {
+		$this->has['id'] = true;
```
JPry: I think this should be `$this->has[ $id ]`

cesar-prospress:
Nope. The idea behind `has` (maybe I need to rename it) is to keep track of
which variables were set or not. In this case 'id' is the correct key.

thenbrent:
I'm going to close this one and start afresh.

The requirement exists, and the idea is sound, but the implementation introduces too many inconsistencies with existing Action Scheduler architecture for it to be the chosen implementation.

Specifically, the setters in `ActionScheduler_Action` should be `protected` to be consistent with existing setters, and this new data (ID, status and claim ID) should be passed into the action's constructor, like other data on an action.

The `$has` property is also used only for this data, while existing data used direct `protected` properties, like `$this->schedule` or `$this->hook`. I don't see any reason to have two inconsistent data storage modes (`$has` could be replace by setting `NULL` defaults).

Changing the setters also means  `make_action_from_post()` is no longer needed, nor is `ActionScheduler_Store::get_claim_id_for_action()`, meaning nothing in this PR would ultimately be used in its current form, which is why I'm going to close this in favour of new commits with similar, but more consistent style, applied directly to #85.

PR #99, opened 2017-10-31
Addded checks for past due actions

cesar-prospress:
This commit adds checks for past due actions. Because Action Scheduler is action agnostic it will trigger an action with all the past due actions. It is the responsability of the action owner (library which created it) to warn their users how to fix their WP-Cron or how to run those actions manually.

This PR would help to improve the [user experience in Subscriptions](https://github.com/Prospress/woocommerce-subscriptions/issues/2380).

### DIFF
--- a/classes/ActionScheduler_AdminView.php
+++ b/classes/ActionScheduler_AdminView.php
@@ -45,29 +45,40 @@ class ActionScheduler_AdminView {
 	 * This method will check if there are past due. Because Action Scheduler is action agnostic,
 	 * it will execute an action with a list of the actions.
 	 *
-	 * The action owners should be listening to this action, and do something to warn the users
-	 * about how to fix their WP-Cron or to run their actions manually.
+	 * To make narrow down possible results, and improve query looking for past due actions, any
+	 * interested plugin must add their hook to an array through the `action_scheduler_past_due_hooks`
+	 * filter.
 	 *
-	 * This method is executed by the `admin_notices`, so it is safe to render the admin notification
-	 * from the `action_scheduler_past_due_action` action.
+	 * The action owner should be listening to the `action_schedule_past_due_action` action, 
+	 * and do something to warn the users about how to fix their WP-Cron or to run their actions manually.
 	 *
+	 * This method is executed on the `admin_notices`, so it is safe to render the admin notification.
 	 */
 	public function check_past_due_actions() {
 		$instance = ActionScheduler_Store::instance();
-		if ( $cache = get_transient( __METHOD__ ) ) {
-			$action_ids = $cache['action_ids'];
-		} else {
-			$action_ids = $instance->query_actions( array(
-				'date' => new Datetime( apply_filters( 'action_scheduler_past_due_time', '-5 days' ) ),
-				'status' => ActionScheduler_Store::STATUS_PENDING,
-			) );
-			set_transient( __METHOD__, compact( 'action_ids' ) );
+
+		$actions = apply_filters( 'action_scheduler_past_due_hooks', array() );
+		
+		if ( ! empty( $actions ) ) {
+			return;
+		}
+
+		$action_ids = $instance->query_actions( array(
+			'date'   => new Datetime( apply_filters( 'action_scheduler_past_due_time', '-5 days' ) ),
+			'status' => ActionScheduler_Store::STATUS_PENDING,
+			'hook'   => $actions,
+		) );
+
+		if ( ! empty( $action_ids ) ) {
+			return;
 		}
 
 		$actions = array();
 		foreach ( $action_ids as $id ) {
-			$actions[ $id ] = $instance->fetch_action( $id );
+			$action = $instance->fetch_action( $id );
+			$actions[ $action->get_hook() ][] = $action;
 		}
+
 		do_action( 'action_scheduler_past_due_action', $actions );
 	}
 
--- a/classes/ActionScheduler_AdminView.php
+++ b/classes/ActionScheduler_AdminView.php
@@ -35,12 +35,12 @@ class ActionScheduler_AdminView {
 			} else {
 				add_action( 'admin_menu', array( $self, 'register_menu' ) );
 			}
-			add_action( 'admin_notices', array( $self, 'check_past_due_actions' ) );
+			add_action( 'admin_notices', array( $self, 'past_due_actions' ) );
 		}
 	}
 
 	/**
-	 * Check if there are any past due actions.
+	 * Trigger API for 3rd parties to handle past due actions.
 	 *
 	 * This method will check if there are past due actions. Because Action Scheduler is action agnostic,
 	 * it will trigger an action with a list of hooks that callbacks can use to handle past due actions
@@ -55,7 +55,7 @@ class ActionScheduler_AdminView {
 	 *
 	 * This method is executed on the `admin_notices`, so it is safe to render the admin notification.
 	 */
-	public function check_past_due_actions() {
+	public function past_due_actions() {
 		$instance = ActionScheduler_Store::instance();
 
 		$actions = apply_filters( 'action_scheduler_past_due_hooks', array() );
--- a/classes/ActionScheduler_AdminView.php
+++ b/classes/ActionScheduler_AdminView.php
@@ -40,13 +40,14 @@ class ActionScheduler_AdminView {
 	}
 
 	/**
-	 * Check if there are any past due action
+	 * Check if there are any past due actions.
 	 *
-	 * This method will check if there are past due. Because Action Scheduler is action agnostic,
-	 * it will execute an action with a list of the actions.
+	 * This method will check if there are past due actions. Because Action Scheduler is action agnostic,
+	 * it will trigger an action with a list of hooks that callbacks can use to handle past due actions
+	 * as needed.
 	 *
-	 * To make narrow down possible results, and improve query looking for past due actions, any
-	 * interested plugin must add their hook to an array through the `action_scheduler_past_due_hooks`
+	 * To narrow down possible results, and improve query looking for past due actions, any plugin
+	 * must add their hook name to an array through the `action_scheduler_past_due_hooks`
 	 * filter.
 	 *
 	 * The action owner should be listening to the `action_schedule_past_due_action` action, 
--- a/classes/ActionScheduler_AdminView.php
+++ b/classes/ActionScheduler_AdminView.php
@@ -57,16 +57,16 @@ class ActionScheduler_AdminView {
 	 */
 	public function past_due_actions() {
 
-		$actions = apply_filters( 'action_scheduler_past_due_hooks', array() );
-		
-		if ( ! empty( $actions ) ) {
+		$action_hooks = apply_filters( 'action_scheduler_past_due_hooks', array() );
+
+		if ( ! empty( $action_hooks ) ) {
 			return;
 		}
 
 		$action_ids = ActionScheduler::store()->query_actions( array(
 			'date'   => new Datetime( apply_filters( 'action_scheduler_past_due_time', '-5 days' ) ),
 			'status' => ActionScheduler_Store::STATUS_PENDING,
-			'hook'   => $actions,
+			'hook'   => $action_hooks,
 		) );
 
 		if ( ! empty( $action_ids ) ) {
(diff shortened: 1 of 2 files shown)


classes/ActionScheduler_AdminView.php
```diff
@@ -35,9 +35,42 @@ public function init() {
...
+	 * The action owners should be listening to this action, and do something to warn the users
+	 * about how to fix their WP-Cron or to run their actions manually.
+	 *
+	 * This method is executed by the `admin_notices`, so it is safe to render the admin notification
+	 * from the `action_scheduler_past_due_action` action.
+	 *
+	 */
+	public function check_past_due_actions() {
+		$instance = ActionScheduler_Store::instance();
+		if ( $cache = get_transient( __METHOD__ ) ) {
```
thenbrent: Good idea to cache it so it repeatedly triggers the action without repeating the query 👍

classes/ActionScheduler_AdminView.php
```diff
@@ -35,9 +35,42 @@ public function init() {
...
+			$action_ids = $cache['action_ids'];
+		} else {
+			$action_ids = $instance->query_actions( array(
+				'date' => new Datetime( apply_filters( 'action_scheduler_past_due_time', '-5 days' ) ),
+				'status' => ActionScheduler_Store::STATUS_PENDING,
+			) );
+			set_transient( __METHOD__, compact( 'action_ids' ) );
+		}
+
+		$action_ids = array( 2151 );
```
thenbrent: @cesar-prospress This looks like test code and it needs to be removed.

classes/ActionScheduler_AdminView.php
```diff
@@ -35,9 +35,42 @@ public function init() {
...
+	public function check_past_due_actions() {
+		$instance = ActionScheduler_Store::instance();
+		if ( $cache = get_transient( __METHOD__ ) ) {
+			$action_ids = $cache['action_ids'];
+		} else {
+			$action_ids = $instance->query_actions( array(
+				'date' => new Datetime( apply_filters( 'action_scheduler_past_due_time', '-5 days' ) ),
+				'status' => ActionScheduler_Store::STATUS_PENDING,
+			) );
+			set_transient( __METHOD__, compact( 'action_ids' ) );
```
thenbrent: Why use `compact()` here when the array only has one value?

Perhaps the transient name needs to be more suitable to avoid that. Something like `action_scheduler_past_due_action_ids`?

classes/ActionScheduler_AdminView.php
thenbrent: Good idea to cache it so it repeatedly triggers the action without repeating the query :+1:

However, what happens if the actions are run? The cache value isn't being invalidated, so it will continue to be passed to callbacks on `'action_scheduler_past_due_action'` and seen as "pending" by those callbacks incorrectly until the transient expires.

The transient needs to be invalidated when an action is run. Ideally in a very efficient and scalable way.

classes/ActionScheduler_AdminView.php
thenbrent: The `set_transient()` call needs an to be passed the 3rd `$expiration` parameter, otherwise actions will continue to show as past due until the entire transient cache is cleared - the default expiration time is indefinite.

classes/ActionScheduler_AdminView.php
```diff
@@ -35,9 +35,42 @@ public function init() {
...
+		}
+
+		$action_ids = array( 2151 );
+		$actions = array();
+		foreach ( $action_ids as $id ) {
+			$actions[ $id ] = $instance->fetch_action( $id );
+		}
+		do_action( 'action_scheduler_past_due_action', $actions );
```
thenbrent: What happens if there are 100,000 actions past due, but nothing hooking to `action_scheduler_past_due_action`? This code would still instantiate 100,000 actions on every admin page load.

I think the code here needs to trigger the `action_scheduler_past_due_action` and to instantiate the actions should run _only_ when:

1. `! empty( $action_ids )`
1.  there are callbacks on `'action_scheduler_past_due_action'` to avoid wasting time if there aren't; and

classes/ActionScheduler_AdminView.php
cesar-prospress: Just for my quick prototype so I could check if the object was valid or not. If the prototype looks good in overall I will implement the following improvements:

1. 24 hours of expiration
2. I will cache only the ID, while fetching the objects from the DB I would check if they are still pending. If they are not, I would invalidate the cache (and I would show the notices in the page load).
3. I would somehow limit to just 10 or so elements. If there are more, it is obviously something that needs to be shown like a red warning in the Action Scheduler page.

classes/ActionScheduler_AdminView.php
cesar-prospress: I think we can limit to 10 or so IDs. We should check if there are more, in the admin, while rendering the rows. Perhaps we add a different background for tasks with past due. This notice thing is just to give a quick hint of things not being executed for some reason, using as little resources as possible.

classes/ActionScheduler_AdminView.php
thenbrent: Ah gotcha. That makes sense. Those 3 things sound good. 👍

classes/ActionScheduler_AdminView.php
thenbrent: The gotcha with that is if the first X number of actions is for a plugin that doesn't hook to `action_scheduler_past_due_action`. For example, if _Memberships_ has 1,000 actions in the queue before the first action for Subscriptions, and they are all past due, then Subscriptions notice will never be displayed (at least it won't if Subscriptions is hooking only to its own actions, which it should do).

To get around that, you could have some way to register a "past due" handler/monitor for certain hooks. Then when querying for the past due actions, you can specify the hook/s which have a registered past due handler/monitor only. There's not really much need to get the rest. The only use case I can think of is to allow for a generic notice for Action Scheduler itself, which could be done by a handler registering with a hook that evaluates to `false` if the handler/monitor registration is designed to allow that.

cesar-prospress: I love that idea. I´ll make it happen.

cesar-prospress:
I decided to take a better approach on f33e65d.

Basically here are the changes in a nutshell.

1. Action-scheduler gives the opportunity to anyone to look for past-due actions on a given hook. If it is not extended (it's a filter), it won't do a thing.
2. I removed any cache. Invalidating the cache is a nightmare and the `wp_posts` table has some indexes that helps to narrow down the possible results.
3. I also changed the query/count method to allow `hook` as an string or an array. I will replicate this behaviour in [custome tables](https://github.com/Prospress/action-scheduler-custom-tables) should this approach makes sense.

Thoughts @thenbrent, @JPry, @james-allan ?

cesar-prospress:
The usage is fairly simple. Subscription would need something like this:

```php
add_filter( 'action_scheduler_past_due_hooks', function( $actions ) {
        $actions[] = 'woocommerce_scheduled_subscription_payment';
        return $actions;
} );
```

The checking only happens on `wp-admin`. I will change the code a bit to show only if the current user is an admin or something like that. But the fact we only look  for past-due tasks on `wp-admin` won't hurt the overall performance.

classes/ActionScheduler_AdminView.php
```diff
@@ -35,9 +35,53 @@ public function init() {
...
+	 * must add their hook name to an array through the `action_scheduler_past_due_hooks`
+	 * filter.
+	 *
+	 * The action owner should be listening to the `action_schedule_past_due_action` action, 
+	 * and do something to warn the users about how to fix their WP-Cron or to run their actions manually.
+	 *
+	 * This method is executed on the `admin_notices`, so it is safe to render the admin notification.
+	 */
+	public function check_past_due_actions() {
+		$instance = ActionScheduler_Store::instance();
```
thenbrent: `$instance` is a generic word - any object can be an "instance", so it doesn't give any indication what this variable contains.

In other locations, the store instance is assigned to a variable named `$store`, which more accurately described its contents. `ActionScheduler::store() is also used to access the instance, for simplicities sake (although it is really just an alias).

IMO, as the store is only being used twice, it doesnt' really need a variable at all, you could call `ActionScheduler::store()->query_actions()` and `ActionScheduler::store()-> fetch_action()` directly and save having to use a variable, while also making it clear what is being used.

classes/ActionScheduler_AdminView.php
```diff
@@ -35,9 +35,53 @@ public function init() {
...
+
+		$actions = apply_filters( 'action_scheduler_past_due_hooks', array() );
```
thenbrent: `$actions` is used later on to refer to an array of `ActionScheduler_Action` instances. It would be safe to avoid using the same variable name to refer to two different data sets in the one function. This variable would be better named `$action_hooks` or `$hooks` for that reason (and because that's more accurately the data type it contains).

classes/ActionScheduler_AdminView.php
```diff
@@ -35,9 +35,53 @@ public function init() {
...
+		) );
+
+		if ( ! empty( $action_ids ) ) {
+			return;
+		}
+
+		$actions = array();
+		foreach ( $action_ids as $id ) {
+			$action = $instance->fetch_action( $id );
+			$actions[ $action->get_hook() ][] = $action;
```
thenbrent: Is it necessary to assign the result of `fetch_action( $id )` to `$action`? Why not compress these two lines into one: `$actions[ $action->get_hook() ][] = $instance->fetch_action()` (or as suggested above, use  `ActionScheduler::store()-> fetch_action()` instead of `$instance`).

classes/ActionScheduler_wpPostStore.php
```diff
@@ -317,9 +321,12 @@ public function query_actions( $query = array() ) {
 		}
 		$sql .= " WHERE post_type=%s";
 		$sql_params[] = self::POST_TYPE;
-		if ( $query['hook'] ) {
+		if ( $query['hook'] && is_scalar( $query['hook'] ) ) {
 			$sql .= " AND p.post_title=%s";
 			$sql_params[] = $query['hook'];
+		} else if ( $query['hook'] ) {
+			$sql .= ' AND p.post_title IN (' . implode( ',', array_fill( 0, count( $query['hook'] ), '%s' ) ) . ')';
+			$sql_params = array_merge( $sql_params, $query['hook'] );
```
thenbrent: This duplication highlights issues with duplicating code in `query_actions_count()` from `query_actions()`. That's beyond the socpe of the code here though, so I've mentioned it here: https://github.com/Prospress/action-scheduler/pull/88/#issuecomment-347346308

classes/ActionScheduler_AdminView.php
JPry: +1 to avoiding storing `ActionScheduler_Store::instance()` in a variable. Realistically, I'd like to see the instance we need be injected into the class constructor, but that's way out of scope for this PR 😃

classes/ActionScheduler_AdminView.php
JPry: @thenbrent It looks like we would still need the `$action` variable. Here the code is accessing a method of the variable to use as the array key (`$action->get_hook()`), and the action object is stored to the array as well.

I still agree with using `ActionScheduler_Store::instance()->fetch()` directly.

cesar-prospress: > Is it necessary to assign the result of fetch_action( $id ) to $action? Why not compress these two lines into one: $actions[ $action->get_hook() ][] = $instance->fetch_action() (or as suggested above, use ActionScheduler::store()-> fetch_action() instead of $instance).
 
It makes way easier for callee functions to know if there is something they care or not. Just by simply checking if a key exists or not.

classes/ActionScheduler_AdminView.php
cesar-prospress: > IMO, as the store is only being used twice, it doesnt' really need a variable at all, you could call ActionScheduler::store()->query_actions() and ActionScheduler::store()-> fetch_action() directly and save having to use a variable, while also making it clear what is being used.
 
Works for me.

thenbrent:
I've just pushed up commits to address all most feedback on this. For the diff, see 599f579...e7462b5.

The pieces not addressed have been added to a new issue here: https://github.com/Prospress/action-scheduler/issues/105

This is ready for final review/merge now.

thenbrent:
Closing this in favour of #106.

As #98 has [now been closed](https://github.com/Prospress/action-scheduler/pull/98#issuecomment-360628261), and this PR is based on the branch from that, and was set to merge into that branch, I've opened #106 with the patches from this applied directly to the `issue_85` branch to fixup the merge conflicts brought about from this PRs dependency on SHA: ade19ead85dab1.

PR #104, opened 2018-01-23
Catch exception when deleting old actions

### LINKED ISSUE #103: Race condition when deleting old actions from multiple processes
If more than one queue is running at the same time, there is potential for a race condition when [deleting old actions](https://github.com/Prospress/action-scheduler/blob/6168e5f4c95a9b25fadfc996a49b780641d207c9/classes/ActionScheduler_QueueCleaner.php#L17-L31), because we don't "lock" actions for deletion.

When this occurs, the `ActionScheduler_Store::delete_action()` method can throw an unhandled `InvalidArgumentException`, because in the time between when `ActionScheduler_QueueCleaner::delete_old_actions()` calls `ActionScheduler_Store::query_actions()` and `ActionScheduler_Store::delete_action()`, the other queue may have deleted the action, meaning the ID of the action still exists in the `$actions_to_delete` array in `ActionScheduler_QueueCleaner::delete_old_actions()`, but the `get_post()` call in `ActionScheduler_wpPostStore::delete_action()` will not instantiate it, because it's since been deleted, meaning `empty($post)` will evaluate `true`.

There are a few possible approaches to fix this:

1. claim actions before deleting them. This is the most complex. The added complexity isn't really required as it should be sufficient to just avoid deleting already deleted actions.
1. check if `ActionScheduler_Store::fetch_action()` returns a valid action before calling `ActionScheduler_Store::delete_action()`. This has minor perf issues, as it requires instantiating the action twice, and also won't always fix the issue, because there are still microseconds between when `ActionScheduler_Store::fetch_action()` will be called by `ActionScheduler_QueueCleaner::delete_old_actions()` and when `get_post()` will be called in `ActionScheduler_Store::delete_action()`
1. catch the `InvalidArgumentException` exception and ignore it (perhaps with some type of logging about the issue?). This will guarantee this race condition is ignored, but it may catch other exceptions for new conditions, so it will need special handling.

Related links:

* [Original Slack thread](https://prospress.slack.com/archives/C74DUNZQ8/p1515778094000388)
* [Follow-up Slack thread](https://prospress.slack.com/archives/C03A22H08/p1516121271000709)
* [Support ticket](https://secure.helpscout.net/conversation/506206000/4119?folderId=1524429)

thenbrent:
And add 'action_scheduler_failed_old_action_deletion' hook to add logging outside of Action Scheduler, or to Action Scheduler itself at a later stage once we have logging not linked to the ID directly (which won't work if it's been deleted).

Fixes #103

---

This doesn't address all the issues mentioned in #103, specifically:

> This will guarantee this race condition is ignored, but it may catch other exceptions for new conditions, so it will need special handling.

If we see issues with that, we may want to add a custom exception class for this case, and only catch that, and throw a different exception type for other cases.

### DIFF
--- a/classes/ActionScheduler_QueueCleaner.php
+++ b/classes/ActionScheduler_QueueCleaner.php
@@ -26,7 +26,25 @@ class ActionScheduler_QueueCleaner {
 		) );
 
 		foreach ( $actions_to_delete as $action_id ) {
-			$this->store->delete_action( $action_id );
+			try {
+				$this->store->delete_action( $action_id );
+			} catch ( Exception $e ) {
+
+				/**
+				 * Notify 3rd party code of exceptions when deleting a completed action older than the retention period
+				 *
+				 * This hook provides a way for 3rd party code to log or otherwise handle exceptions relating to their
+				 * actions.
+				 *
+				 * @since 1.6.0
+				 *
+				 * @param int $action_id The scheduled actions ID in the data store
+				 * @param Exception $e The exception thrown when attempting to delete the action from the data store
+				 * @param int $lifespan The retention period, in seconds, for old actions
+				 * @param int $count_of_actions_to_delete The number of old actions being deleted in this batch
+				 */
+				do_action( 'action_scheduler_failed_old_action_deletion', $action_id, $e, $lifespan, count( $actions_to_delete ) );
+			}
 		}
 	}
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,8 @@ env:
     - WP_VERSION=4.5 WP_MULTISITE=1
     - WP_VERSION=4.4 WP_MULTISITE=1
 
+dist: precise
+
 # Grab the setup script and execute
 before_script:
     - source tests/travis/setup.sh $TRAVIS_PHP_VERSION


classes/ActionScheduler_QueueCleaner.php
```diff
@@ -26,7 +26,11 @@ public function delete_old_actions() {
 		) );
 
 		foreach ( $actions_to_delete as $action_id ) {
-			$this->store->delete_action( $action_id );
+			try {
+				$this->store->delete_action( $action_id );
+			} catch ( Exception $e ) {
+				do_action( 'action_scheduler_failed_old_action_deletion', $action_id, $e, $lifespan, count( $actions_to_delete ) );
```
JPry: One minor thing here: I'd like to see a docblock for this action so that we can start having some inline documentation for the actions and filters.

thenbrent: @JPry good idea. I see that's also part of WordPress' [PHP Documentation Standards](https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#4-hooks-actions-and-filters) now. I've fixed that up and rebased in the changed. 👍

JPry:
LGTM 👍

PR #110, opened 2018-02-24
Refactor stored actions APIs

thenbrent:
This is a proposal to refactor the stored actions handling approach introduced in #88 (the PR is using `issue_85` as a base, which is the branch used for #88). I'm introducing it as a separate PR outside of #88 to keep a separate place for discussion and review of it (#88 is already unwieldy enough).

The approach taken here builds on the one proposed by @JPry in this comment https://github.com/Prospress/action-scheduler/pull/88#issuecomment-363942542 to use interfaces to define the shared, public methods available between stored and unstored actions.

It also introduces a new `ActionScheduler_AbstractAction` class to use as a base for both stored and unstored actions. I think this was the missing piece when tackling it previously. Using this shared abstract class allows for different constructors, which means stored actions can require an ID and status to be passed in, while also allow them to be used interchangeably when that data isn't required.

**This makes it possible to maintain full backward compatibility, while still introducing new classes well suited for instantiating stored actions.**

The best demonstration of this is with `ActionScheduler_AbstractAction::save_action()`. Previously, it required an instance of `ActionScheduler_Action` as the first parameter. Therefore, any stored action would need to extend `ActionScheduler_Action`, meaning stored actions needed to honour that class's constructor. However, by changing it to now require an instance of `ActionScheduler_AbstractAction`, it will now gladly accept an instance of `ActionScheduler_Action` or any child class of it, which maintains backward compatibility. It will also now accept an instance of `ActionScheduler_StoredAction`, or any of its child classes.

I also snuck in another related commit to this PR to introduce `ActionFactory::get_stored_action_instance()` in response to feedback on #88.

### DIFF
--- a/classes/ActionScheduler_AbstractAction.php
+++ b/classes/ActionScheduler_AbstractAction.php
@@ -1,0 +1,81 @@
+<?php
+/**
+ * Class ActionScheduler_AbstractAction
+ *
+ * Define shared methods and data for all action types.
+ *
+ * @since 1.6.0
+ */
+abstract class ActionScheduler_AbstractAction implements ActionScheduler_Interface_Action, ActionScheduler_Interface_Scheduled {
+
+	/** @var string */
+	protected $hook = '';
+
+	/** @var array */
+	protected $args = array();
+
+	/** @var ActionScheduler_Schedule */
+	protected $schedule = NULL;
+
+	/** @var string */
+	protected $group = '';
+
+	/** @var bool */
+	protected $is_finished = false;
+
+	public function execute() {
+		return do_action_ref_array( $this->get_hook(), $this->get_args() );
+	}
+
+	/**
+	 * @param string $hook
+	 * @return void
+	 */
+	protected function set_hook( $hook ) {
+		$this->hook = $hook;
+	}
+
+	public function get_hook() {
+		return $this->hook;
+	}
+
+	protected function set_schedule( ActionScheduler_Schedule $schedule ) {
+		$this->schedule = $schedule;
+	}
+
+	/**
+	 * @return ActionScheduler_Schedule
+	 */
+	public function get_schedule() {
+		return $this->schedule;
+	}
+
+	protected function set_args( array $args ) {
+		$this->args = $args;
+	}
+
+	public function get_args() {
+		return $this->args;
+	}
+
+	/**
+	 * @param string $group
+	 */
+	protected function set_group( $group ) {
+		$this->group = $group;
+	}
+
+	/**
+	 * @return string
+	 */
+	public function get_group() {
+		return $this->group;
+	}
+
+	/**
+	 * @return bool If the action has been finished
+	 */
+	public function is_finished() {
+		return $this->is_finished;
+	}
+}
--- a/classes/ActionScheduler_AbstractAction.php
+++ b/classes/ActionScheduler_AbstractAction.php
@@ -24,7 +24,7 @@ abstract class ActionScheduler_AbstractAction implements ActionScheduler_Interfa
 	protected $is_finished = false;
 
 	public function execute() {
-		return do_action_ref_array( $this->get_hook(), $this->get_args() );
+		do_action_ref_array( $this->get_hook(), $this->get_args() );
 	}
 
 	/**
--- a/classes/ActionScheduler_ActionFactory.php
+++ b/classes/ActionScheduler_ActionFactory.php
@@ -4,6 +4,39 @@
  * Class ActionScheduler_ActionFactory
  */
 class ActionScheduler_ActionFactory {
+
+	/**
+	 * @param int $action_id The action's ID in the data store
+	 * @param string $hook The hook to trigger when this action runs
+	 * @param string $status The action's status in the data store
+	 * @param string $claim_id The identifier for the claim this action belongs to, if any, derived from ActionScheduler_ActionClaim::get_id()
+	 * @param array $args Args to pass to callbacks when the hook is triggered
+	 * @param ActionScheduler_Schedule $schedule The action's schedule
+	 * @param string $group A group to put the action in
+	 *
+	 * @return ActionScheduler_StoredAction An instance of the stored action
+	 */
+	public function get_stored_action_instance( $action_id, $hook, $status, $claim_id = '', array $args = array(), ActionScheduler_Schedule $schedule = NULL, $group = '' ) {
+
+		switch ( $status ) {
+			case ActionScheduler_Store::STATUS_PENDING :
+				$action_class = 'ActionScheduler_StoredAction';
+				break;
+			case ActionScheduler_Store::STATUS_CANCELED :
+				$action_class = 'ActionScheduler_CanceledAction';
+				break;
+			default :
+				$action_class = 'ActionScheduler_NonexecutableAction';
+				break;
+		}
+
+		$action_class = apply_filters( 'action_scheduler_stored_action_class', $action_class, $action_id, $hook, $status, $claim_id, $args, $schedule, $group );
+
+		$action = new $action_class( $action_id, $hook, $status, $claim_id, $args, $schedule, $group );
+
+		return apply_filters( 'action_scheduler_stored_action_instance', $action, $action_id, $hook, $status, $claim_id, $args, $schedule, $group )
+	}
+
 	/**
 	 * @param string $hook The hook to trigger when this action runs
 	 * @param array $args Args to pass when the hook is triggered
--- a/classes/ActionScheduler_CanceledAction.php
+++ b/classes/ActionScheduler_CanceledAction.php
@@ -6,10 +6,10 @@
  * Stored action which was canceled and therefore acts like a finished action but should always return a null schedule,
  * regardless of schedule passed to its constructor.
  */
-class ActionScheduler_CanceledAction extends ActionScheduler_FinishedAction {
+class ActionScheduler_CanceledAction extends ActionScheduler_NonexecutableAction {
 
-	public function __construct( $hook, array $args = array(), ActionScheduler_Schedule $schedule = NULL, $group = '', $id = NULL, $status = '', $claim_id = '' ) {
-		parent::__construct( $hook, $args, $schedule, $group );
-		$this->set_schedule( new ActionScheduler_NullSchedule() );
+	public function __construct( $id, $hook, $status, $claim_id = '', array $args = array(), ActionScheduler_Schedule $schedule = NULL, $group = '' ) {
+		$null_schedule = new ActionScheduler_NullSchedule();
+		parent::__construct( $id, $hook, $status, $claim_id, $args, $null_schedule, $group );
 	}
 }
--- a/classes/ActionScheduler_FinishedAction.php
+++ b/classes/ActionScheduler_FinishedAction.php
@@ -3,16 +3,19 @@
 /**
  * Class ActionScheduler_FinishedAction
  *
- * Stored action which has finished, which could be because it was cancelled, completed or failed.
+ * Deprecated to avoid ambiguity with stored actions and objects instantiated purely to call the hook.
+ *
+ * @deprecated 1.6.0
  */
-class ActionScheduler_FinishedAction extends ActionScheduler_StoredAction {
+class ActionScheduler_FinishedAction extends ActionScheduler_Action {
+
+	public function __construct( $hook, array $args = array(), ActionScheduler_Schedule $schedule = NULL, $group = '' ) {
+		trigger_error( sprintf( __('%1$s is <strong>deprecated</strong> since version 1.6.0! Use an instance of ActionScheduler_NonexecutableAction instead.'), __CLASS__ ) );
+		parent::__construct( $hook, $args, $schedule, $group );
+		$this->is_finished = true;
+	}
 
 	public function execute() {
 		// don't execute
 	}
-
-	public function is_finished() {
-		return TRUE;
-	}
 }
- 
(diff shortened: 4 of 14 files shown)


classes/ActionScheduler_ActionFactory.php
```diff
@@ -4,6 +4,31 @@
...
+	 * @param string $claim_id The identifier for the claim this action belongs to, if any, derived from ActionScheduler_ActionClaim::get_id()
+	 * @param array $args Args to pass to callbacks when the hook is triggered
+	 * @param ActionScheduler_Schedule $schedule The action's schedule
+	 * @param string $group A group to put the action in
+	 *
+	 * @return string The ID of the stored action
+	 */
+	public function get_stored_action_instance( $action_id, $hook, $status, $claim_id = '', array $args = array(), ActionScheduler_Schedule $schedule = NULL, $group = '' ) {
+
+		if ( ActionScheduler_Store::STATUS_PENDING === $status ) {
```
JPry: Seeing this here makes me think that perhaps it would be better to have the various status constants in an interface, rather than an abstract class. But I recognize also that we may not really want to attempt to change that right now, so consider this a public note rather than a request for a change.

With that said, would it make sense here to use a `switch` statement instead of if/elseif/else? Or perhaps to add a filter to the final class string for extensibility? Those are real questions, not necessarily change requests.

classes/ActionScheduler_CanceledAction.php
```diff
@@ -6,10 +6,10 @@
  * Stored action which was canceled and therefore acts like a finished action but should always return a null schedule,
  * regardless of schedule passed to its constructor.
  */
-class ActionScheduler_CanceledAction extends ActionScheduler_FinishedAction {
+class ActionScheduler_CanceledAction extends ActionScheduler_NonexecutableAction {
 
-	public function __construct( $hook, array $args = array(), ActionScheduler_Schedule $schedule = NULL, $group = '', $id = NULL, $status = '', $claim_id = '' ) {
-		parent::__construct( $hook, $args, $schedule, $group );
-		$this->set_schedule( new ActionScheduler_NullSchedule() );
+	public function __construct( $id, $hook, $status, $claim_id = '', array $args = array(), ActionScheduler_Schedule $schedule = NULL, $group = '' ) {
```
JPry: Sine we're not using the `$schedule` variable at all, it should be removed from the constructor.

classes/ActionScheduler_Interface_Storable.php
```diff
@@ -0,0 +1,26 @@
+<?php
+/**
+ * Interface ActionScheduler_Interface_Action
```
JPry: Copy pasta 🙂

classes/ActionScheduler_FinishedAction.php
```diff
@@ -3,16 +3,19 @@
...
+	public function __construct( $hook, array $args = array(), ActionScheduler_Schedule $schedule = NULL, $group = '' ) {
+		trigger_error( sprintf( __('%1$s is <strong>deprecated</strong> since version 1.6.0! Use an instance of ActionScheduler_NonexecutableAction instead.'), __CLASS__ ) );
+		parent::__construct( $hook, $args, $schedule, $group );
+		$this->is_finished = true;
 	}
 
-	public function is_finished() {
-		return TRUE;
+	public function execute() {
+		// don't execute
```
JPry: With an additional if statement I mentioned in the `execute()` method above, we wouldn't need to override the method here.

classes/ActionScheduler_StoredAction.php
```diff
@@ -5,22 +5,26 @@
  *
  * An action which has been saved in the data store and is pending execution (i.e. not finished).
  */
-class ActionScheduler_StoredAction extends ActionScheduler_Action {
+class ActionScheduler_StoredAction extends ActionScheduler_AbstractAction implements ActionScheduler_Interface_Storable, ActionScheduler_Interface_Claimable {
```
JPry: This should be `ActionScheduler_Stored_Action` for consistency. Assuming it's possible rename the class at this point? If not, we can leave it for another day.

classes/ActionScheduler_NonexecutableAction.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+/**
+ * Class ActionScheduler_NonexecutableAction
+ *
+ * Stored action which should not be executed because it has already been executed, and has the
+ * failed or completed status, or becuase it has been cancelled, and has the cancelled status.
+ */
+class ActionScheduler_NonexecutableAction extends ActionScheduler_StoredAction {
```
JPry: This should be `ActionScheduler_Nonexecutable_Action` for consistency.

classes/ActionScheduler_FinishedAction.php
```diff
@@ -3,16 +3,19 @@
 /**
  * Class ActionScheduler_FinishedAction
  *
- * Stored action which has finished, which could be because it was cancelled, completed or failed.
+ * Deprecated to avoid ambiguity with stored actions and objects instantiated purely to call the hook.
+ *
+ * @deprecated 1.6.0
  */
-class ActionScheduler_FinishedAction extends ActionScheduler_StoredAction {
+class ActionScheduler_FinishedAction extends ActionScheduler_Action {
```
JPry: If possible, this should be `ActionScheduler_Finished_Action` for consistency.

classes/ActionScheduler_CanceledAction.php
```diff
@@ -6,10 +6,10 @@
  * Stored action which was canceled and therefore acts like a finished action but should always return a null schedule,
  * regardless of schedule passed to its constructor.
  */
-class ActionScheduler_CanceledAction extends ActionScheduler_FinishedAction {
+class ActionScheduler_CanceledAction extends ActionScheduler_NonexecutableAction {
```
JPry: If possible, this should be `ActionScheduler_Canceled_Action` for consistency.

classes/ActionScheduler_NullAction.php
```diff
@@ -3,14 +3,14 @@
 /**
  * Class ActionScheduler_NullAction
  */
-class ActionScheduler_NullAction extends ActionScheduler_Action {
+class ActionScheduler_NullAction extends ActionScheduler_AbstractAction {
```
JPry: If possible, this should be `ActionScheduler_Null_Action` for consistency.

classes/ActionScheduler_StoredAction.php
```diff
@@ -5,22 +5,26 @@
...
 	/** @var string */
-	protected $status = '';
+	protected $status;
 
 	/** @var string */
 	protected $claim_id = '';
 
-	public function __construct( $hook, array $args = array(), ActionScheduler_Schedule $schedule = NULL, $group = '', $id = NULL, $status = '', $claim_id = '' ) {
-		parent::__construct( $hook, $args, $schedule, $group );
+	public function __construct( $id, $hook, $status, $claim_id = '', array $args = array(), ActionScheduler_Schedule $schedule = NULL, $group = '' ) {
```
JPry: `null` should be lowercase.

classes/ActionScheduler_wpPostStore.php
```diff
@@ -11,7 +11,7 @@ class ActionScheduler_wpPostStore extends ActionScheduler_Store {
 	/** @var DateTimeZone */
 	protected $local_timezone = NULL;
 
-	public function save_action( ActionScheduler_Action $action, DateTime $date = NULL ){
+	public function save_action( ActionScheduler_AbstractAction $action, DateTime $date = NULL ){
```
JPry: If possible it would be better to type hint for an interface rather than a class, even an abstract class. With that said, that may not be as feasible without a bunch of rewriting, so if it won't be easy then we can keep it as is.

Also, `null` should be lowercase.

classes/ActionScheduler_AbstractAction.php
```diff
@@ -0,0 +1,81 @@
...
+	protected $schedule = NULL;
+
+	/** @var string */
+	protected $group = '';
+
+	/** @var bool */
+	protected $is_finished = false;
+
+	public function execute() {
+		return do_action_ref_array( $this->get_hook(), $this->get_args() );
```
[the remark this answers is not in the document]
thenbrent: > `do_action_ref_array()` doesn't actually return anything, so we could eliminate the preceding return here.

I've changed that with SHA: cd6c1ac It was carried over from `ActionScheduler_Action`, but seeing as we're introducing the changeset, now seems as good a time as any to fix that up.

> should the call to `do_action_ref_array()` be preceded with `if ( ! $this->is_finished )`, followed by setting `$this->is_finished` to true? Or have I missed where the `$is_finished` property is used elsewhere?

I toyed with that idea, and while it's correct, it would be a breaking change. Previously, calling `ActionScheduler_Action->execute()` twice consecutively would execute it twice. If we introduced that change, it 
would only be executed once (there are valid reasons why someone might do that, like testing an action for fatal errors before rolling back, albeit unlikely to actually be being done by anyone).

I also toyed with deprecating `execute()` and instead using a new `run()` method which could implement that, and/or update the action's internal status once it had run rather than it relying on `ActionScheduler_QueueRunner::process_action()` to call `ActionScheduler_QueueRunner->store->mark_complete()` to do it.

Really though, `is_finished()` and the `$is_finished` property are only used when saving an action's status now so it could also theoretically be removed entirely and the status property used instead.

For now, I've kept it simple, but if we're going to change AS to be v2.0, we could go for more extreme breaking changes like this.

classes/ActionScheduler_ActionFactory.php
```diff
@@ -4,6 +4,31 @@
...
+	/**
+	 * @param int $action_id The action's ID in the data store
+	 * @param string $hook The hook to trigger when this action runs
+	 * @param string $status The action's status in the data store
+	 * @param string $claim_id The identifier for the claim this action belongs to, if any, derived from ActionScheduler_ActionClaim::get_id()
+	 * @param array $args Args to pass to callbacks when the hook is triggered
+	 * @param ActionScheduler_Schedule $schedule The action's schedule
+	 * @param string $group A group to put the action in
+	 *
+	 * @return string The ID of the stored action
```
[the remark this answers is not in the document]
thenbrent: Fixed and rebased, thanks!

classes/ActionScheduler_ActionFactory.php
```diff
@@ -4,6 +4,31 @@
...
+	 */
+	public function get_stored_action_instance( $action_id, $hook, $status, $claim_id = '', array $args = array(), ActionScheduler_Schedule $schedule = NULL, $group = '' ) {
```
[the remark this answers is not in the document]
thenbrent: I'd prefer this PR maintains consistency with existing instances and a new PR/patch is introduced to change all instances at once.

classes/ActionScheduler_ActionFactory.php
thenbrent: A `switch` makes sense because it reduces the changeset if we introduce additional status based classes in future. I think a filter makes sense also to allow 3rd parties to add their own classes too. I've changed both of those and rebased in the the changes.

classes/ActionScheduler_CanceledAction.php
thenbrent: Consistency with what? No other class in Action Scheduler uses an underscore to separate a word and action type. 🙂

classes/ActionScheduler_CanceledAction.php
thenbrent: @JPry this goes back to the old issues originally raised on #88 around having child classes with different constructor param sets. `$schedule` can't be removed without breaking polymorphism of other stored action constructors. e.g. we'd need to update `ActionScheduler_ActionFactory::get_stored_action_instance()` to pass special params for cancelled actions vs. other stored action types. It feels like there should be a neater/cleaner solution to this, but I can't see it yet.

classes/ActionScheduler_FinishedAction.php
thenbrent: @JPry I think you may be confusing Action Scheduler's consistency with WP standards and/or Subscriptions? See other comments about use of underscores in AS. 🙂

classes/ActionScheduler_FinishedAction.php
thenbrent: That's true, and possibly a good justification for going with a new `run()` method and/or deprecating `execute()` as discussed in the other comment about maintaining backward compatibility.

classes/ActionScheduler_NonexecutableAction.php
thenbrent: @JPry see other responses RE: consistency.

classes/ActionScheduler_NullAction.php
thenbrent: `ActionScheduler_NullAction` has been in AS since v1.0. It _is_ the base for what others should be consistent with. 🙂

classes/ActionScheduler_wpPostStore.php
thenbrent: > With that said, that may not be as feasible without a bunch of rewriting

I don't think it would require any rewriting if we used `ActionScheduler_Interface_Action`. It would still maintain backward compatibility as `ActionScheduler_AbstractAction` implements it, and `ActionScheduler_Action` extends `ActionScheduler_AbstractAction`.

Can you share a good article or two explaining why it's better to help me learn?

thenbrent:
@JPry (and anyone else). Hold off on any further review on this. I'm going to explore moving more of this functionality back into the datastore to see if it will provide a cleaner solution to the issues in #88.

thenbrent:
Closing this now in favour of #113, which returns to the older approach of accessing the data directly on the datastore. I'll leave the branch here for now, and delete it only after #113 is merged.

PR #113, opened 2018-03-12
Add new Admin UI using data store instead of post APIs

### LINKED ISSUE #85: Update admin view to use data stores

thenbrent:
This is a redo of PR #88 for Issue #85. I'm opening a new PR as #88 has become unwieldy in its history. This PR has merged `1-6-dev` (I couldn't rebase without losing subtree data for the pp list table), and adds additional commits to add sortable columns, display Claim ID and a few other patches to bring the admin UI closer to completeness. The only feature remaining for it to have feature parity with the old admin UI and close #85 is a _Search_ box.

This PR also negates #110, which attempted to solve the problem of accessing an actions data by moving that data into the action itself. The changes required to achieve that are excessive, so I've opted to return to the approach of using the data store for accessing that data, as it is much cleaner for now. It also avoids issues like whether data on the action should be changed, e.g. if it's a _pending_ action's status should be changed to _completed_ after `execute()` is called - something which makes sense, but which is an entirely new paradigm for Action Scheduler, and which may have unforeseen implications.

This PR is ready for review, but I'd encourage you to avoid reviewing the commit history, as it's still unwieldy.

### DIFF
--- a/classes/ActionScheduler_ActionFactory.php
+++ b/classes/ActionScheduler_ActionFactory.php
@@ -14,7 +14,7 @@ class ActionScheduler_ActionFactory {
 	 *
 	 * @return ActionScheduler_StoredAction An instance of the stored action
 	 */
-	public function get_stored_action( $status, $hook, array $args = array(), ActionScheduler_Schedule $schedule = NULL, $group = '' ) {
+	public function get_stored_action( $status, $hook, array $args = array(), ActionScheduler_Schedule $schedule = null, $group = '' ) {
 
 		switch ( $status ) {
 			case ActionScheduler_Store::STATUS_PENDING :
@@ -52,7 +52,7 @@ class ActionScheduler_ActionFactory {
 	 *
 	 * @return string The ID of the stored action
 	 */
-	public function single( $hook, $args = array(), $when = NULL, $group = '' ) {
+	public function single( $hook, $args = array(), $when = null, $group = '' ) {
 		$date = as_get_datetime_object( $when );
 		$schedule = new ActionScheduler_SimpleSchedule( $date );
 		$action = new ActionScheduler_Action( $hook, $args, $schedule, $group );
@@ -68,7 +68,7 @@ class ActionScheduler_ActionFactory {
 	 *
 	 * @return string The ID of the stored action
 	 */
-	public function recurring( $hook, $args = array(), $first = NULL, $interval = NULL, $group = '' ) {
+	public function recurring( $hook, $args = array(), $first = null, $interval = null, $group = '' ) {
 		if ( empty($interval) ) {
 			return $this->single( $hook, $args, $first, $group );
 		}
@@ -88,7 +88,7 @@ class ActionScheduler_ActionFactory {
 	 *
 	 * @return string The ID of the stored action
 	 */
-	public function cron( $hook, $args = array(), $first = NULL, $schedule = NULL, $group = '' ) {
+	public function cron( $hook, $args = array(), $first = null, $schedule = null, $group = '' ) {
 		if ( empty($schedule) ) {
 			return $this->single( $hook, $args, $first, $group );
 		}
--- a/classes/ActionScheduler_ActionFactory.php
+++ b/classes/ActionScheduler_ActionFactory.php
@@ -28,7 +28,7 @@ class ActionScheduler_ActionFactory {
 				break;
 		}
 
-		$action_class = apply_filters( 'action_scheduler_stored_action_class', $action_class, $hook, $args, $schedule, $group );
+		$action_class = apply_filters( 'action_scheduler_stored_action_class', $action_class, $status, $hook, $args, $schedule, $group );
 
 		$action = new $action_class( $hook, $args, $schedule, $group );
 
--- a/classes/ActionScheduler_ActionFactory.php
+++ b/classes/ActionScheduler_ActionFactory.php
@@ -12,7 +12,7 @@ class ActionScheduler_ActionFactory {
 	 * @param ActionScheduler_Schedule $schedule The action's schedule
 	 * @param string $group A group to put the action in
 	 *
-	 * @return ActionScheduler_StoredAction An instance of the stored action
+	 * @return ActionScheduler_Action An instance of the stored action
 	 */
 	public function get_stored_action( $status, $hook, array $args = array(), ActionScheduler_Schedule $schedule = null, $group = '' ) {
 
@@ -109,4 +109,3 @@ class ActionScheduler_ActionFactory {
 		return $store->save_action( $action );
 	}
 }
- 
--- a/classes/ActionScheduler_ActionFactory.php
+++ b/classes/ActionScheduler_ActionFactory.php
@@ -32,6 +32,15 @@ class ActionScheduler_ActionFactory {
 
 		$action = new $action_class( $hook, $args, $schedule, $group );
 
+		/**
+		 * Allow 3rd party code to change the instantiated action for a given hook, args, schedule and group.
+		 *
+		 * @param ActionScheduler_Action $action The instantiated action.
+		 * @param string $hook The instantiated action's hook.
+		 * @param array $args The instantiated action's args.
+		 * @param ActionScheduler_Schedule $schedule The instantiated action's schedule.
+		 * @param string $group The instantiated action's group.
+		 */
 		return apply_filters( 'action_scheduler_stored_action_instance', $action, $hook, $args, $schedule, $group );
 	}
 
--- a/classes/ActionScheduler_AdminView.php
+++ b/classes/ActionScheduler_AdminView.php
@@ -70,7 +70,7 @@ class ActionScheduler_AdminView {
 	 * Renders the Admin UI
 	 */
 	public function render_admin_ui() {
-		$table = new ActionScheduler_ListTable( ActionScheduler::store(), ActionScheduler::logger() );
+		$table = new ActionScheduler_ListTable( ActionScheduler::store(), ActionScheduler::logger(), ActionScheduler::runner() );
 		$table->prepare_items();
 
 		echo '<h1>' . __( 'Scheduled Actions', 'action-scheduler' ) . '</h1>';
--- a/classes/ActionScheduler_AdminView.php
+++ b/classes/ActionScheduler_AdminView.php
@@ -75,12 +75,7 @@ class ActionScheduler_AdminView {
 	 */
 	public function render_admin_ui() {
 		$table = new ActionScheduler_ListTable( ActionScheduler::store(), ActionScheduler::logger(), ActionScheduler::runner() );
-		$table->prepare_items();
-
-		echo '<div class="wrap">';
-		echo '<h1>' . __( 'Scheduled Actions', 'action-scheduler' ) . '</h1>';
-		$table->display();
-		echo '</div>';
+		$table->display_page();
 	}
 
 	/** Deprecated Functions **/
--- a/classes/ActionScheduler_AdminView.php
+++ b/classes/ActionScheduler_AdminView.php
@@ -26,7 +26,7 @@ class ActionScheduler_AdminView {
 	 * @codeCoverageIgnore
 	 */
 	public function init() {
-		$self = self::instance();
+		$self = $this->instance();
 
 		if ( is_admin() && ( ! defined( 'DOING_AJAX' ) || false == DOING_AJAX ) ) {
 
--- a/classes/ActionScheduler_AdminView.php
+++ b/classes/ActionScheduler_AdminView.php
@@ -32,7 +32,7 @@ class ActionScheduler_AdminView {
 
 			if ( class_exists( 'WooCommerce' ) ) {
 				add_action( 'woocommerce_admin_status_content_action-scheduler', array( $self, 'render_admin_ui' ) );
-				add_filter( 'woocommerce_admin_status_tabs', array( $self, 'register_menu_woo_tab' ) );
+				add_filter( 'woocommerce_admin_status_tabs', array( $self, 'register_system_status_tab' ) );
 			}
 
 			add_action( 'admin_menu', array( $self, 'register_menu' ) );
@@ -43,7 +43,7 @@ class ActionScheduler_AdminView {
 	/**
 	 * Registers action-scheduler into WooCommerce > System status.
 	 */
-	public function register_menu_woo_tab( array $tabs ) {
+	public function register_system_status_tab( array $tabs ) {
 		$tabs[ 'action-scheduler' ] = __( 'Scheduled Actions', 'action-scheduler' );
 
 		return $tabs;
--- a/classes/ActionScheduler_AdminView.php
+++ b/classes/ActionScheduler_AdminView.php
@@ -26,16 +26,14 @@ class ActionScheduler_AdminView {
 	 * @codeCoverageIgnore
 	 */
 	public function init() {
-		$self = $this->instance();
-
 		if ( is_admin() && ( ! defined( 'DOING_AJAX' ) || false == DOING_AJAX ) ) {
 
 			if ( class_exists( 'WooCommerce' ) ) {
-				add_action( 'woocommerce_admin_status_content_action-scheduler', array( $self, 'render_admin_ui' ) );
-				add_filter( 'woocommerce_admin_status_tabs', array( $self, 'register_system_status_tab' ) );
+				add_action( 'woocommerce_admin_status_content_action-scheduler', array( $this, 'render_admin_ui' ) );
+				add_filter( 'woocommerce_admin_status_tabs', array( $this, 'register_system_status_tab' ) );
 			}
 
-			add_action( 'admin_menu', array( $self, 'register_menu' ) );
+			add_action( 'admin_menu', array( $this, 'register_menu' ) );
 		}
 	}
 
--- a/classes/ActionScheduler_AdminView.php
+++ b/classes/ActionScheduler_AdminView.php
@@ -42,18 +42,22 @@ class ActionScheduler_AdminView {
 
 	/**
 	 * Registers action-scheduler into WooCommerce > System status.
+	 *
+	 * @param array $tabs An associative array of tab key => label.
+	 * @return array $tabs An associative array of tab key => label, including Action Scheduler's tabs
 	 */
 	public function register_system_status_tab( array $tabs ) {
-		$tabs[ 'action-scheduler' ] = __( 'Scheduled Actions', 'action-scheduler' );
+		$tabs['action-scheduler'] = __( 'Scheduled Actions', 'action-scheduler' );
 
 		return $tabs;
 	}
 
 	/**
-	 * Registers action-scheduler under Tools.
+	 * Include Action Scheduler's administration under the Tools menu.
 	 *
-	 * This method is called if woocommerce is not activated and we can't register our page
-	 * to WooCommerce's System status page.
+	 * A menu under the Tools menu is important for backward compatibility (as that's
+	 * where it started), and also provides more convenient access than the WooCommerce
+	 * System Status page, and for sites where WooCommerce isn't active.
 	 */
 	public function register_menu() {
 		add_submenu_page(
--- a/classes/ActionScheduler_AdminView.php
+++ b/classes/ActionScheduler_AdminView.php
@@ -123,6 +123,8 @@ class ActionScheduler_AdminView {
 	 */
 	public function list_table_columns( $columns ) {
 		_deprecated_function( __METHOD__, '1.6' );
+
+		return $columns;
 	}
 
 	/**
--- a/classes/ActionScheduler_AdminView.php
+++ b/classes/ActionScheduler_AdminView.php
@@ -29,12 +29,13 @@ class ActionScheduler_AdminView {
 		$self = self::instance();
 
 		if ( is_admin() && ( ! defined( 'DOING_AJAX' ) || false == DOING_AJAX ) ) {
+
 			if ( class_exists( 'WooCommerce' ) ) {
 				add_action( 'woocommerce_admin_status_content_action-scheduler', array( $self, 'render_admin_ui' ) );
 				add_filter( 'woocommerce_admin_status_tabs', array( $self, 'register_menu_woo_tab' ) );
-			} else {
-				add_action( 'admin_menu', array( $self, 'register_menu' ) );
 			}
+
+			add_action( 'admin_menu', array( $self, 'register_menu' ) );
 		}
 	}
 
--- a/classes/ActionScheduler_AdminView.php
+++ b/classes/ActionScheduler_AdminView.php
@@ -77,8 +77,10 @@ class ActionScheduler_AdminView {
 		$table = new ActionScheduler_ListTable( ActionScheduler::store(), ActionScheduler::logger(), ActionScheduler::runner() );
 		$table->prepare_items();
 
+		echo '<div class="wrap">';
 		echo '<h1>' . __( 'Scheduled Actions', 'action-scheduler' ) . '</h1>';
 		$table->display();
+		echo '</div>';
 	}
 
 	/** Deprecated Functions **/
--- a/classes/ActionScheduler_ListTable.php
+++ b/classes/ActionScheduler_ListTable.php
@@ -400,7 +400,7 @@ class ActionScheduler_ListTable extends PP_List_Table {
 
 		$status_list_items = array();
 
-		foreach ( $this->store->actions_count() as $status_name => $count ) {
+		foreach ( $this->store->action_counts() as $status_name => $count ) {
 
 			if ( 0 === $count ) {
 				continue;
--- a/classes/ActionScheduler_ListTable.php
+++ b/classes/ActionScheduler_ListTable.php
@@ -349,6 +349,7 @@ class ActionScheduler_ListTable extends PP_List_Table {
 
 		$this->items = array();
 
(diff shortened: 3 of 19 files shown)


thenbrent:
@JPry I've realised there's still a bit more work to do here, so there will be some rebasing and force-pushing going on today. Feel free to ignore for another day if you want to avoid seeing commits come-and-go. 🙂

thenbrent:
@JPry this is now ready for review. 🙂

classes/ActionScheduler_LogEntry.php
```diff
@@ -4,12 +4,31 @@
  * Class ActionScheduler_LogEntry
  */
 class ActionScheduler_LogEntry {
-	protected $action_id = '';
-	protected $message = '';
+	protected $action_id =  '';
+	protected $message =  '';
+	protected $date;
```
JPry: These should have docblocks including `@var` tags.

classes/ActionScheduler_LogEntry.php
```diff
@@ -4,12 +4,31 @@
...
 
-	public function __construct( $action_id, $message ) {
+	/**
+	 * Constructor
+	 *
+	 * @param mixed  $action_id	Action ID
```
JPry: There should be a space instead of a tab between `$action_id` and `Action ID`. Also, is there a reason the type is "mixed" instead of "int"?

classes/ActionScheduler_LogEntry.php
```diff
@@ -4,12 +4,31 @@
...
+	/**
+	 * Returns the date when this log entry was created
+	 *
+	 * @return Datetime
+	 */
+	public function get_date() {
+		return $this->date;
 	}
 
 	public function get_action_id() {
```
JPry: Let's go ahead and add docblocks to the other methods here.

classes/ActionScheduler_QueueRunner.php
```diff
@@ -44,7 +44,7 @@ public function run() {
 		do_action( 'action_scheduler_before_process_queue' );
 		$this->run_cleanup();
 		$count = 0;
-		if ( $this->store->get_claim_count() < apply_filters( 'action_scheduler_queue_runner_concurrent_batches', 5 ) ) {
+		if ( $this->store->get_claim_count() < $this->get_concurrent_batch_count() ) {
```
JPry: 🎉

classes/ActionScheduler_QueueRunner.php
```diff
@@ -53,6 +53,10 @@ public function run() {
 		return $count;
 	}
 
+	public function get_concurrent_batch_count() {
```
JPry: Would it make sense to move this to the `ActionScheduler_Abstract_QueueRunner` class? I think other queue runners could benefit from it.

classes/ActionScheduler_Store.php
```diff
@@ -120,12 +127,37 @@
 	 */
 	abstract public function get_status( $action_id );
 
+	/**
+	 * @return mixed
+	 */
+	abstract public function get_claim_id( $action_id );
```
JPry: This needs the parameter included in the docblock. Also, should this be a defined type, such as an int?

classes/ActionScheduler_Store.php
```diff
@@ -120,12 +127,37 @@
...
+
 	/**
 	 * @param string $claim_id
 	 * @return array
 	 */
 	abstract public function find_actions_by_claim_id( $claim_id );
 
+	protected function validate_sql_comparator( $comp ) {
```
JPry: This method should include a docblock

classes/ActionScheduler_Store.php
```diff
@@ -120,12 +127,37 @@
...
+		if ( in_array($comp, array('!=', '>', '>=', '<', '<=', '=')) ) {
```
JPry: This needs some formatting adjustments to meet with WP standards.

I think it would be preferable to use a key-based array, but it also looks like this will typically only be used a couple times during a given request, so the performance benefit would be minimal. Meaning, it's also OK to leave as-is 🙂

classes/ActionScheduler_wpPostStore.php
```diff
@@ -93,7 +93,7 @@ protected function save_action_group( $post_id, $group ) {
 
 	public function fetch_action( $action_id ) {
 		$post = $this->get_post( $action_id );
-		if ( empty($post) || $post->post_type != self::POST_TYPE || $post->post_status == 'trash' ) {
+		if ( empty($post) || $post->post_type != self::POST_TYPE ) {
```
JPry: Since this line is already being modified, the `empty($post)` statement should be formatted for WP standards

classes/ActionScheduler_wpPostStore.php
```diff
@@ -119,12 +119,64 @@ protected function make_action_from_post( $post ) {
...
+	/**
+	 * @param string $post_status
+	 *
+	 * @return string
+	 */
+	protected function get_action_status_by_post_status( $post_status ) {
+
+		switch ( $post_status ) {
+			case 'any' :
+				$post_status = 'all';
```
JPry: Copy/pasta? Should this be `$action_status`?

classes/ActionScheduler_wpPostStore.php
```diff
@@ -119,12 +119,64 @@ protected function make_action_from_post( $post ) {
...
-		} else {
-			$action = new ActionScheduler_FinishedAction( $hook, $args, $schedule, $group );
+
+		return ActionScheduler::factory()->get_stored_action( $this->get_action_status_by_post_status( $post->post_status ), $hook, $args, $schedule, $group );
+	}
+
+	/**
+	 * @param string $post_status
+	 *
+	 * @return string
```
JPry: This should include an `@throws` tag indicating what causes the `InvalidArgumentException` to be thrown.

classes/ActionScheduler_wpPostStore.php
```diff
@@ -119,12 +119,64 @@ protected function make_action_from_post( $post ) {
...
 		}
-		return $action;
+
+		return $action_status;
+	}
+
+	/**
+	 * @param string $action_status
+	 *
+	 * @return string
```
JPry: This should include an `@throws` tag indicating what causes the `InvalidArgumentException` to be thrown.

classes/ActionScheduler_wpPostStore.php
```diff
@@ -207,7 +261,8 @@ public function query_actions( $query = array() ) {
 
 		/** @var wpdb $wpdb */
 		global $wpdb;
-		$sql = "SELECT p.ID FROM {$wpdb->posts} p";
+		$sql  = ( 'count' === $select_or_count ) ? 'SELECT count(p.ID)' : 'SELECT p.ID ';
```
JPry: I think it would be a good idea to validate the value of the `$select_or_count` variable, since there are only 2 valid values.

classes/ActionScheduler_wpPostStore.php
```diff
@@ -265,44 +313,84 @@ public function query_actions( $query = array() ) {
...
+	 * @param string $query_type Whether to select or count the results. Default, select.
+	 * @return array The IDs of actions matching the query
+	 */
+	public function query_actions( $query = array(), $query_type = 'select' ) {
+		/** @var wpdb $wpdb */
+		global $wpdb;
+
+		$sql = $this->get_query_actions_sql( $query, $query_type );
+
+		return ( 'count' === $query_type ) ? $wpdb->get_var( $sql ) : $wpdb->get_col( $sql );
```
JPry: `$wpdb->get_var()` returns a string, not an array. This should either be converted to an array, or the docblock should be updated to indicate that it can return either type.

classes/ActionScheduler_wpPostStore.php
```diff
@@ -466,30 +554,24 @@ public function mark_failure( $action_id ) {
 		}
 	}
 
+	public function get_claim_id( $action_id ) {
```
JPry: This should include a docblock

classes/ActionScheduler_wpPostStore.php
```diff
@@ -466,30 +554,24 @@ public function mark_failure( $action_id ) {
...
+		return $this->get_post_column( $action_id, 'post_password' );
+	}
+
```
JPry: Let's add a docblock here too, including an `@throws` tag.

classes/ActionScheduler_wpPostStore.php
```diff
@@ -466,30 +554,24 @@ public function mark_failure( $action_id ) {
...
+		$status = $this->get_post_column( $action_id, 'post_status' );
+
+		if ( $status === null ) {
+			throw new \InvalidArgumentException( __( 'Invalid action ID. No status found.', 'action-scheduler' ) );
 		}
 
+		return $this->get_action_status_by_post_status( $status );
+	}
+
+	private function get_post_column( $action_id, $column_name ) {
```
JPry: This should include a docblock.

While this is private and won't be used outside of this class, it may be worth considering whether we want to do any kind of validation on the column name.

classes/ActionScheduler_AdminView.php
```diff
@@ -28,42 +26,63 @@ public static function instance() {
...
-		add_action( 'admin_notices', array( self::instance(), 'admin_notices' ) );
+		return $tabs;
+	}
 
-		add_filter( 'post_updated_messages', array( self::instance(), 'post_updated_messages' ) );
+	/**
+	 * Registers action-scheduler under Tools.
+	 *
+	 * This method is called if woocommerce is not activated and we can't register our page
+	 * to WooCommerce's System status page.
```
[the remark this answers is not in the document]
thenbrent: A good example of why docblocks should explain _why_ a method exists and/or does what it does, not _what_ it does. 👍

classes/ActionScheduler_wpPostStore.php
thenbrent: I'm going to remove this method instead and use `get_post_field()` in its place. That gets us a bit of extra WP magic sauce, even if it will be less efficient than a `$wpdb` query.

classes/ActionScheduler_ActionFactory.php
```diff
@@ -4,6 +4,37 @@
...
+			default :
+				$action_class = 'ActionScheduler_FinishedAction';
+				break;
+		}
+
+		$action_class = apply_filters( 'action_scheduler_stored_action_class', $action_class, $hook, $args, $schedule, $group );
+
+		$action = new $action_class( $hook, $args, $schedule, $group );
+
+		return apply_filters( 'action_scheduler_stored_action_instance', $action, $hook, $args, $schedule, $group );
```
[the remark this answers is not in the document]
thenbrent: I've added one with SHA: 2a5dca27cc86e7892dd988e13732e51d6f448699. 

It feels quite redundant amidst the code though, because variable names, type declarations on the method and general understanding of what `apply_filters()` does, it feels like the docblock is just repeating the code.

Do you want to see these mostly for use with generated documentation? (i.e. docs outside the code)

classes/ActionScheduler_wpPostStore.php
thenbrent: @JPry my personal preference is generally to maintain internal consistency rather than to apply progressive enhancement. That way, when you're working with a code base, you're not being presented with multiple competing standards. For me, that avoids interrupting my thinking on the more important parts of the code.

I'm OK with doing a widespread standards update on AS though (unlike WP core). It's not really something I feel the need to do, but if you want to do it, I'll happily review/merge the PR.

classes/ActionScheduler_QueueRunner.php
thenbrent: @JPry yep for sure. I'll also rename it to `get_allowed_concurrent_batches()` to avoid any ambiguity - the current `get_concurrent_batch_count()` could be confused with returning the number of batches in-progress rather than the maximum number allowed to run concurrently.

classes/ActionScheduler_LogEntry.php
thenbrent: I've opened https://github.com/Prospress/action-scheduler/issues/125 to address this.

Code review isn't the time to raise or address issues in code that is not being added/modified by the PR.

classes/ActionScheduler_ListTable.php
```diff
@@ -0,0 +1,517 @@
...
+		$row_html .= '</ul>';
+
+		return $row_html;
+	}
+
+	/**
+	 * Prints the logs entries inline. We do so to avoid loading Javascript and other hacks to show it in a modal.
+	 *
+	 * @param array $row Action array.
+	 */
```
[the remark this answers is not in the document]
thenbrent: @JPry it doesn't return anything

classes/ActionScheduler_ListTable.php
```diff
@@ -0,0 +1,517 @@
...
+	public function column_log_entries( array $row ) {
```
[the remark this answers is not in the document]
thenbrent: @JPry why? Are you confusing it with `print_log_entry()`?

classes/ActionScheduler_ListTable.php
```diff
@@ -0,0 +1,517 @@
...
+		echo '<ol>';
```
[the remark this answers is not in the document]
thenbrent: @JPry log entries will never be empty because an entry is set when an action is created. So if an action exists, it will have at least one log entry.

classes/ActionScheduler_ListTable.php
```diff
@@ -0,0 +1,517 @@
...
+				'names'   => _n_noop( '%s minute', '%s minutes', 'action-scheduler' ),
+			),
+			array(
+				'seconds' => 1,
+				'names'   => _n_noop( '%s second', '%s seconds', 'action-scheduler' ),
+			),
+		);
+
+		parent::__construct( array(
+			'singular' => __( 'action-scheduler', 'action-scheduler' ),
```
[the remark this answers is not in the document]
thenbrent: @cesar-prospress can you explain this one?

classes/ActionScheduler_LogEntry.php
thenbrent: > is there a reason the type is "mixed" instead of "int"?

I assume it's because there is no guarantee an ID will always be an `int`. It is an `int` with built-in datastores, but it could just as easily be a string if a hash or SHA was used to refer to an action by other data stores.

thenbrent:
@JPry thanks for the review, I've addressed all the comments (mostly that means adding patches, occasionally it means explaining why the proposed change isn't suitable or asking for more information/clarification about why the proposed change is needed).

I'll probably do more work on this next week to try to clean up the git history a bit more (i.e. rebase all those whitespace/docblock changes into originating commits).

classes/ActionScheduler_ActionFactory.php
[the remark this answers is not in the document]
JPry: > Do you see these as valuable mainly for use with generated documentation? (i.e. docs outside the code) Or is it for something else?

It would certainly be useful for generated documentation, and I do hope we can make use of that. But I also think it's valuable to be consistent 🤷‍♂️

classes/ActionScheduler_ListTable.php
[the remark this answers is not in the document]
JPry: My comment about not translating them came from looking at the constructor for `WP_List_Table`. While I don't *think* that translating these will break anything, it looks like they're meant to be used as slugs, so translation doesn't seem appropriate.

classes/ActionScheduler_ListTable.php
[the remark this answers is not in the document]
JPry: Regarding this and [my other comment](https://github.com/Prospress/action-scheduler/pull/113#discussion_r174603300), this actually goes back to the `WP_List_Table` class itself. Inside `WP_List_Table::single_row_columns()`, these methods are called thanks to this `elseif` block (`*` added to indicate the appropriate line):

```php
} elseif ( method_exists( $this, 'column_' . $column_name ) ) {
	echo "<td $attributes>";
*	echo call_user_func( array( $this, 'column_' . $column_name ), $item );
	echo $this->handle_row_actions( $item, $column_name, $primary );
	echo "</td>";
}
```

The returned values of the method for each row are immediately echoed by the calling function. Also, the `ActionScheduler_ListTable::column_args()` method returns a string rather than calling `echo` directly. This is why I was saying that the generated strings should be returned instead of echoed.

As for the black magic of `PP_List_Table`, here's the order of calls that ultimately lead to each column being called:

* `PP_List_Table::display()`
* `WP_List_Table::display()`
* `WP_List_Table::display_rows_or_placeholder()`
* `WP_List_Table::display_rows()`
* `WP_List_Table::single_row()` (called within `foreach` loop for all rows)
* `WP_List_Table::single_row_columns()`
* `echo call_user_func( array( $this, 'column_' . $column_name ), $item );` (called within a `foreach` loop for with all columns)

classes/ActionScheduler_ListTable.php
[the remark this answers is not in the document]
JPry: > So if an action exists, it will have at least one log entry.

Good to know, thanks for the clarification

classes/ActionScheduler_LogEntry.php
JPry: Ah-hah, makes perfect sense. Thanks 🙂

classes/ActionScheduler_LogEntry.php
JPry: I'm completely fine with doing docblocks separately. I've mainly mentioned them here just to make sure everything is covered. Personally, I find it easier to include simple things like docblocks when I'm already working in a file, but I can understand that's not the same for everyone.

classes/ActionScheduler_wpPostStore.php
JPry: > For me, that avoids interrupting my thinking on the more important parts of the code.

As far as I'm concerned, this is a completely valid reason to avoid making these changes now and leave it as a separate PR like you mentioned. For me, making these kinds of changes as I go is trivial and doesn't interrupt my flow, since I can highlight any relevant lines and use a single keyboard shortcut to reformat according to WP standards. But as I said in another comment, I understand that's not the same for everyone.

JPry:
@thenbrent I've responded to all of the comments that needed a response, and hopefully clarified some of my more confusing remarks. Let me know when you're ready for a re-review of the code.

classes/ActionScheduler_ListTable.php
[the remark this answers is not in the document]
thenbrent: @JPry thanks for explaining. That makes sense. It's not entirely clear from the docblock on `WP_List_Table::__construct()` if they should be keys or translated strings, it does say they'll be used for things like class names, and given `sanitize_key()` is applied to them in `WP_List_Table::__construct()`, I think you're right so I've changed them. 👍

classes/ActionScheduler_ListTable.php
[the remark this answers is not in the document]
thenbrent: @JPry thanks for the explanation. That means some other `column_` methods need to be updated too, both in here and `PP_List_Table`. I'll get that fixed up now.

thenbrent:
@JPry thanks for the updates. I've addressed those final issues now. I've also changed the base of this to `master` now that `1-6-dev` has been merged into `master` and I've merged `master` into it to make sure it will merge cleanly. The commit history is still pretty gross, but I don't think I'll be able to spend the time cleaning it up.

JPry:
@thenbrent Looks like this PR has conflicts in the `ActionScheduler_wpPostStore` class.

thenbrent:
I've just added some more commits to this monstrosity. See 591f664...6925e1d. The most notable items are:

* fixes to display of the next schedule date, and row actions on pending actions
* addition of a search box
* removal of the `'all'` status concept

I've also extended the `PP_List_Table` class and cleaned it up a little so that it takes care of a bit more common work, like displaying status filters.

While the code (and docblocks etc.) here could still use a lot of cleaning up, it's ready for review and with the addition of the _Search_ box, it now checks off all items on #85.

vendor/pp-list-table/src/class-pp-list-table.php
```diff
@@ -0,0 +1,659 @@
...
+		$column_html = esc_html( $item[ $column_name ] );
+		$column_html .= $this->maybe_render_actions( $item, $column_name );
+		return $column_html;
+	}
+
+	/**
+	 * Display the table heading and search query, if any
+	 */
+	protected function display_header() {
+		echo '<h1 class="wp-heading-inline">' . esc_attr( $this->table_header ) . '</h1>';
```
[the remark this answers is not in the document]
thenbrent: @JPry yeah that's outside the scope of this PR, you can submit an issue at https://github.com/Prospress/pp-list-table

classes/ActionScheduler_ListTable.php
```diff
@@ -0,0 +1,518 @@
...
+	/**
+	 * Only display row actions for pending actions.
+	 *
+	 * @param array $row     Row to render
+	 * @param $column_name   Current row
+	 */
+	protected function maybe_render_actions( $row, $column_name ) {
+		if ( 'pending' === strtolower( $row['status'] ) ) {
+			return parent::maybe_render_actions( $row, $column_name );
+		}
```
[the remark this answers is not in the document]
thenbrent: @JPry `parent::maybe_render_actions()`, which is `PP_List_Table::maybe_render_actions()` calls `return;` if `empty( $this->row_actions[ $column_name ] )`, which means it returns `null`, which is the same as not calling `return` I believe. Happy for that to be improved if you want, it's pretty sloppy/unnecessarily inconsistent, but I don't see a need for that to block this PR being merged.

classes/ActionScheduler_ListTable.php
```diff
@@ -0,0 +1,518 @@
...
+	/**
+	 * Returns the recurrence of an action or 'Non-repeating'. The output is human readable.
+	 *
+	 * @param ActionScheduler_Action $action
+	 *
+	 * @return string
+	 */
+	protected function get_recurrence( $action ) {
+		if ( $action->get_schedule()->is_recurring() ) {
+			return sprintf( __( 'Every %s', 'action-scheduler' ), self::human_interval( $action->get_schedule()->interval_in_seconds() ) );
```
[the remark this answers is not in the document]
thenbrent: @JPry that is beyond the scope fo this PR. It's a legitimate issue though, so feel free to open a new Issue for it.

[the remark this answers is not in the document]
thenbrent: @JPry I've added https://github.com/Prospress/action-scheduler/issues/133 to address the broader issue with that.

To address the specific issue in this PR, I've reverted https://github.com/Prospress/action-scheduler/pull/113/commits/73fcd9e0bfb936602c513dadce637fb8b77ac7fa.

Not ideal using `method_exists()`, but until #133 is implemented, it's the only reliable approach (I don't want to add a fix for #133 to this already bloated PR).

thenbrent:
@JPry thanks for finding/fixing those issues! Commits https://github.com/Prospress/action-scheduler/compare/6925e1df...3a9ff9ec6a all look good to me.

> Just calling out that this is still here instead of the `get_post_field()` function mentioned above.

Not sure why, but GitHub won't let me reply inline to that comment.

I had to remove the use of `get_post_field()`. For an explanation of why, see comments on https://github.com/Prospress/action-scheduler/pull/113/commits/f3c76e7980ac8f8bc692b484eae6980dc717b4ed

vendor/pp-list-table/src/class-pp-list-table.php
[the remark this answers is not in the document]
JPry: @thenbrent Could you please give me access to that repo?

classes/ActionScheduler_ListTable.php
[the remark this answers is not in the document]
JPry: >  Happy for that to be improved if you want, it's pretty sloppy/unnecessarily inconsistent, but I don't see a need for that to block this PR being merged.

It's no longer blocking, as I included it in the updates I made to the PR while you were out 🙂

PR #114, opened 2018-03-12
Let WP CLI run multiple batches at once

### LINKED ISSUE #112: Run batches via WP CLI Queue Runner until no more actions to run
At the moment, the WP CLI Queue Runner will process one batch of size `--batch-size` (100 by default), then exits. This matches the default Queue Runner's behaviour, but it doesn't need to.

The reason only one batch is processed in each request with the default Queue Runner is to handle PHP, server and other timeouts imposed on web requests to a site.

WP CLI avoids those, and provides an outstanding opportunity to claim and process batches repeatedly and until all pending actions scheduled to run are processed.

Of course, there will likely need to be some hosts where we need to restrict the number of batches to 1, or something else. For those cases, we should also provide a new flag to disable the repeated processing (not sure on naming, perhaps a boolean `--repeat` or int `--batch-count` which can be set to `0` by default to process batches indefinitely).

JPry:
Fixes #112.

### DIFF
--- a/classes/ActionScheduler_WPCLI_QueueRunner.php
+++ b/classes/ActionScheduler_WPCLI_QueueRunner.php
@@ -142,7 +142,7 @@ class ActionScheduler_WPCLI_QueueRunner extends ActionScheduler_Abstract_QueueRu
 	 */
 	public function before_execute( $action_id ) {
 		/* translators: %s refers to the action ID */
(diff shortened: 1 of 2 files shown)


JPry:
Accidentally submitted this from my personal repo instead of the Prospress repo (bad CLI defaults). Closing in favor of a native branch PR

PR #115, opened 2018-03-12
Let WP CLI run multiple batches at once

### LINKED ISSUE #112: Run batches via WP CLI Queue Runner until no more actions to run

JPry:
Fixes #112.

### DIFF
--- a/classes/ActionScheduler_WPCLI_Scheduler_command.php
+++ b/classes/ActionScheduler_WPCLI_Scheduler_command.php
@@ -38,13 +38,7 @@ class ActionScheduler_WPCLI_Scheduler_command extends WP_CLI_Command {
 
 			$completed = $runner->run();
 		} catch ( Exception $e ) {
-			WP_CLI::error(
-				sprintf(
-					/* translators: %s refers to the exception error message. */
-					__( 'There was an error running the action scheduler: %s', 'action-scheduler' ),
-					$e->getMessage()
-				)
-			);
+			$this->print_error( $e );
 		}
 
 		$this->print_total_batches( $batches_completed );
@@ -63,6 +57,25 @@ class ActionScheduler_WPCLI_Scheduler_command extends WP_CLI_Command {
 				/* translators: %d refers to the total number of batches executed */
 				_n( '%d batch executed.', '%d batches executed.', $batches_completed, 'action-scheduler' ),
 				number_format_i18n( $batches_completed )
+			)
+		);
+	}
+
+	/**
+	 * Convert an exception into a WP CLI error.
+	 *
+	 * @author Jeremy Pry
+	 *
+	 * @param Exception $e The error object.
+	 *
+	 * @throws \WP_CLI\ExitException
+	 */
+	protected function print_error( Exception $e ) {
+		WP_CLI::error(
+			sprintf(
+				/* translators: %s refers to the exception error message. */
+				__( 'There was an error running the action scheduler: %s', 'action-scheduler' ),
+				$e->getMessage()
 			)
 		);
 	}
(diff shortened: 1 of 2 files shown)


classes/ActionScheduler_WPCLI_Scheduler_command.php
```diff
@@ -13,44 +13,116 @@ class ActionScheduler_WPCLI_Scheduler_command extends WP_CLI_Command {
...
-			$completed = $runner->run();
+
+			// Run actions for as long as possible.
+			while ( $total > 0 && ( $unlimited || $batches_completed < $batches ) ) {
+				$this->print_total_actions( $total );
+				$actions_completed += $runner->run();
+				$batches_completed++;
+
+				// Set up tasks for the next batch.
+				$total = $runner->setup( $batch, $force );
```
thenbrent: `$runner->setup()` claims actions, which means in the last iteration of the loop, a batch of actions is claimed, but never run, because the `while`'s batch count condition fails and `$runner->run()` is not called.

Initially, I thought the fix might be as simple as replacing `$total` with the `$total = $runner->setup()` assignment in the `while` conditions. That will still claim actions though, because `$runner->setup()` is called even when `$batches_completed < $batches` proves false, so the order of those two conditions will also need to be switched to make sure the batch count condition fails before claiming anymore actions, e.g.

```php
// Run actions for as long as possible.
while ( ( $unlimited || $batches_completed < $batches ) && ( $total = $runner->setup( $batch, $force ) ) > 0 ) {
```

JPry: @thenbrent Good catch!

An alternate solution might be to add a `__destruct()` method to the `ActionScheduler_WPCLI_QueueRunner` class:

```php
public function __destruct() {
    $this->store->release_claim( $this->claim );
}
```

This would generally ensure that the claim is released regardless of the reason for the process stopping, although of course there might be exceptions for certain kinds of failures.

thenbrent: > regardless of the reason for the process stopping

@JPry generally, we want the claim to remain if the process stopped for unknown reasons so we can mark `in-progress` actions as failed after the timeout period. Releasing the claim anytime `ActionScheduler_WPCLI_QueueRunner::__desctruct()` is called could lead some actions to fail silently.

Leaving the bug in `run()` and addressing it outside of there can also create other problems (that will be difficult to diagnose). For example, if the claim changes between when `ActionScheduler_WPCLI_QueueRunner::run()` is called and when `ActionScheduler_WPCLI_QueueRunner::__desctruct()` is called, which could (unintentionally) happen seeing as the claim is staked and set on the instance in `ActionScheduler_WPCLI_QueueRunner::setup()`, which can be called independently of `run()`.

JPry: Ah-hah, that makes sense. I can see why the `__destruct()` method isn't the best here. I'll post another commit with some updates in the morning.

JPry:
@thenbrent This is take 2 at solving the issue you mentioned. I've removed the `__destruct()` method, and instead adjusted the location of the `( $unlimited || $batches_completed < $batches )` condition. This is now inline when `$total` is redefined, so that a new batch isn't claimed if we aren't actually going to use it. I think this makes the `while` condition a bit more intuitive while still accomplishing the same goal of not claiming what we won't process.

thenbrent:
@JPry LGTM. 👍 

It's not ideal having duplicate calls to `$runner->setup( )`, but one advantage of doing it this way is that the batch count check is only applied _after_ the first iteration, which makes sense given`--batch=0` means unlimited not "none", so we know we _always_ want to run at least one batch when there are actions to process.

PR #123, opened 2018-03-15
Update readme to include WP CLI details

### LINKED ISSUE #116: Add docs on WP CLI usage to README.md
- [x] document that a WP CLI command exists
- [x] explain when WP CLI is a more appropriate choice for running queues than the default WP Cron runner e.g. long running tasks, large queues, other plugins using WP Cron extensively etc. Explain the benefits of CLI context and freedom from normal constraints of web requests, like script timeouts when using WP CLI... sometimes. Also explain [how to disable the WP Cron runner](https://github.com/Prospress/action-scheduler-disable-default-runner). 
- [x] document the options:
    - [x] `--batch-size`
    - [x] `--batches` - mentioning default is to run batches as long as there are pending action, but that it can also be explicitly limited to 1...n batches of `--batch-size`
    - [x] `--force` 
- [x] explain how to get a full list of options/flags from the command line, in case docs fall out-of-sync with available options
- [x] mention `ActionScheduler_QueueRunner` as an example for creating a custom runner in the [section on using a custom runner](https://github.com/Prospress/action-scheduler/tree/1-6-dev#i-dont-want-to-use-wp-cron-does-action-scheduler-depend-on-wp-cron)

JPry:
Fixes #116

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,33 @@ Among other tasks, from the admin screen you can:
 
 Still have questions? Check out the [FAQ below](#faq).
 
+## WP CLI
+
+Action Scheduler has custom [WP CLI](http://wp-cli.org) commands available for processing actions.
+
+For many sites, WP CLI is a much better choice for running queues of actions than the default WP Cron runner. These are some common cases where WP CLI is a better option:
+
+* long-running tasks - Tasks that take a significant amount of time to run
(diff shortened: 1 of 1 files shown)


thenbrent:
Perfect 💯

PR #124, opened 2018-03-15
Update WP CLI queue cleaning default

### LINKED ISSUE #118: Queue cleaning defaults with larger batch sizes, especially with WP CLI
The WP CLI runner is amazing. So amazing, I've successfully run batches of renewals > 10,000. That's something I can definitely see large, real sites doing once it's available.

That will be great, until a batch or batches are terminated and leave unclaimed actions. When that happens, you can end up with very large batches of pending actions still claimed. This is 
a problem because `ActionScheduler_QueueCleaner`'s default cleanup batch size is only 20 (albeit filterable through `action_scheduler_cleanup_batch_size`).

If you end up in a situation like one of my test sites, with over 40,000 actions claimed and `pending`/`in-progress`, it would take hours or days to clean up the queue and allow new queues to run. The actual time will depend on the frequency the WP CLI command is run, which presumably, would be less frequent than WP Cron when the batch size is large.

There's a few things we could do to avoid this. The simplest would be to:

1. use the `--batch-size` flag for WP CLI commands to also change the `action_scheduler_cleanup_batch_size` value. That means the same number of actions being claimed would also be cleaned up.
1. and/or provide a new `--cleanup-batch-size` option which can be used to explicitly set the cleanup batch size via WP CLI. Defaulting to the `--batch-size` helps avoid this gotcha, so this command would be more useful to opt-out of the larger cleanup, or perhaps to increase the clean-up size, rather than to replace the use of `--batch-size` as the default clean-up size.

That's not foolproof though. It has the following issue:

1. this is not just a WP CLI issue. The same issue would surface with the default WP Cron Queue Runner's batch size is increased substantially. This is far less likely though, given the timeout and other restrictions imposed on web requests. That's also likely why we haven't seen it surface before.
1. if the `--batch-size` is decreased, or regularly changed, there could still be large numbers of actions left claimed for long periods of time.

To handle those situations, we may also want to look at more generic solutions, like:

1. using the value returned by the `action_scheduler_queue_runner_batch_size` filter as the default passed to `action_scheduler_cleanup_batch_size`, rather than a hardcoded `20` - this takes care of the issue regardless of the runner used
1. adding some kind of auto-detection to see when the cleanup batch size is substantially lower (like 2 orders of magnitude) than the total number of claimed actions that should be reset.

JPry:
Fixes #118

### DIFF
--- a/classes/ActionScheduler_QueueCleaner.php
+++ b/classes/ActionScheduler_QueueCleaner.php
@@ -4,14 +4,36 @@
  * Class ActionScheduler_QueueCleaner
  */
 class ActionScheduler_QueueCleaner {
-	/** @var ActionScheduler_Store */
-	private $store = NULL;
 
-	private $month_in_seconds = 2678400; // 31 days
+	/** @var int */
+	protected $batch_size;
+
+	/** @var ActionScheduler_Store */
+	private $store = null;
+
+	/**
+	 * 31 days in seconds.
+	 *
+	 * @var int
+	 */
+	private $month_in_seconds = 2678400;
+
+	/**
+	 * Five minutes in seconds
+	 *
+	 * @var int
+	 */
 	private $five_minutes = 300;
 
-	public function __construct( ActionScheduler_Store $store = NULL ) {
+	/**
+	 * ActionScheduler_QueueCleaner constructor.
+	 *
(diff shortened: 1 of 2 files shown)


thenbrent:
Very elegant solution. Love the approach of reducing filter calls, and injecting the batch size to the cleaner. 😻

PR #127, opened 2018-03-26
Add get_last_attempt methods to Abstract Data Store

JPry:
This facilitates changes needed in Prospress/action-scheduler-custom-tables#12.

### DIFF
--- a/classes/ActionScheduler_Store.php
+++ b/classes/ActionScheduler_Store.php
@@ -49,12 +49,12 @@ abstract class ActionScheduler_Store {
 	 *
 	 * @author Jeremy Pry
 	 *
-	 * @param ActionScheduler_Action $action The action ID to update.
-	 * @param array                  $fields The array of field data to update.
+	 * @param string $action_id The action ID to update.
+	 * @param array  $fields    The array of field data to update.
 	 *
-	 * @return false|int
+	 * @return mixed False if the update fails, or the action ID on success.
 	 */
-	abstract public function update_action( ActionScheduler_Action $action, array $fields );
+	abstract public function update_action( $action_id, array $fields );
 
 	/**
 	 * @param string $action_id
--- a/classes/ActionScheduler_Store.php
+++ b/classes/ActionScheduler_Store.php
@@ -176,6 +176,16 @@ abstract class ActionScheduler_Store {
 	abstract public function get_last_attempt_local( $action_id );
 
 	/**
+	 * Set the last attempt for the given action.
+	 *
+	 * @param string $action_id
+	 * @param DateTime $date
+	 *
+	 * @return bool Whether setting the last attempt was successful.
+	 */
+	abstract public function set_last_attempt_date( $action_id, DateTime $date );
+
+	/**
 	 * @param string $claim_id
 	 * @return array
 	 */
--- a/classes/ActionScheduler_Store.php
+++ b/classes/ActionScheduler_Store.php
@@ -172,7 +172,7 @@ abstract class ActionScheduler_Store {
 	 *
 	 * @param string $action_id
 	 *
-	 * @return DateTime
+	 * @return DateTime|null
 	 */
 	abstract public function get_last_attempt( $action_id );
 
@@ -183,7 +183,7 @@ abstract class ActionScheduler_Store {
 	 *
 	 * @param string $action_id
 	 *
-	 * @return DateTime
+	 * @return DateTime|null
 	 */
 	abstract public function get_last_attempt_local( $action_id );
 
--- a/classes/ActionScheduler_Store.php
+++ b/classes/ActionScheduler_Store.php
@@ -178,12 +178,13 @@ abstract class ActionScheduler_Store {
 	/**
 	 * Set the last attempt for the given action.
 	 *
-	 * @param string $action_id
-	 * @param DateTime $date
+	 * @param string   $action_id The action ID to update.
+	 * @param DateTime $date      The DateTime object representing the last attempt. If not provided, the current
+	 *                            time will be used for the last attempt.
 	 *
 	 * @return bool Whether setting the last attempt was successful.
 	 */
-	abstract public function set_last_attempt_date( $action_id, DateTime $date );
+	abstract public function update_last_attempt_date( $action_id, DateTime $date = null );
 
 	/**
 	 * @param string $claim_id
--- a/classes/ActionScheduler_Store.php
+++ b/classes/ActionScheduler_Store.php
@@ -27,7 +27,7 @@ abstract class ActionScheduler_Store {
 		'scheduled_date_local' => 1,
 		'args'                 => 1,
 		'schedule'             => 1,
-		'group_id'             => 1,
+		'group'                => 1,
 		'attempts'             => 1,
 		'last_attempt_gmt'     => 1,
 		'last_attempt_local'   => 1,
--- a/classes/ActionScheduler_Store.php
+++ b/classes/ActionScheduler_Store.php
@@ -134,6 +134,28 @@ abstract class ActionScheduler_Store {
 	abstract public function get_claim_id( $action_id );
 
 	/**
+	 * Get the last time the action was attempted.
+	 *
+	 * The time should be given in GMT.
+	 *
+	 * @param string $action_id
+	 *
+	 * @return DateTime
+	 */
+	abstract public function get_last_attempt( $action_id );
+
+	/**
+	 * Get the last time the action was attempted.
+	 *
+	 * The time should be given in the local time of the site.
+	 *
+	 * @param string $action_id
+	 *
+	 * @return DateTime
+	 */
+	abstract public function get_last_attempt_local( $action_id );
+
+	/**
 	 * @param string $claim_id
 	 * @return array
 	 */
@@ -176,4 +198,3 @@ abstract class ActionScheduler_Store {
 		return self::$store;
 	}
 }
- 
--- a/classes/ActionScheduler_Store.php
+++ b/classes/ActionScheduler_Store.php
@@ -218,7 +218,8 @@ abstract class ActionScheduler_Store {
 	/**
 	 * Get valid action fields based on known valid fields.
 	 *
-	 * @param array $fields Array of fields.
+	 * @param array $fields Array of fields in field name => value format. Field names must be
+	 *                      present in $this->action_fields to be kept.
 	 *
 	 * @return array Array of valid fields.
 	 */
--- a/classes/ActionScheduler_Store.php
+++ b/classes/ActionScheduler_Store.php
@@ -25,6 +25,18 @@ abstract class ActionScheduler_Store {
 	abstract public function save_action( ActionScheduler_Action $action, DateTime $scheduled_date = NULL );
 
 	/**
+	 * Update an existing action by ID.
+	 *
+	 * @author Jeremy Pry
+	 *
+	 * @param ActionScheduler_Action $action The action ID to update.
+	 * @param array                  $fields The array of field data to update.
+	 *
+	 * @return false|int
+	 */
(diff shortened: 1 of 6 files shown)


classes/ActionScheduler_wpPostStore.php
```diff
@@ -24,6 +41,64 @@ public function save_action( ActionScheduler_Action $action, DateTime $date = NU
...
+	 * Update an existing action by ID.
+	 *
+	 * @author Jeremy Pry
+	 *
+	 * @param string $action_id The action ID to update.
+	 * @param array  $fields    The array of field data to update.
+	 *
+	 * @return mixed False if the update fails, or the action ID on success.
+	 */
+	public function update_action( $action_id, array $fields ) {
```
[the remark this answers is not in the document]
JPry: I've done some de-duplication on these methods, and I think it's looking better at this point. 

I would ultimately like to do much more to make `save_action()` and `update_action()` work together, but I think that would require more rearchitecting.

JPry:
@thenbrent this is ready for another review.

classes/ActionScheduler_wpPostStore.php
```diff
@@ -11,28 +11,137 @@ class ActionScheduler_wpPostStore extends ActionScheduler_Store {
...
 			$post_id = $this->save_post_array( $post_array );
-			$this->save_post_schedule( $post_id, $action->get_schedule() );
-			$this->save_action_group( $post_id, $action->get_group() );
-			do_action( 'action_scheduler_stored_action', $post_id );
+
+			if ( null !== $schedule ) {
+				$this->save_post_schedule( $post_id, $schedule );
+			}
+
+			if ( null !== $schedule ) {
```
thenbrent: Should this be `null !== $group `?

classes/ActionScheduler_Store.php
```diff
@@ -24,6 +44,18 @@
 	 */
 	abstract public function save_action( ActionScheduler_Action $action, DateTime $date = NULL );
 
+	/**
+	 * Update an existing action by ID.
+	 *
+	 * @author Jeremy Pry
+	 *
+	 * @param string $action_id The action ID to update.
+	 * @param array  $fields    The array of field data to update.
```
thenbrent: Might be worth saying the field (most likely) needs to be in `$this->action_fields`. Although there is no way to know that for sure, or enforce that in child classes given the generic nature of this method.

classes/ActionScheduler_Store.php
```diff
@@ -14,6 +14,26 @@
...
+	 */
+	protected $action_fields = array(
+		'action_id'            => 1,
+		'hook'                 => 1,
+		'status'               => 1,
+		'scheduled_date_gmt'   => 1,
+		'scheduled_date_local' => 1,
+		'args'                 => 1,
+		'schedule'             => 1,
+		'group_id'             => 1,
```
thenbrent: There is no concept of a `group_id` in Action Scheduler, only a group slug/name, which is a string. The group ID exists in custom tables, because it has a group table. But in Action Scheduler itself, a custom taxonomy is used, and only the taxonomy name, not the taxonomy ID, is used to refer to the group.

classes/ActionScheduler_Store.php
```diff
@@ -14,6 +14,26 @@
...
+		'attempts'             => 1,
```
thenbrent: There is no concept of `attempts` in Action Scheduler core, only in the custom tables plugin. The number of attemps is stored in the `menu_order` column, but it's never used. It's handy to be able to use/update, but given it's not implemented in `ActionScheduler_wpPostStore` (because it's not set in `ActionScheduler_wpPostStore::$field_map`), I'm not sure whether it should be included.

classes/ActionScheduler_wpPostStore.php
```diff
@@ -11,28 +11,137 @@ class ActionScheduler_wpPostStore extends ActionScheduler_Store {
...
+	protected $field_map = array(
+		'action_id'            => 'ID',
+		'hook'                 => 'post_title',
+		'status'               => 'post_status',
+		'scheduled_date_gmt'   => 'post_date_gmt',
+		'scheduled_date_local' => 'post_date',
+		'args'                 => 'post_content',
+		'last_attempt_gmt'     => 'post_modified_gmt',
+		'last_attempt_local'   => 'post_modified',
+		'claim_id'             => 'post_password',
```
thenbrent: I'm not really sure we want claims to be updated via a generic `update_action()` call. There is a fairly specific process already for staking a claim, and allowing it to be set outside that could easily lead to unexpected results (like invalid claim IDs, or worse, reused claim IDs).

JPry:
> Did you see that previous comment? If so, can you explain why you are avoiding a specific method?

I'm sorry, I did see it and then when I was actually working through your comments I forgot about it somehow. Totally my mistake there.

With all of the comments that you mentioned, I can definitely agree that the more general update method introduces more problems than it solves, especially in the existing post store. With that said, I still feel like a general update method is needed, especially for the custom tables. However, I could concede that the general update method could be protected or even private, and expose specific methods for specific data that is public.

I'll make some updates here, but I'll still respond to some of the items you mentioned above separately, since I think there might be some good discussion.

classes/ActionScheduler_wpPostStore.php
JPry: Yep. Copy/pasta

classes/ActionScheduler_Store.php
JPry: You're absolutely right. The reason I stuck with `group_id` is mainly for the needs of custom tables.

> But in Action Scheduler itself, a custom taxonomy is used, and only the taxonomy name, not the taxonomy ID, is used to refer to the group.

True. However, one gotcha (the good kind?) of the code is that WordPress can accept an ID in place of a slug when using `wp_set_object_terms()`, provided it is an integer. So technically the taxonomy ID (`group_id` here) can still work the same way.

My thought here was that we should model the Custom Tables functionality as the future, and therefore preferred, functionality, and fit the older data store into it. I think I'm just getting ahead of myself here.

classes/ActionScheduler_Store.php
JPry: > The number of attemps is stored in the `menu_order` column, but it's never used. 

I didn't realize that was the case. Good to know. I was assuming I would just have to ignore the number of attempts for the post data store. I could certainly include that functionality in the post data store.

classes/ActionScheduler_wpPostStore.php
JPry: I agree. This makes me think all the more that it would be good to have the generic update being protected/private, and have specific dedicated methods for the various pieces of data.

classes/ActionScheduler_Store.php
```diff
@@ -133,6 +153,38 @@
...
+
+	/**
+	 * Set the last attempt for the given action.
+	 *
+	 * @param string $action_id
+	 * @param DateTime $date
+	 *
+	 * @return bool Whether setting the last attempt was successful.
+	 */
+	abstract public function set_last_attempt_date( $action_id, DateTime $date );
```
thenbrent: I think this method should probably be renamed `update_last_attempt_date()`.

The issue wioth `set_` is that, at least in the WC CRUD context, it means "set this data on the object in memory" not "save this data to the database", which is what this method actually does.

Initially I was thinking `save_` or `store_` would be a better prefix to match the other methods which write to the database, but I think `update_` is suitable because the action has to exist, so it must have a last attempt date, and that will align it with the wording used in WordPress.

classes/ActionScheduler_wpPostStore.php
```diff
@@ -638,4 +749,83 @@ public function init() {
...
+	 * @param string   $action_id The action ID to update.
+	 * @param DateTime $date The DateTime object representing the last attempt. If not provided, the current
+	 *                       time will be used for the last attempt.
+	 *
+	 * @return bool Whether setting the last attempt was successful.
+	 */
+	public function set_last_attempt_date( $action_id, DateTime $date = null ) {
+		if ( null === $date ) {
+			$date = as_get_datetime_object();
+		}
```
thenbrent: Why is `$date` optional here when it's required on the definition in `ActionScheduler_Store`?

classes/ActionScheduler_wpPostStore.php
```diff
@@ -638,4 +749,83 @@ public function init() {
...
+	 *
+	 * @param string $action_id
+	 *
+	 * @return DateTime
+	 * @throws InvalidArgumentException When the action cannot be found.
+	 */
+	public function get_last_attempt_local( $action_id ) {
+		$post = $this->get_valid_post_object( $action_id );
+
+		return as_get_datetime_object( $post->post_modified, get_option( 'gmt_offset', '0' ) );
```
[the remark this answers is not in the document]
JPry: `ActionScheduler_TimezoneHelper::get_local_timezone()`, and by extension, `$this->get_local_timezone()` returns a `DateTimeZone` object, while we need a string representation of the timezone.

I think it would be worthwhile to extract the portion of `ActionScheduler_TimezoneHelper::get_local_timezone()` that determines the timezone string into a separate method, so I'll go that route.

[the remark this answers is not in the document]
thenbrent: > we need a string representation of the timezone.

That wouldn't be needed if this method is refactored to:

```
$last_attempt_date = $this->get_last_attempt();
$last_attempt_date->setTimezone( ActionScheduler_TimezoneHelper::get_local_timezone() );
return $last_attempt_date;
```

Which is better anyway as it avoids duplicating the calls to `get_valid_post_object ()` to `as_get_datetime_object()` in both methods and instead _only_ applies logic to change what is different between this method and `get_last_attempt()` - the timezone.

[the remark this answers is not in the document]
JPry: A much better option indeed, thanks! I'll use that instead.

classes/ActionScheduler_wpPostStore.php
JPry: I meant to go back and update the abstract definition, but missed that. I'll include that change.

classes/ActionScheduler_Store.php
JPry: Good call. I'm still getting used to thinking about things in context of WC CRUD (even after all these months). I think it's a good idea to change the name to `update_*`.

classes/ActionScheduler_wpPostStore.php
```diff
@@ -11,28 +11,151 @@ class ActionScheduler_wpPostStore extends ActionScheduler_Store {
...
+						break;
+
+					case 'args':
+						$post[ $this->field_map[ $field ] ] = is_string( $value ) ? $value : json_encode( $value );
+						break;
+
+					default:
+						if ( ! isset( $this->field_map[ $field ] ) ) {
+							continue;
+						}
```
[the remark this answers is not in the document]
JPry: Thanks for pointing that out. I'll have updates to this method as a whole, which should eliminate this issue and some others you mentioned.

classes/ActionScheduler_wpPostStore.php
```diff
@@ -11,28 +11,151 @@ class ActionScheduler_wpPostStore extends ActionScheduler_Store {
...
+
+						// Validate that we actually need to do an update to the field.
+						if ( $post[ $this->field_map[ $field ] ] === $old_post[ $this->field_map[ $field ] ] ) {
```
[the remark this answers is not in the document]
JPry: Good catch. I had meant to compare `$value` as you mentioned in your comment.

thenbrent:
Closing this for reasons outlined in https://github.com/Prospress/action-scheduler/pull/144#issuecomment-390959024

Replacement is https://github.com/Prospress/action-scheduler-custom-tables/pull/26

PR #132, opened 2018-04-27
Don't stop WP CLI execution when claim lost on one batch

### LINKED ISSUE #130: WP CLI: "Error: The claim has been lost. Aborting."
While trying to run 3 concurrent queues of a large number of scheduled actions on a WP Engine site via WP CLI using the SSH Gateway, I keep bumping into this: `Error: The claim has been lost. Aborting.`

It happens midway through and terminates the processing, even when `--batches=0` or some value higher than the current batches.

With a cursory investigation, I think what is happening is that another batch is being claimed in another process. That calls `ActionScheduler_QueueCleaner::reset_timeouts()` (via `ActionScheduler_Abstract_QueueRunner::run_cleanup()`).

Because the batch sizes are so large (100 by default for WP CLI queue runner), they are taking more than `'action_scheduler_timeout_period'` time (5 minutes by default) so they are being reclaimed.

That's all fine, and as intended.

Except that WP CLI can now run multiple batches since #115, so we need to make sure that when a claim is lost (because another queue called `ActionScheduler_QueueCleaner::reset_timeouts()` when claiming a batch), that instead of aborting the current process completely, we abort the processing of just that batch, and then continue with another batch whenever `--batches=0` or some value higher than the current number of processed batches. Otherwise, there will always be a gotcha when running multiple concurrent batches and hoping to run all pending actions (the gotcha being that `--batch-size` needs to be reduced to an amount that can safely run in < `'action_scheduler_timeout_period'`).

JPry:
Fixes #130.

This converts the WP CLI error into a warning, which will allow the processing to continue.

In looking at this, I wonder whether it would be a good idea to add a filter to the batch timeout for WP CLI operations, so that "old" claims are never actually removed. That might also help here, since the underlying problem of batches being unclaimed still remains.

### DIFF
--- a/classes/ActionScheduler_WPCLI_QueueRunner.php
+++ b/classes/ActionScheduler_WPCLI_QueueRunner.php
@@ -92,15 +92,6 @@ class ActionScheduler_WPCLI_QueueRunner extends ActionScheduler_Abstract_QueueRu
 	}
 
 	/**
-	 * Ensure the progress bar has finished properly.
-	 *
-	 * @author Jeremy Pry
-	 */
-	protected function finish_progress_bar() {
-		$this->progress_bar->finish();
-	}
-
-	/**
 	 * Process actions in the queue.
 	 *
 	 * @author Jeremy Pry
@@ -126,7 +117,7 @@ class ActionScheduler_WPCLI_QueueRunner extends ActionScheduler_Abstract_QueueRu
 		}
 
 		$completed = $this->progress_bar->current();
-		$this->finish_progress_bar();
+		$this->progress_bar->finish();
 		$this->store->release_claim( $this->claim );
 		do_action( 'action_scheduler_after_process_queue' );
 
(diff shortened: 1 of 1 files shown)


thenbrent:
LGTM. 👍

> I wonder whether it would be a good idea to add a filter to the batch timeout for WP CLI operations, so that "old" claims are never actually removed

It's best to let "old" claims still be removed/cleaned. That will always mean that batch size won't be guaranteed when concurrency > 1, but there's not really anyway around that. We'll never know for sure how long it will take to process a batch, so unless we set some really large number on the batch timeout (i.e. `WEEK_IN_SECONDS`) we might still end up having batches get cleaned midway through by another process.

If we did remove the cleaning, it's also possible we'd halt Action Scheduler completely if the WP CLI processes all terminate midway through to a point where number of queue failures > allowed concurrency, because then no new claims would be allowed (that's primarily why we clear the old queues).

PR #136, opened 2018-05-08
Travis and Unit test updates

JPry:
This update aims to make it easier to run unit tests locally, while also making some updates to Travis. In particular:

 1. We're now using the [Travis Build Matrix](https://docs.travis-ci.com/user/customizing-the-build/#Build-Matrix) more effectively. 
1. We're not testing against the latest version of WordPress (4.9 at the time of this PR), as well as testing against WordPress trunk. Additionally, any test failures against WordPress trunk will not trigger a failed build.
1. Trusty is used with Travis for all version of PHP 5.4 and later. 5.3 is tested against Precise
1. Code coverage is only run when we've actually generated the `clover.xml` file. This prevents needlessly running codecov.io.
1. PHPUnit is now included with Composer, which means we don't have to overwrite the version provided by Travis, and we're able to get the same results locally
1. There's a dedicated `run_tests.sh` script to run tests, whether locally or on Travis.
1. The `bootstrap.php` file does a better job of determining where the WordPress tests directory lives.


JPry:
@thenbrent I've added a first round of updates to the Readme file based on your suggestions. I used VVV as the tool of choice since that's what I'm most familiar with. It's not quite as verbose as the readme for subscriptions that you linked to, but I think it's closer to what we're looking for.

> The composer dependencies also require PHP 7.1: https://cl.ly/1W162Z1O233q

This should no longer be the case. Apparently, when `composer update` is run with PHP 7.1 (or greater), the dependencies that are pulled in could potentially be higher than what we really need. I ran `composer update` again using PHP 7.0, and rebased the changes into the commit history. I've also mentioned the need for PHP 7.0+ in the readme.

Let me know what other changes you think are necessary here, and I'll be glad to make additional updates.

JPry:
I noticed that this had merge conflicts with `composer.json` (thanks to #137). I've rebased against `master` and the merge conflicts are now fixed.

JPry:
@thenbrent This has been rebased again with the latest `composer.json` changes

JPry:
@thenbrent I'm planning to update with instructions for Valet+, for a couple reasons:

1. I've stopped using VVV/Vagrant for my own local development
1. Valet instructions are already in use for Subscriptions.
1. Because of the above point, I should have just gone that route to begin with 😬

JPry:
@thenbrent Yep, that's still in the plan. I have this one on my list for the week.

rrennick:
@JPry This one is quite stale so I am closing it.

PR #137, opened 2018-05-10
Small housekeeping patches

### LINKED ISSUE #109: Housekeeping Todos
This is a placeholder for housekeeping todos for me to come back to once ~~#88~~ #113 is finally merged:

- [x] update copyright year
- [x] bump next version to 2.0.0 instead of 1.6.0 given the major work that's been done to make the new admin UI usable, and also because of breaking changes in the data store - new methods are required by custom data stores
- [x] display the new admin UI both on the **WooCommerce > Status** page _and_ under the current **Tools** WordPress tools menu, because it's painful requiring that extra click, and it will be confusing having the admin screen in one location for some versions and another place for others

thenbrent:
Fixes #109

### DIFF
--- a/action-scheduler.php
+++ b/action-scheduler.php
@@ -5,7 +5,7 @@
  * Description: A robust scheduling library for use in WordPress plugins.
  * Author: Prospress
  * Author URI: http://prospress.com/
- * Version: 1.6.0-dev
+ * Version: 2.0.0-dev
  * License: GPLv3
  *
  * Copyright 2018 Prospress, Inc.  (email : freedoms@prospress.com)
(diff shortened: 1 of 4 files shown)


action-scheduler.php
```diff
@@ -26,21 +25,21 @@
...
 	}
 
-	add_action( 'plugins_loaded', 'action_scheduler_register_1_dot_6_dot_0_dev', 0, 0 );
+	add_action( 'plugins_loaded', 'action_scheduler_register_2_dot_0_dot_0_dev', 0, 0 );
 
-	function action_scheduler_register_1_dot_6_dot_0_dev() {
+	function action_scheduler_register_2_dot_0_dot_0_dev() {
 		$versions = ActionScheduler_Versions::instance();
-		$versions->register( '1.6.0', 'action_scheduler_initialize_1_dot_6_dot_0_dev' );
+		$versions->register( '1.6.0', 'action_scheduler_initialize_2_dot_0_dot_0_dev' );
```
JPry: I think this line also needs `2.0.0` in place of `1.6.0`.

thenbrent: Good catch. All the references to 1.6/1.6.0 in doc block and deprecation messages also need to be updated! I've fixed that up, rebased in the fixes and force pushed the branch.

action-scheduler.php
```diff
@@ -5,11 +5,10 @@
  * Description: A robust scheduling library for use in WordPress plugins.
  * Author: Prospress
  * Author URI: http://prospress.com/
- * Version: 1.6.0-dev
+ * Version: 2.0.0-dev
  * License: GPLv3
  *
- *
- * Copyright 2017 Prospress, Inc.  (email : freedoms@prospress.com)
+ * Copyright 2018 Prospress, Inc.  (email : freedoms@prospress.com)
```
JPry: Question: Should this actually include the prior copyright years, rather than replacing them? e.g. `Copyright 2017-2018 Prospress, Inc.`?

JPry: This isn't enough to prevent merge, so I'm going to go ahead. It's easy enough to update this later as needed.

PR #139, opened 2018-05-11
Claim actions by oldest ID

thenbrent:
Order actions by ID when staking a claim so that if multiple actions are scheduled for the same exact time (to the second), they will be run in the order in which they were created/scheduled.

However, be sure to apply the order only after ordering actions being claimed by number of previous attempts (`menu_order`) and scheduled date (`post_date_gmt`), to preserve backward compatibility.

Instigated by this [Slack conversation](https://woocommercecommunity.slack.com/archives/C4TNYTR28/p1525960776000345).


thenbrent:
@JPry thanks for adding a unit test! I had it on my todo list to add them, but didn't get far with it. The test looks good to me, so I'll merge now. 👍

PR #142, opened 2018-05-16
Fix incorrect unit test counts

### LINKED ISSUE #135: Fix Unit tests to have proper count
<a href="https://cl.ly/0l3J451K2n04" target="_blank"></a>

Currently, unit tests for Action Scheduler have an incorrect count. This is because of the `ActionScheduler_UnitTestCase` class, where each test is run multiple times:

https://github.com/Prospress/action-scheduler/blob/master/tests/ActionScheduler_UnitTestCase.php#L20-L29

Right now, we're essentially setting the default timezone multiple times for each test, rather than for the test suite as a whole. I think it would be better to adjust this to test the entire suite multiple times with each timezone.

I've done some digging into the documentation for PHPUnit, and it looks like its possible (but annoying) to customize the internals of how PHPUnit runs. Depending on how far we wanted to go, we could do some or all of the following:

* [x] Implement the `count()` method in `ActionScheduler_UnitTestCase` to provide the proper count to PHPUnit.
* [ ] Extend the `PHPUnit_TextUI_Command` class in order to override the `createRunner()` method. We would then need to ...
* [ ] Extend the `PHPUnit_TextUI_TestRunner` class, so that we can override the `doRun()` method. The reason to do this would be so we can add a custom repeater. E.g., we would have our own version that replaces the default repeat test class with one that changes the timezones in between:

```php
        if (is_int($arguments['repeat'])) {
            $test = new PHPUnit_Extensions_RepeatedTest(
                $suite,
                $arguments['repeat'],
                $arguments['processIsolation']
            );

            $suite = new PHPUnit_Framework_TestSuite();
            $suite->addTest($test);
        }
```

* [ ] Then we'd need to model our own extension on `PHPUnit_Extensions_RepeatedTest`
* [ ] After all of this, we would probably need to create a custom script that replaces `./vendor/bin/phpunit` as the entry point for running tests

The rabbit hole here is potentially quite deep 😅 

Realistically, I think the simplest option would be to just go with the very first item, and don't bother with the rest. With that said, I still wanted to document the rest, since I ended up way down the rabbit hole of PHPUnit modification and didn't want all that I learned to disappear 🙂

JPry:
Fixes #135

### DIFF
--- a/tests/ActionScheduler_UnitTestCase.php
+++ b/tests/ActionScheduler_UnitTestCase.php
@@ -8,6 +8,15 @@ class ActionScheduler_UnitTestCase extends WP_UnitTestCase {
 	protected $existing_timezone;
 
 	/**
+	 * Counts the number of test cases executed by run(TestResult result).
+	 *
+	 * @return int
+	 */
+	public function count() {
+		return 'UTC' == date_default_timezone_get() ? 2 : 3;
+	}
+
+	/**
 	 * We want to run every test multiple times using a different timezone to make sure
 	 * that they are unaffected by changes to PHP's timezone.
 	 */
@@ -33,4 +42,3 @@ class ActionScheduler_UnitTestCase extends WP_UnitTestCase {
 		return $result;
 	}
 }
- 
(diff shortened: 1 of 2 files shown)


thenbrent:
LGTM. 👍

PR #144, opened 2018-05-21
Save last attempt date via ActionScheduler_Store::save_action()

thenbrent:
This PR is an alternative to #127  for resolving Prospress/action-scheduler-custom-tables#12.

I wanted to see if there was a simpler method to solve Prospress/action-scheduler-custom-tables#12 as I had a sense #127 was blowing out of proportion for what was needed - it has 25 commits, a changeset over 400 lines, and has been open for 2 months.

The approach used in this PR requires only one commit and a changeset of < 50 lines. It's also fully unit tested (i.e. 1 test) and (mostly) backward compatible.

The reason a simpler alternative is needed is because https://github.com/Prospress/action-scheduler-custom-tables/pull/12 is the final piece holding up both Subscriptions 2.3 RC, and Action Scheduler 2.0.

### DIFF
--- a/classes/ActionScheduler_Store.php
+++ b/classes/ActionScheduler_Store.php
@@ -19,10 +19,12 @@ abstract class ActionScheduler_Store {
 	 * @param DateTime $date Optional date of the first instance
 	 *        to store. Otherwise uses the first date of the action's
 	 *        schedule.
+	 * @param DateTime $last_attempt_date Optional date of the last time
+	 *        the action was attempted. Otherwise set to 0000-00-00 00:00:00.
 	 *
 	 * @return string The action ID
 	 */
-	abstract public function save_action( ActionScheduler_Action $action, DateTime $date = NULL );
+	abstract public function save_action( ActionScheduler_Action $action, DateTime $date = NULL, DateTime $last_attempt_date = NULL );
 
 	/**
 	 * @param string $action_id
--- a/classes/ActionScheduler_wpPostStore.php
+++ b/classes/ActionScheduler_wpPostStore.php
@@ -11,9 +11,9 @@ class ActionScheduler_wpPostStore extends ActionScheduler_Store {
 	/** @var DateTimeZone */
 	protected $local_timezone = NULL;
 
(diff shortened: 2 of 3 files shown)


thenbrent:
This feedback has made me realise Action Scheduler doesn't actually, and shouldn't actually, be changed at all to address Prospress/action-scheduler-custom-tables#10.

The specific issue in Prospress/action-scheduler-custom-tables#12 is that _when **migrating** a completed action to a new data store, the action's attempt date is not being migrated correctly_.

This has nothing to do with saving or updating actions. It's exclusively a migration bug, which is partly why attempts to resolve it for more generic saving or updating functions have come up short.

Because of that, I'm going to close this PR, and #127, and instead submit a new patch to AC Custom Tables that focusing on fixing the issue during migration.

> we certainly do need to make sure there's a way to update a given action rather than just creating a new one

For posterities sake, I don't see why we need that?

There's no need to update the last attempt date on an existing action to solve Prospress/action-scheduler-custom-tables#10. There is a need to _set_ it on a newly created action when it's migrated, but not to update it (that's why this PR fixed Prospress/action-scheduler-custom-tables#10).

I've also haven't seen the need in Action Scheduler, Subscriptions or other plugins to update the attempt date outside of staking the claim either.

Really though, the last attempt date should normally _only_ be updated when an action is claimed, not via some other API acting on the action either on creation or update, because that introduces the possibility of it being the last attempt date being desynchronised with the actual attempt, meaning instead of actually representing a valid attempt date, it may represent just some arbitrary date someone wanted to set on it  (this is closely related to something I mentioned on #127 previously).

> the `ActionScheduler_Action` class has no concept of storing the action ID.

Yeah, that is a much bigger issue that is beyond the scope of what’s needed to fix Prospress/action-scheduler-custom-tables#10 though.

It's the same issue we encountered, discussed and attempted to resolve when looking at the admin UI.

I’m still not sure whether it’s a real issue or just a perceived issue. For example, it’s perceived to be an issue here, but really, it’s actually a signal that attempting to solve Prospress/action-scheduler-custom-tables#10 via an approach that requires the action ID to be accessible is the wrong path to take.

PR #146, opened 2018-05-21
Refactor scheduled date string APIs

thenbrent:
The custom tables `DB_Store` class currently duplicates:

* `ActionScheduler_wpPostStore::get_timestamp()`
* `ActionScheduler_wpPostStore::get_local_timestamp()` and
* `ActionScheduler_wpPostStore::get_local_timezone()`

To avoid that, these methods can be moved into the shared `ActionScheduler_Store` base class. Which is part of what this PR proposes doing.

This also provides an opportunity to more accurately rename the first two of those methods, as the current method names indicate the methods return a Unix timestamp, when the actual return value is a MySQL formatted date/time string.

Because this patch is being proposed as part of v2.0, I've removed the previous `ActionScheduler_wpPostStore::get_timestamp()` methods instead of deprecating them, as it is a major release, so breaking backward compatibility is acceptable.

### DIFF
--- a/classes/ActionScheduler_Store.php
+++ b/classes/ActionScheduler_Store.php
@@ -185,6 +185,15 @@ abstract class ActionScheduler_Store {
 	}
 
 	/**
+	 * Get the site's local time. Wrapper for ActionScheduler_TimezoneHelper::get_local_timezone().
+	 *
+	 * @return DateTimeZone
+	 */
+	protected function get_local_timezone() {
+		return ActionScheduler_TimezoneHelper::get_local_timezone();
+	}
+
+	/**
 	 * @return array
 	 */
 	public function get_status_labels() {
--- a/classes/ActionScheduler_Store.php
+++ b/classes/ActionScheduler_Store.php
@@ -151,6 +151,40 @@ abstract class ActionScheduler_Store {
 	}
 
 	/**
+	 * Get the time MySQL formated date/time string for an action's (next) scheduled date.
+	 *
+	 * @param ActionScheduler_Action $action
+	 * @param DateTime $scheduled_date (optional)
+	 * @return string
+	 */
+	protected function get_scheduled_date_string( ActionScheduler_Action $action, DateTime $scheduled_date = NULL ) {
+		$next = null === $scheduled_date ? $action->get_schedule()->next() : $scheduled_date;
+		if ( ! $next ) {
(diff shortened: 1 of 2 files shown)


classes/ActionScheduler_Store.php
```diff
@@ -150,6 +150,49 @@ protected function validate_sql_comparator( $comparison_operator ) {
...
 
+	/**
+	 * Get the time MySQL formated date/time string for an action's (next) scheduled date.
+	 *
+	 * @param ActionScheduler_Action $action
+	 * @param DateTime $scheduled_date (optional)
+	 * @return string
+	 */
+	protected function get_scheduled_date_string( ActionScheduler_Action $action, DateTime $scheduled_date = NULL ) {
+		$next = is_null( $date ) ? $action->get_schedule()->next() : $date;
```
JPry: These instances of `$date` should instead be `$scheduled_date`.

From a performance perspective, it would be preferable to use `null === $scheduled_date` instead of `is_null( $schdeuled_date )`. The reason is that a function call is a bit slower than the explicit comparison (example here: https://stackoverflow.com/a/4662691). With that said, it's also worth noting that the difference seems to much less, or even nonexistent, as of PHP7 (see the replies to the previous link).

classes/ActionScheduler_Store.php
```diff
@@ -150,6 +150,49 @@ protected function validate_sql_comparator( $comparison_operator ) {
...
+
+	/**
+	 * Get the time MySQL formated date/time string for an action's (next) scheduled date.
+	 *
+	 * @param ActionScheduler_Action $action
+	 * @param DateTime $scheduled_date (optional)
+	 * @return string
+	 */
+	protected function get_scheduled_date_string_local( ActionScheduler_Action $action, DateTime $scheduled_date = NULL ) {
+		$next = is_null( $date ) ? $action->get_schedule()->next() : $date;
```
JPry: Same issues as the comment above regarding the variable and performance.

classes/ActionScheduler_Store.php
thenbrent: I'll wait to see what happens with https://github.com/Prospress/action-scheduler-custom-tables/pull/26/ before addressing the `$date` vs `$scheduled_date` issue here as it will likely require/benefit from more substantial refactoring if that is merged, like abstracting part of into new `get_date_string()`/`get_date_string_local()` methods.

JPry:
I rebased this against `master` to fix merge conflicts, and while so doing I corrected the issues mentioned above.

PR #150, opened 2018-05-25
Avoid the use of `@return void`

### LINKED ISSUE #107: Remove @return void
The `@return` DocBlock should not include `void`: https://stackoverflow.com/a/2061559

> The datatype should be a valid PHP type

AS has a number of instances of that which should be fixed up.

JPry:
Fixes #107


thenbrent:
LGTM 👍

Issue #151, opened 2018-05-30
Admin issues with 2.0.0


sirtawast:
Hi,

thanks for all the great work to get `action-scheduler` to where it is now. It's really the most comprehensive solution for WordPress background worker, event/action queue I've tried.

There's a couple of major bugs in the latest beta release that prevents me to use it in production (I've been using a master commit build previously with better luck, thinking it was 591f66436).

My docker builds:
**WordPress / PHP-FPM**
`wordpress:4.9.4-php7.1-fpm` utilizing `PHP 7.1.14`

**NGINX**
`nginx:1.13.8`

---

### Delete posts

`/wp-admin/tools.php?page=action-scheduler`

Deletion of `scheduled-action` posts gives a blank admin template with following warning:

```
Warning: Cannot modify header information - headers already sent by (output started at /var/www/html/wp-admin/includes/template.php:2039) in /var/www/html/wp-includes/pluggable.php on line 1216
```

The deletion itself goes through.

---

### Ordering posts by any column

`/wp-admin/tools.php?page=action-scheduler`

Changing the order of `scheduled-action` posts by any column gives the following error:

```
Fatal error: Uncaught TypeError: Argument 3 passed to ActionScheduler_ActionFactory::get_stored_action() must be of the type array, null given, called in /var/www/html/wp-content/plugins/action-scheduler/classes/ActionScheduler_wpPostStore.php on line 101 and defined in /var/www/html/wp-content/plugins/action-scheduler/classes/ActionScheduler_ActionFactory.php:17 

Stack trace: 
#0 /var/www/html/wp-content/plugins/action-scheduler/classes/ActionScheduler_wpPostStore.php(101): ActionScheduler_ActionFactory->get_stored_action('pending', 'my_scheduled_task...', NULL, Object(ActionScheduler_SimpleSchedule), 'my_bg_request') 
#1 /var/www/html/wp-content/plugins/action-scheduler/classes/ActionScheduler_wpPostStore.php(77): ActionScheduler_wpPostStore->make_action_from_post(Object(WP_Post)) 
#2 /var/www/html/wp-content/plugins/action-scheduler/classes/ActionScheduler_ListTable.php(486): ActionScheduler_wpPostStore->fetch_action('41359') 
[...]

PR #152, opened 2018-05-31
Better handling for malformed args

JPry:
Helps with an issue brought up in #151.

### DIFF
--- a/classes/ActionScheduler.php
+++ b/classes/ActionScheduler.php
@@ -67,8 +67,8 @@ abstract class ActionScheduler {
 			return;
 		}
 
-		if ( file_exists( $dir.$class.'.php' ) ) {
-			include( $dir.$class.'.php' );
+		if ( file_exists( "{$dir}{$class}.php" ) ) {
+			include( "{$dir}{$class}.php" );
 			return;
 		}
 	}
--- a/classes/ActionScheduler_Exception.php
+++ b/classes/ActionScheduler_Exception.php
@@ -1,0 +1,11 @@
+<?php
+
+/**
+ * ActionScheduler Exception Interface.
+ *
+ * Facilitates catching Exceptions unique to Action Scheduler.
+ *
+ * @package Prospress\ActionScheduler
+ * @since %VERSION%
+ */
+interface ActionScheduler_Exception {}
--- a/classes/ActionScheduler_InvalidActionException.php
+++ b/classes/ActionScheduler_InvalidActionException.php
@@ -1,0 +1,28 @@
+<?php
+
+/**
+ * InvalidAction Exception.
+ *
+ * Used for identifying actions that are invalid in some way.
+ *
+ * @package Prospress\ActionScheduler
+ */
+class ActionScheduler_InvalidActionException extends \InvalidArgumentException implements ActionScheduler_Exception {
+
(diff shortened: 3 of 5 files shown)


thenbrent:
@JPry I think AS should throw an exception instead of defaulting to empty args. Empty args could potentially let things continue without any notice or logged issue, when there was an error that should be addressed (and could potentially lead to things like missed payments etc.).

thenbrent:
> what we do with any thrown exceptions

Definitely a valid concern.

I think that needs to be treated as a separate issue to how malformed args are handled though. Largely because there are many different ways these malformed args can occur, or be encountered, and the ideal solutions to most of those (the code responsible for scheduling the action handling it based on its knowledge of the what the action is/does) is outside of our control.

> This is because ActionScheduler_ListTable::prepare_items() just ignores Exceptions:

That's a good example of a case, and also something that should be handled outside this PR. Doing it as part of this PR would slow down the better handling of malformed args because I think `ActionScheduler_ListTable::prepare_items()` should really be handling _all_ exceptions better, not just any custom AS exceptions, or known exceptions. Ideally, `ActionScheduler_ListTable` would print something about the exception in the table (or not catch and ignore the exception at least, to let PHP do that for us). But that is more around an issue with the list table's handling of exception than it is with handling for malformed args.

thenbrent:
I should have also said above, definitely 👍 for using custom exceptions for this (and other) exceptions. :)

JPry:
> definitely 👍 for using custom exceptions for this (and other) exceptions.

This brings to mind something I've been thinking about but haven't yet written up into a proper issue. I was thinking that it would ultimately be good to move Action Scheduler into using a namespace and implementing a PSR-4 structure.

In order to accomplish this without taking time away from other more important things, this could be done more incrementally. For example, in this case custom exceptions could be added in the proper Structure while maintaining existing structure for everything else. Then we can gradually move everything over. I'll try to get this written up into a proper issue tomorrow with more fleshed out details.

JPry:
@thenbrent I've pushed some new commits that start to implement the changes I mentioned, without going way out of scope for this PR (in my opinion):

* Added the `src/` directory. This will be the root of the `Prospress\ActionScheduler` namespace.
* Added a PSR-4 autoloader. This keeps us from needing to rely on composer, and we'll be able to handle the updated Namespace going forward.
* Created a custom Exception interface. All of our custom exceptions will implement this type, and it will allow us to start catching exceptions that are unique to Action Scheduler.
* Created the first custom exception, `InvalidAction`. This exception currently has a method that will return a new exception object when an action ID has invalid arguments. We can add more methods over time as needed, as well as different types of exceptions.

This admittedly does go a bit beyond what we're looking to accomplish here, but I think it's within reason considering my prior comment.

Let me know what you think in general, and feel free to review when you're able 🙂

JPry:
@thenbrent I've amended the commits for this PR to **only** include the new Exceptions. The existing autoloader method has been updated to handle locating these files. Additionally, they are located in the existing `classes/` directory instead of a new `src/` directory.

JPry:
@thenbrent I've updated this PR so that the changes fit better with the existing codebase structure.

thenbrent:
LGTM. 👍

PR #154, opened 2018-06-04
Adding ability to execute actions by group

### LINKED ISSUE #121: Add --group option for WP CLI
Being able to run queues for specific groups of actions is valuable at scale. Why? Because it means you can restrict the concurrency for similar actions.

For example, let's say you have 300,000 actions queued up comprised of:

* 100,000 renewals payments
* 100,000 email notifications
* 100,000 membership status updates

Action Scheduler's default WP Cron queue runner will process them all together. e.g. when it claims a batch of actions, some may be emails, some membership updates and some renewals.

When you add concurrency to that, you can end up with issues. For example, if you have 3 queues running, they may all be attempting to process similar actions at the same time, which can easily lead to querying the same database tables with similar queries. Depending on the code/queries running, this can lead to database locks or other issues.

If you can batch based on each action's group, then you can improve performance by processing like actions consecutively, but still processing the full set of actions concurrently.

For example, if one queue is created to process emails, another to process membership updates, and another to process renewal payments, then the same queries won't be run at the same time, and 3 separate queues will be able to run more efficiently.

The WP CLI runner is a good opportunity to implement this. It will almost certainly become the most recommended option for larger stores, which are the ones who will most need this option.

JPry:
This adds the ability for Action Scheduler to claim actions based on a given group.

* `ActionScheduler_Store::stake_claim()` (as well as child classes) now include a `$group` parameter. When a group is passed, only actions in that group will be returned. When omitted, all matching actions regardless of a group (or no group) will be returned.
* `ActionScheduler_WPCLI_QueueRunner::setup()` also includes a `$group` parameter. 
* `ActionScheduler_WPCLI_Scheduler_Command` now allows passing `--group=<group>` using WP CLI.

Fixes #121

### DIFF
--- a/classes/ActionScheduler_WPCLI_QueueRunner.php
+++ b/classes/ActionScheduler_WPCLI_QueueRunner.php
@@ -38,13 +38,14 @@ class ActionScheduler_WPCLI_QueueRunner extends ActionScheduler_Abstract_QueueRu
 	 *
 	 * @author Jeremy Pry
 	 *
-	 * @param int  $batch_size The batch size to process.
-	 * @param bool $force      Whether to force running even with too many concurrent processes.
+	 * @param int    $batch_size The batch size to process.
+	 * @param string $group      The group of actions to claim with this batch.
+	 * @param bool   $force      Whether to force running even with too many concurrent processes.
 	 *
 	 * @return int The number of actions that will be run.
 	 * @throws \WP_CLI\ExitException When there are too many concurrent batches.
 	 */
-	public function setup( $batch_size, $force = false ) {
+	public function setup( $batch_size, $group = '', $force = false ) {
 		$this->run_cleanup();
 		$this->add_hooks();
 
@@ -60,7 +61,7 @@ class ActionScheduler_WPCLI_QueueRunner extends ActionScheduler_Abstract_QueueRu
 		}
 
 		// Stake a claim and store it.
-		$this->claim = $this->store->stake_claim( $batch_size );
+		$this->claim = $this->store->stake_claim( $batch_size, null, $group );
 		$this->monitor->attach( $this->claim );
 		$this->actions = $this->claim->get_actions();
 
--- a/classes/ActionScheduler_wpPostStore.php
+++ b/classes/ActionScheduler_wpPostStore.php
@@ -432,15 +432,18 @@ class ActionScheduler_wpPostStore extends ActionScheduler_Store {
 	}
 
 	/**
-	 * @param int $max_actions
+	 * @param int      $max_actions
 	 * @param DateTime $before_date Jobs must be schedule before this date. Defaults to now.
+	 * @param string   $group       Claim only actions in the given group.
 	 *
 	 * @return ActionScheduler_ActionClaim
+	 * @throws RuntimeException When there is an error staking a claim.
 	 */
-	public function stake_claim( $max_actions = 10, DateTime $before_date = NULL ){
+	public function stake_claim( $max_actions = 10, DateTime $before_date = null, $group = '' ) {
 		$claim_id = $this->generate_claim_id();
-		$this->claim_actions( $claim_id, $max_actions, $before_date );
+		$this->claim_actions( $claim_id, $max_actions, $before_date, $group );
 		$action_ids = $this->find_actions_by_claim_id( $claim_id );
+
 		return new ActionScheduler_ActionClaim( $claim_id, $action_ids );
 	}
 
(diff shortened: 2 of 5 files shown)


classes/ActionScheduler_wpPostStore.php
```diff
@@ -462,25 +465,93 @@ protected function generate_claim_id() {
...
+		$where  = "WHERE post_type = %s AND post_status = %s AND post_password = ''";
+		$params[] = self::POST_TYPE;
+		$params[] = ActionScheduler_Store::STATUS_PENDING;
+
+		// If we have IDs, include them in the WHERE clause. Otherwise use the old date parameter.
+		if ( count( $ids ) > 0 ) {
+			$where .= ' AND ID IN (' . join( ',', $ids ) . ')';
+//		} else {
+//			$where    .= ' AND post_date_gmt <= %s';
+//			$params[] = $date->format( 'Y-m-d H:i:s' );
```
JPry: @thenbrent I'd like to solicit some feedback on what I've been doing here. Here's my thought process on how this should be updated to work:

1. To enable querying via a group (which in this case is a taxonomy), I though it would make the most sense to use `WP_Query` if possible. In this case, we would be querying for IDs using the taxonomy.
1. With a set of IDs, we could then limit our own query to the set of IDs we identified, and thus we accomplish the limit by group.

Here's where I'm having a problem, and I think that I'm stuck enough to question whether I'm even approaching this the right way.

Using `WP_Query` requires that we recreate the same conditions that are being applied using `$wpdb->query()` below. Otherwise, we may end up with post IDs that don't match. This leads me to setting `post_type`, `post_status`, and also `date_query`. The problem is in using `date_query`, it's not quite exactly what we use in our condition below. *In theory* it should end up with the same results, but of course that's not a good assumption.

In trying to avoid making assumptions, I've been actively running our existing Unit Tests to ensure I'm not breaking anything, and that has been very helpful. Where I'm at now is that there is a particular test that keeps failing: `ActionScheduler_QueueRunner_Test::test_next_instance_of_action()`. Well, it fails so long as these three lines remain commented out.

This tells me that the my `WP_Query` parameters are ending up with an empty result, which leads me to think that I'm doing something incorrectly. Or, I'm otherwise missing something.

I'm open to any thoughts you have, whether it's a different approach entirely or a better way to run `WP_Query`. I've also tried something like this for `date_query`, with even less successful results:

```
			'date_query'     => array(
				array(
					'column' => 'post_date_gmt',
					'compare' => '<=',
					'before' => $date->format( 'Y-m-d H:i:s' ),
					'inclusive' => true,
				),
			),
```

thenbrent: Hrm, both of those date queries look OK to me. At least based on revising the [WP Query](https://codex.wordpress.org/Class_Reference/WP_Query#Date_Parameters) and [`WP_Date_Query`](https://developer.wordpress.org/reference/classes/wp_date_query/) docs.

I double checked that `WP_Date_Query` supported `<=` because of the comment in the existing code about not using `$wpdb->update()` because it doesn't support `<=`, but it looks fine too.

The `'tax_query'` also looks good.

It's unlikely, but something else could be adding another `tax_query` or `date_query` (via a lower level WP filter), and because the default `relation` is `AND`, it's returning empty results. That's so unlikely it feels barely worth mentioning, but I'm not seeing any more likely answers for the reason it's not working. Other than some obvious error we're both missing around syntax of the `tax_query` or `date_query` (I often get tripped up on the multidimensional arrays).

I think this is a good approach. If you want to explore an alternative seeing as you can't get it to work yet, you could try using a subquery in the existing `UPDATE` query. That would sure be easier to test than using `WP_Query` as you can just run the query directly as SQL. It's not as simple as it could be though given the disjointed schema for taxonomies and terms.

JPry:
@thenbrent I worked on some other things today, and then I had an "ah-hah!" moment. I realized the issue that I previously described was essentially caused by continuing to attempt to update the DB when there were no matching actions. The fact that there were no matching actions was actually correct. So I solved this by exiting early with `return 0` when `WP_Query` didn't find any matching actions.

This is good for review now.

classes/ActionScheduler_WPCLI_QueueRunner.php
```diff
@@ -38,13 +38,14 @@ public function __construct( ActionScheduler_Store $store = null, ActionSchedule
...
-	 * @param bool $force      Whether to force running even with too many concurrent processes.
+	 * @param int    $batch_size The batch size to process.
+	 * @param string $group      The group of actions to claim with this batch.
+	 * @param bool   $force      Whether to force running even with too many concurrent processes.
 	 *
 	 * @return int The number of actions that will be run.
 	 * @throws \WP_CLI\ExitException When there are too many concurrent batches.
 	 */
-	public function setup( $batch_size, $force = false ) {
+	public function setup( $batch_size, $group = '', $force = false ) {
```
thenbrent: I had to do a double take at changing the param order here, but as it's unreleased, it is fine (and I like having `$group` as the 2nd param). 👍

classes/ActionScheduler_wpPostStore.php
```diff
@@ -462,25 +465,101 @@ protected function generate_claim_id() {
...
+		$date = null === $before_date ? as_get_datetime_object() : clone $before_date;
+
+		// Set up a query for post IDs to use later.
+		$query      = new WP_Query();
+		$query_args = array(
+			'fields'           => 'ids',
+			'post_type'        => self::POST_TYPE,
+			'post_status'      => ActionScheduler_Store::STATUS_PENDING,
+			'has_password'     => false,
+			'posts_per_page'   => $limit,
```
thenbrent: Because this select and the claim are being run in separate queries, there is a race condition where:

1. this query returns a valid set of IDs that are due to be claimed
1. another process claims that same set of results, because it ran the same query with the same order by clause
1. the claim query is run below, but doesn't stake anything because `post_password` is no longer `''`

That's not a critical issue. It just means this attempt to stake a claim will be ignored.

However, I think it can be avoided quite easily by increasing the `'posts_per_page'` value here to something like `2 * $limit` OR `3 * $limit`. That would allow 2 or 3 processes to be attempting to stake claims at the same time. Because of the ease of addressing it, IMO it's worth doing.

The `$limit` is necessary of course to avoid the MySQL query below becoming too long and running into things like WP Engine's [1024 character limit on MySQL queries](https://wpengine.com/support/troubleshoot-wordpress-wp-engine-error-log/#longquery). But I think even with a 2x or 3x increase in the limit, we'll generally fall safely within that with default configurations.

classes/ActionScheduler_wpPostStore.php
```diff
@@ -462,25 +465,101 @@ protected function generate_claim_id() {
...
+					'taxonomy'         => self::GROUP_TAXONOMY,
+					'field'            => 'slug',
+					'terms'            => $group,
+					'include_children' => false,
+				),
+			);
+		}
+
+		// Gather relevant IDs.
+		$ids = $query->query( $query_args );
```
thenbrent: There is an overheard in running this query. Especially on large databases. Because of that, I'd only use it when we need to filter by group and do a `'tax_query'` rather than every time a claim is staked, regardless of whether this query is needed or not.

classes/ActionScheduler_WPCLI_QueueRunner.php
JPry: Yeah, I thought group would be better as the second parameter.

classes/ActionScheduler_wpPostStore.php
JPry: This makes sense. I think `3 * $limit` is reasonable.

classes/ActionScheduler_wpPostStore.php
JPry: I understand. That would necessitate dividing up this function to either use `WP_Query` with the `tax_query` and `date_query`, or doing it the old way. I'll take a look at architecting that to avoid duplication as much as possible.

JPry:
@thenbrent Thanks for all the feedback! My updates have been submitted. Ready for another review.

classes/ActionScheduler_wpPostStore.php
```diff
@@ -462,25 +466,127 @@ protected function generate_claim_id() {
...
+		if ( ! term_exists( $group, self::GROUP_TAXONOMY )) {
+			throw new InvalidArgumentException( __( 'The group "%s" does not exist.', 'action-scheduler' ), $group );
+		}
+
+		// Set up a query for post IDs to use later.
+		$query      = new WP_Query();
+		$query_args = array(
+			'fields'           => 'ids',
+			'post_type'        => self::POST_TYPE,
+			'post_status'      => ActionScheduler_Store::STATUS_PENDING,
```
thenbrent: Because this status is not mentioned in the method name, it'll be a gotcha calling `get_actions_by_group()` and expecting to get completed, cancelled or failed actions as well. That's not a blocking issue though. Just something to keep in mind - we may need to add a `$status` param later if necessary, or change the method name to explicitly declare that it acts only on pending actions.

PR #161, opened 2018-06-13
Support hooks filter when staking a claim and add a WP CLI --hooks option to take advantage of it

thenbrent:
This is a quick prototype to see what it would take to implement #160. Still a WIP - I need to test passing arrays with the `--hooks` option for WP CLI to ensure it works. It appears the syntax is just comma separated strings, e.g.

```
action-scheduler run --hooks=woocommerce_scheduled_subscription_trial_end,woocommerce_scheduled_subscription_expiration
```

Also need to add docs to the README once I confirm the syntax for that.

### DIFF
--- a/classes/ActionScheduler_WPCLI_QueueRunner.php
+++ b/classes/ActionScheduler_WPCLI_QueueRunner.php
@@ -39,13 +39,14 @@ class ActionScheduler_WPCLI_QueueRunner extends ActionScheduler_Abstract_QueueRu
 	 * @author Jeremy Pry
 	 *
 	 * @param int    $batch_size The batch size to process.
+	 * @param array  $hooks      The hooks being used to filter the actions claimed in this batch.
 	 * @param string $group      The group of actions to claim with this batch.
 	 * @param bool   $force      Whether to force running even with too many concurrent processes.
 	 *
 	 * @return int The number of actions that will be run.
 	 * @throws \WP_CLI\ExitException When there are too many concurrent batches.
 	 */
-	public function setup( $batch_size, $group = '', $force = false ) {
+	public function setup( $batch_size, $hooks = array(), $group = '', $force = false ) {
 		$this->run_cleanup();
 		$this->add_hooks();
 
@@ -61,7 +62,7 @@ class ActionScheduler_WPCLI_QueueRunner extends ActionScheduler_Abstract_QueueRu
 		}
 
 		// Stake a claim and store it.
-		$this->claim = $this->store->stake_claim( $batch_size, null, $group );
+		$this->claim = $this->store->stake_claim( $batch_size, null, $hooks, $group );
 		$this->monitor->attach( $this->claim );
 		$this->actions = $this->claim->get_actions();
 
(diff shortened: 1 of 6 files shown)


JPry:
@thenbrent Regarding this:

> It appears the syntax is just comma separated strings

That is one commonly used method, but you'll need to manually convert the comma separated string to an array. WP CLI will only read it as a string. I [opened an issue](https://github.com/wp-cli/wp-cli/issues/2212) for allowing reuse of a flag a long time ago, but it was rejected. Hypothetically that would have looked like this:

```
action-scheduler run --hook=woocommerce_scheduled_subscription_trial_end --hook= woocommerce_scheduled_subscription_expiration
```

In any case, it won't be challenging to manually parse a comma separated list to an array.

thenbrent:
Thanks for the tips @JPry. I've added that to make sure the string is parsed correctly and multiple args are handled. This is ready for review now.

classes/ActionScheduler_WPCLI_Scheduler_command.php
```diff
@@ -34,6 +37,7 @@ public function run( $args, $assoc_args ) {
 		$batch   = absint( \WP_CLI\Utils\get_flag_value( $assoc_args, 'batch-size', 100 ) );
 		$batches = absint( \WP_CLI\Utils\get_flag_value( $assoc_args, 'batches', 0 ) );
 		$clean   = absint( \WP_CLI\Utils\get_flag_value( $assoc_args, 'cleanup-batch-size', $batch ) );
+		$hooks   = explode( ',', trim( \WP_CLI\Utils\get_flag_value( $assoc_args, 'hooks', array() ) ) );
```
JPry: 🛑 Since this is inside the call to `explode()`, I believe the 3rd parameter passed to `\WP_CLI\Utils\get_flag_value()` should be an empty string instead of `array()`. Otherwise we'll get a warning like this when no `--hooks` parameter is passed:

> Warning: explode() expects parameter 2 to be string, array given in php shell code on line 1

Unfortunately, explode is a bit annoying, because if we pass an empty string, instead of returning an empty array it will return an array with only the empty string. So I think that we should possibly do something like instead:

```php
$raw_hooks = explode( ',', WP_CLI\Utils\get_flag_value( $assoc_args, 'hooks', '' ) );
$hooks     = array_filter( array_map( 'trim', $raw_hooks ) );
```
*(I'm illustrating separate variables here because the line was starting to get long)*

I think it's worth doing `array_map( 'trim' )` to ensure that no whitespace is passed in. While the shell won't pass it through if it's unquoted, it's possible someone could do this: `--hooks='foo_hook, bar_hook'`. I think it's a good idea to protect against that to reduce any annoyance to users later.

Then of course `array_filter()` will clean up any empty elements.

classes/ActionScheduler_WPCLI_QueueRunner.php
```diff
@@ -39,13 +39,14 @@ public function __construct( ActionScheduler_Store $store = null, ActionSchedule
...
 	 * @param int    $batch_size The batch size to process.
+	 * @param array  $hooks      The hooks being used to filter the actions claimed in this batch.
 	 * @param string $group      The group of actions to claim with this batch.
 	 * @param bool   $force      Whether to force running even with too many concurrent processes.
 	 *
 	 * @return int The number of actions that will be run.
 	 * @throws \WP_CLI\ExitException When there are too many concurrent batches.
 	 */
-	public function setup( $batch_size, $group = '', $force = false ) {
+	public function setup( $batch_size, $hooks = array(), $group = '', $force = false ) {
```
JPry: ❔ I don't see this as an issue, but I'm just curious, why put hooks before `$group`?

thenbrent: That was to try and match the order of params on `wc_schedule_single_action()` etc. (they're different param sets still, but `$hook` is before `$group` there, so figured it made sense to try to maintain that order here). :)

classes/ActionScheduler_WPCLI_Scheduler_command.php
thenbrent: Ah good catch, thanks! I'll get that fixed up tomorrow.

thenbrent:
Self-merging as it was previously approved, and the only change since then was to implement the code provided https://github.com/Prospress/action-scheduler/pull/161#discussion_r196750350 (which fixes that issue).

PR #162, opened 2018-06-15
Implement custom DateTime class

### LINKED ISSUE #120: Replace use of DateTime::format( 'U' ) with DateTime::getTimestamp()
Some PHP versions prior to 7.2 have an issue where `DateTimeZone` is set with UTC offsets. Specifically, when `DateTimeZone` is set with a UTC offset, `DateTime::format( 'U' )` will return a different value than `DateTime::getTimestamp()`.

Bug report here: https://bugs.php.net/bug.php?id=73489

Sample code demonstrating the issue, and versions of PHP on which it is present: https://3v4l.org/IFuca

The easiest solution for this is to replace use of `DateTime::format( 'U' )` with `DateTime::getTimestamp()` whenever possible (i.e. when PHP 5.3 or newer is active).

If we add our own `DateTime` child class which extends `DateTime`, but implements `getTimestamp()`, we can also rely on `getTimestamp()` existing regardless of PHP version (similar to `WC_DateTime`).

Source: https://poststatus.slack.com/archives/C0413TGQZ/p1520550295000228

JPry:
Fixes #120.

This does the following:

* Creates a custom class, `ActionScheduler_DateTime`. For now, this class simply implements the `getTimestamp()` method to ensure compatibility with PHP 5.2.
* Uses the new class everywhere in the plugin
* Adds a unit test for the `as_get_datetime_object()` function

### DIFF
--- a/classes/ActionScheduler_CronSchedule.php
+++ b/classes/ActionScheduler_CronSchedule.php
@@ -36,7 +36,7 @@ class ActionScheduler_CronSchedule implements ActionScheduler_Schedule {
 	 * @return array
 	 */
 	public function __sleep() {
-		$this->start_timestamp = $this->start->format('U');
+		$this->start_timestamp = $this->start->getTimestamp();
 		return array(
 			'start_timestamp',
 			'cron'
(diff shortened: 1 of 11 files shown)


thenbrent:
LGTM. 👍

PR #163, opened 2018-06-15
Add max execution time for Siteground

### LINKED ISSUE #149: Handle SiteGround max execution time better
SiteGroud shared plans enforce a 120 second max execution time. That's plenty of time, so we don't see many issues, but we should still prevent any issues (see #108 for deeper background).

Now that #134, we have all the scaffolding in place for handling this timeout. We just need a way to detect if plugin code is running on SiteGround.

e.g. with WP Engine, we check `defined( 'WPENGINE_ACCOUNT' )`, with Pantheon `defined( 'PANTHEON_ENVIRONMENT' )`, but I don't know of a way to do it with SiteGround.

I asked SiteGround technical support about this, and they advised me there was no way for them to detect what code was running on a site… the opposite of what I was asking. :upside_down_face: 🙂

So I've asked in Post Status Slack now and will update here if I get an answer.

Props @jrick1229 for finding the max execution time.

JPry:
Fixes #149.


thenbrent:
LGTM. 👍

PR #177, opened 2018-07-11
Add cautionary note on action dependencies to README

thenbrent:
As well as some other readme tweaks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,28 @@ These are the commands available to use with Action Scheduler:
 
 The best way to get a full list of commands and their available options is to use WP CLI itself. This can be done by running `wp action-scheduler` to list all Action Scheduler commands, or by including the `--help` flag with any of the individual commands. This will provide all relevant parameters and flags for the command.
 
+### Cautionary Note on Action Dependencies when using `--group` or `--hooks` Options
+
+The `--group` and `--hooks` options should be used with caution if you have an implicit dependency between scheduled actions based on their schedule.
+
+For example, consider two scheduled actions for the same subscription:
+
+* `scheduled_payment` scheduled for `2015-11-13 00:00:00` and
+* `scheduled_expiration` scheduled for `2015-11-13 00:01:00`.
+
+Under normal conditions, Action Scheduler will ensure the `scheduled_payment` action is run before the `scheduled_expiration` action. Becuase that's how they are scheduled.
+
+However, when using the `--hooks` option, the `scheduled_payment` and `scheduled_expiration` actions will be processed in separate queues. As a result, this dependency is not guaranteed.
+
+For example, consider a site with both:
+
+* 100,000 `scheduled_payment` actions, scheduled for `2015-11-13 00:00:00`
+* 100 `scheduled_expiration` actions, scheduled for `2015-11-13 00:01:00`
+
+If two queue runners are running alongside each other with each runner dedicated to just one of these hooks, the queue runner handling expiration hooks will complete the processing of the expiration hooks more quickly than the queue runner handling all the payment actions.
+
+**Because of this, the `--group` and `--hooks` options should be used with caution to avoid processing actions with an implicit dependency based on their schedule in separate queues.**
+
 ### Improving Performance with `--group` or `--hooks`
 
 Being able to run queues for specific hooks or groups of actions is valuable at scale. Why? Because it means you can restrict the concurrency for similar actions.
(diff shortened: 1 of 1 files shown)


lkraav:
Hmmm, this patch didn't make it clear to me if would still need https://github.com/Prospress/action-scheduler-disable-default-runner when I already have `define( 'DISABLE_WP_CRON`, true )`.

thenbrent:
@lkraav disabling WP's default Cron has very little to do with the _ Action Scheduler - Disable Default Queue Runner_ plugin. They're different things for different purposes. The info being added here also is exclusively about running Action Scheduler via WP CLI. It's not got anything to do with WP Cron runner (other than the mention that this wasn't an issue previously when there was only a WP Cron runner).

README.md
```diff
@@ -87,9 +87,31 @@ These are the commands available to use with Action Scheduler:
...
+The `--group` and `--hooks` options should be used with caution if you have an implicit dependency between scheduled actions based on their schedule.
+
+For example, consider two scheduled actions for the same subscription:
+
+* `scheduled_payment` scheduled for `2015-11-13 00:00:00` and
+* `scheduled_expiration` scheduled for `2015-11-13 00:01:00`.
+
+Under normal conditions, Action Scheduler will ensure the `scheduled_payment` action is run before the `scheduled_expiration` action. Becuase that's how they are scheduled.
+
+However, if using the `--hooks` option, and processing payment actions in a separate queue to expirations action, this dependency is not guaranteed.
```
JPry: This sentence reads a bit awkwardly to me. I think it should be tweaked to something like this instead:

However, when using the `--hooks` option to process the `scheduled_payment` and `scheduled_expiration` actions separately, this dependency is not guaranteed.

README.md
```diff
@@ -87,9 +87,31 @@ These are the commands available to use with Action Scheduler:
...
+
+For example, consider the case of:
```
JPry: Rather than starting a new paragraph, I think it would be good to include this sentence at the end of the preceding paragraph.

README.md
```diff
@@ -87,9 +87,31 @@ These are the commands available to use with Action Scheduler:
...
+
+* 100,000 `scheduled_payment` actions all for `2015-11-13 00:00:00`, and
```
JPry: There should be a comma after "actions", and you can omit `, and` from the end.

README.md
```diff
@@ -87,9 +87,31 @@ These are the commands available to use with Action Scheduler:
...
+* only 100 `scheduled_expiration` scheduled for `2015-11-13 00:01:00`.
+
+In this case, if the two queue runners are running alongside each other, the queue runner handling expiration hooks will complete the processing of the expiration hooks more quickly than the queue handling all the payment actions.
```
JPry: "than the queue handling all..." should have "runner" after "queue"

lkraav:
> disabling WP's default Cron has very little to do with the _ Action Scheduler - Disable Default Queue Runner_ plugin. They're different things for different purposes. The info being added here also is exclusively about running Action Scheduler via WP CLI. It's not got anything to do with WP Cron runner (other than the mention that this wasn't an issue previously when there was only a WP Cron runner).

Perhaps. Was just thinking if there is a connection between running my cron exclusively via `wp cron event run --due-now`.

I think this already takes care of running AS via wp-cli cron in one shot - I have no need for disabling AS Default Queue Runner, as that would force me to run a separate wp-cli command just to process AS queue.

README.md
thenbrent: Using more paragraph breaks is the better approach for writing for the web, where people skim a lot more.

Many readers won't make it to the end of the preceding paragraph, meaning they'd end up at a list with no idea what it is referring to, and have to backtrack to figure that out.

README.md
thenbrent: > you can omit `, and` from the end

Removing `and` introduces ambiguity - _are both of these sets of actions required for the example? Or only one or them? Or can there be either both or just one?_

I've removed the `and` but addressed the ambiguity by changing the leading sentence to read:

> For example, consider a site with both:

I think that's better, but it also means [it's even more important this information isn't missed](5 days ago).

thenbrent:
Thanks for the review @JPry. I've address those issues and merging now.

> Was just thinking if there is a connection between running my cron exclusively via `wp cron event run --due-now.`

Ah not I see the connection, thanks for clarifying @lkraav. There's no connection. Running WP Cron via WP CLI and running Action Scheduler via WP CLI directly are two independent things.

> I think this already takes care of running AS via wp-cli cron in one shot - I have no need for disabling AS default queue runner, as that would force me to run a separate wp-cli command just to process AS queue.

Yes that's right. It really depends on a store's unique requirements.

There are a number of benefits to running AS via WP CLI directly, like preventing large queues of  Action Scheduler actions hogging WP Cron requests, or the opposite - having other WP Cron actions slowing down the rate of AS queue processing.

PR #180, opened 2018-07-19
Loosen timezone string check.

### LINKED ISSUE #179: Manual timezone offsets trigger fatal errors
There have been a few customer tickets recently where we see fatal errors associated with Action Scheduler. Specifically, the error looks something like this:

```
Fatal error : Uncaught RuntimeException: Error saving action: DateTimeZone::__construct(): Unknown or bad timezone () in /path/to/site/wp-content/plugins/woocommerce-subscriptions/includes/libraries/action-scheduler/classes/ActionScheduler_wpPostStore.php:23
```

The actual error is being thrown here:

https://github.com/Prospress/action-scheduler/blob/350f3ac05c1fb6ae619d9c4b7cdbb1f98b899bc6/classes/ActionScheduler_TimezoneHelper.php#L39

I'm able to reproduce this when the `timezone_string` setting is in the form of `UTC+11`.

Normally, a timezone string in this format shouldn't be possible. Even though WordPress provides these offsets in the Timezone selection box, it will map them to the `gmt_offset` option instead. This can be seen here in the WP codebase: 

https://github.com/WordPress/WordPress/blob/4c2fcf36a3193a7e63a82e5ff459f2587c69a859/wp-admin/options.php#L179-L184

With that said, it's possible that a plugin (or even a theme) could have set this value incorrectly. Reproducing the fatal can be done by doing something like this:

```php
add_filter( 'option_timezone_string', function () { return 'UTC+11'; } );
ActionScheduler_TimezoneHelper::get_local_timezone();
```

JPry:
Fixes #179.

For certain timezone strings, the `$city['timezone_id']` is set to `null`. While normally it's desirable to use the `===` operator, in this case the timezone string wasn't set to UTC.

### DIFF
--- a/classes/ActionScheduler_DateTime.php
+++ b/classes/ActionScheduler_DateTime.php
@@ -8,6 +8,16 @@
 class ActionScheduler_DateTime extends DateTime {
 
 	/**
+	 * UTC offset.
+	 *
+	 * Only used when a timezone is not set. When a timezone string is
+	 * used, this will be set to 0.
+	 *
+	 * @var int
+	 */
+	protected $utcOffset = 0;
+
+	/**
 	 * Get the unix timestamp of the current object.
 	 *
 	 * Missing in PHP 5.2 so just here so it can be supported consistently.
@@ -16,5 +26,51 @@ class ActionScheduler_DateTime extends DateTime {
 	 */
 	public function getTimestamp() {
 		return method_exists( 'DateTime', 'getTimestamp' ) ? parent::getTimestamp() : $this->format( 'U' );
+	}
+
+	/**
+	 * Set the UTC offset.
+	 *
+	 * This represents a fixed offset instead of a timezone setting.
+	 *
+	 * @param $offset
+	 */
+	public function setUtcOffset( $offset ) {
+		$this->utcOffset = intval( $offset );
+	}
+
+	/**
+	 * Returns the timezone offset.
+	 *
+	 * @return int
+	 * @link http://php.net/manual/en/datetime.getoffset.php
+	 */
+	public function getOffset() {
+		return $this->utcOffset ? $this->utcOffset : parent::getOffset();
+	}
+
+	/**
+	 * Set the TimeZone associated with the DateTime
+	 *
+	 * @param DateTimeZone $timezone
+	 *
+	 * @return static
+	 * @link http://php.net/manual/en/datetime.settimezone.php
+	 */
+	public function setTimezone( $timezone ) {
+		$this->utcOffset = 0;
+		parent::setTimezone( $timezone );
+
+		return $this;
+	}
+
+	/**
+	 * Get the timestamp with the WordPress timezone offset added or subtracted.
+	 *
+	 * @since  3.0.0
+	 * @return int
+	 */
+	public function getOffsetTimestamp() {
+		return $this->getTimestamp() + $this->getOffset();
 	}
 }
--- a/classes/ActionScheduler_TimezoneHelper.php
+++ b/classes/ActionScheduler_TimezoneHelper.php
@@ -25,6 +25,24 @@ abstract class ActionScheduler_TimezoneHelper {
 						$tzstring = timezone_name_from_abbr( '', $gmt_offset, 0 );
 					}
 
+					// Try mapping to the first abbreviation we can find.
+					if ( false === $tzstring ) {
+						$is_dst = date( 'I' );
+						foreach ( timezone_abbreviations_list() as $abbr ) {
+							foreach ( $abbr as $city ) {
+								if ( $city['dst'] == $is_dst && $city['offset'] == $gmt_offset ) {
+									// If there's no valid timezone ID, keep looking.
+									if ( null === $city['timezone_id'] ) {
+										continue;
+									}
+
+									$tzstring = $city['timezone_id'];
+									break 2;
+								}
+							}
+						}
+					}
+
 					// If we still have no valid string, then fall back to UTC.
 					if ( false === $tzstring ) {
 						$tzstring = 'UTC';
--- a/classes/ActionScheduler_TimezoneHelper.php
+++ b/classes/ActionScheduler_TimezoneHelper.php
@@ -18,18 +18,14 @@ abstract class ActionScheduler_TimezoneHelper {
 					$tzstring = 'UTC';
 				} else {
 					$gmt_offset *= HOUR_IN_SECONDS;
-					$tzstring = timezone_name_from_abbr('', $gmt_offset);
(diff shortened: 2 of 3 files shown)


thenbrent:
@JPry correct me if I'm wrong, but won't this result in silently scheduling things for the wrong time? (i.e. in UTC time, instead of whatever GMT offset is set).

I understand that by the time we get to the `empty( $tzstring )` check, we have no few options left for mapping the `get_option('timezone_string')` to a valid timezone string, but I still think throwing an exception is better than scheduling things at the wrong time. Especially as it's silently changing the timezone, because it would be extremely hard to diagnose why actions are occurring on the wrong day/time when that's done (and in the particular case of #179, where it's 11 hours off, that could easily move a scheduled event to an entirely different date).

I had a quick look at `wc_timezone_string()` to see if it had any other options. It's doing the same as you're proposing here by just checking if `$city['timezone_id']` is any type of `true` value before defaulting to `UTC`, so that's why it's not throwing the same exception, but it will also map things to the wrong timezone so has the same bug.

What do you think of doing something like [the code you shared in #179 from WP Core](https://github.com/WordPress/WordPress/blob/4c2fcf36a3193a7e63a82e5ff459f2587c69a859/wp-admin/options.php#L179-L184), and either:

1. mapping the value at runtime in `ActionScheduler_TimezoneHelper::get_local_timezone()` to handle `'/^UTC[+-]/'` type string values for `get_option('timezone_string')`, and/or 
1. setting the `'gmt_offset'` option value in the database when when the `'timezone_string'` value matches the `'/^UTC[+-]/'` regex, but no `'gmt_offset'` option value  is set

That still won't handle _all_ invalid timezone strings of course, but it will handle the one case we know WP may lead to if it's own validation/mapping code isn't run.

My preference would probably be just to go with the first item, as I'm not sure AS should be updating core WP options. But we probably want to make sure a similar patch is submitted to WC core too, to fix the bug there.

JPry:
@thenbrent I had a nice long reply typed out, then I accidentally closed my browser 😭 I'll try to cover it all again.

> won't this result in silently scheduling things for the wrong time? (i.e. in UTC time, instead of whatever GMT offset is set).

That's correct. But the alternative is having things not scheduled at all because it's killed by a fatal error. While I agree that it's better to surface the error, I think a problem there is that there's no easy way for a user to reschedule an action that didn't actually get scheduled. For example, how would they do that with Subscriptions? IMO, it's better to have something scheduled for the incorrect time than to have it not scheduled at all.

I revisited this today based on your feedback to see if we could implement a better solution. Since it's not clear exactly what circumstances trigger failures, I decided to add unit tests around the timezone helper functionality. This was actually really helpful once I pushed the changes up and allowed travis to run, as it showed that there appear to be issues only with PHP 7.0+ (see the logs for [this travis build](https://travis-ci.org/Prospress/action-scheduler/builds/406409585)).

After this, I decided to go searching on Google, and I came across [this stack overflow thread](https://stackoverflow.com/questions/11820718/convert-utc-offset-to-timezone-or-date). It appears that the fix might be far simpler, so I've updated/rebased the code to include this fix.

thenbrent:
The latest patches look good, but I also fear it's too easy. Specifically, I feel like there must be a reason SkyVerge, the WooCommerce core team, the original `get_local_timezone()` author, and others I've seen fallback to `'timezone_id'` instead of using `timezone_name_from_abbr( '', $gmt_offset, 0 )`.

My theories for why that is the case:

1. everyone is just copying someone else, and the original source didn't realise `timezone_name_from_abbr( '', $gmt_offset, 0 )` was an option
1. `timezone_name_from_abbr( '', $gmt_offset, 0 )` isn't backward compatible in PHP
1. `timezone_name_from_abbr( '', $gmt_offset, 0 )` has other corner cases not discussed or encountered by those finding https://stackoverflow.com/questions/11820718/convert-utc-offset-to-timezone-or-date

To guard against all of those, we could either:

1. increase the unit tests to provide complete coverage of [all timezones](http://php.net/manual/en/timezones.php) and [UTC offsets](https://en.wikipedia.org/wiki/List_of_UTC_time_offsets) (there's a lot of cases not handled in those atm, especially around partial offsets, like `UTC+5:45`); or
1. use `timezone_name_from_abbr( '', $gmt_offset, 0 )` when `false === timezone_name_from_abbr( '', $gmt_offset, 1 )` as proposed here, but then also double check the return value of `timezone_name_from_abbr( '', $gmt_offset, 0 )` is a valid string, and if its not, fallback to the existing code attempting to derive it from `$city['timezone_id']` in case it handles some other case.

The former of those is the better, and will give us 100% certainty we've covered all cases. But I'm OK going with the 2nd given the burden of going with the first.

> Since it's not clear exactly what circumstances trigger failures

Do you mean there are no clear/documented circumstances for when `timezone_name_from_abbr()` doesn't return a valid `$tzstring` when passed an empty string for the first param?

thenbrent:
@JPry in case you haven't seen yet, the unit tests are failing here. Otherwise, current approach LGTM. 👍

JPry:
@thenbrent I updated the tests here to include two groups:

* Offsets that map to an actual timezone string
* Offsets that do not map to an actual timezone string

The latter group will be mapped to UTC. Just to recap, the reason they're mapped to UTC is because we've tried 3 different methods to find a compatible timezone string that can be used to instantiate a `DateTimeZone` object, and couldn't find one. So instead, UTC is the fallback.

These tests cover all of the offsets that WP makes available.

JPry:
Well, it looks like there are some other tests failing now. I'll need to take a look into these on Monday to see what the issue might be, as it only seems to be a problem on versions of PHP less than 7.

thenbrent:
> we've tried 3 different methods to find a compatible timezone string that can be used to instantiate a `DateTimeZone` object, and couldn't find one.

That's true, but we also know how to instantiate a `DateTime` object when we have no compatible timezone string - the [same way WC does it](https://github.com/woocommerce/woocommerce/blob/master/includes/wc-formatting-functions.php#L665:L670) - use the UTC offset.

That means setting the wrong timezone is avoidable, it just needs AS to be refactored to avoid the dependency on a `DateTimezone`.

I've done up a quick mockup to show how that could be achieved: 
https://github.com/Prospress/action-scheduler/compare/issue_179...issue_179_redux

What's wrong with doing that to ensure we set correct timezones?

thenbrent:
Closing this in favour of https://github.com/Prospress/action-scheduler/pull/202.

The issue with this approach is that GMT/UTC offsets do not map 1-to-1 to timezone strings, because more than one place may fall in the same timezone, e.g. in `UTC-12`, both `'Pacific/Kwajalein'` and `'Pacific/Fiji'` are valid timezone strings.

PR #181, opened 2018-07-26
Use as_ prefix not wc_ prefix for public API functions

thenbrent:
To avoid confusion once Action Scheduler is included in WooCommerce core, and it provides a different set of public APIs for working with the action queue. As raised here: https://github.com/woocommerce/woocommerce/pull/20030#pullrequestreview-139753264 (I also think this is probably how the functions should have been prefixed all along, so it's a worthwhile change).

The existing functions are still loaded by default, and will continue to work normally. A deprecated notice will be thrown however to notify developers of the change (we'll also need to publicly announce this change, ideally well before the release of WC 3.5.0 to give 3rd parties time to prepare).

I've set the version in the deprecated notice to 2.1.0. I think renaming all public API functions justifies bumping to a 2.1.0 instead of just 2.0.1 or 2.0.2.

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,13 +137,13 @@ The Action Scheduler API functions are designed to mirror the WordPress [WP-Cron
 
 Functions return similar values and accept similar arguments to their WP-Cron counterparts. The notable differences are:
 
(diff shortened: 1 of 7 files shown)


JPry:
For posterity, here's a link to the passing build from before the rebase: https://travis-ci.org/Prospress/action-scheduler/builds/408316251.

PR #183, opened 2018-07-26
Don't use wc_clean()

### LINKED ISSUE #182: wc_clean() is used in the list table
While reviewing #181, I noticed that we're using `wc_clean()`, as seen here:

https://github.com/Prospress/action-scheduler/blob/13cfb2a03500f8b339dfe69b71c860986fcaa834/classes/ActionScheduler_Abstract_ListTable.php#L277-L281

Since Action Scheduler is meant to not rely on WooCommerce, we should remove this dependency.

JPry:
Fixes #182.


thenbrent:
Good catch. LGTM. 👍

PR #192, opened 2018-08-14
Restore batch processing loop

### LINKED ISSUE #184: Restore batch processing loop to increase the default processing speed
By default, Action Scheduler will process 25 actions per run. That's it. No more, despite [what the `README.md` file says](https://github.com/Prospress/action-scheduler/blob/2.0.0/README.md#batch-processing), until the next WP Cron request.

This is because the [loop for processing actions until running out of time or memory was removed](https://github.com/Prospress/action-scheduler/commit/5145a9d66e8a12639f3ebd870a0381ca977803db) to prevent the timeout errors that have since been addressed with #134.

To increase the default processing speed, we should reinstate this loop.

While we have timeout prevention in place with #134, we still need memory monitoring to safely introduce a loop.

The `WP_Background_Process` provides a good approach for how to do this with the `memory_exceeded()` method.

I checked both the [WC Implementation of `WP_Background_Process::memory_exceeded()`](https://github.com/woocommerce/woocommerce/blob/3.4.0/includes/libraries/wp-background-process.php#L343-L353) and the [SkyVerge implementation](https://github.com/woocommerce/woocommerce/blob/3.4.0/includes/libraries/wp-background-process.php#L335-L353), and neither of this have added any additional patches to that, so the approach is likely working reliably at their scale and can be relied upon in Action Scheduler.

Once we have the memory limit in place, we should be able to implement a method like `WC_Background_Process ::batch_limit_exceeded()`](https://github.com/woocommerce/woocommerce/blob/master/includes/abstracts/class-wc-background-process.php#L80-L87) in `ActionScheduler_Abstract_QueueRunner` to control the loop. This means that method can be overridden in other runners, namely `ActionScheduler_WPCLI_QueueRunner`, that do not want to be constrained by timeouts (or likely memory usage, as it has the `stop_the_insanity()` method).

### LINKED ISSUE #188: Use wp_raise_memory_limit() to increase memory limit
`ActionScheduler_QueueRunner::run()` attempts to increase the memory limit when running the queue by calling `ini_set()` directly. Specifically, it calls:

```
@ini_set( 'memory_limit', apply_filters( 'admin_memory_limit', WP_MAX_MEMORY_LIMIT ) )
```

WordPress 4.6 introduced a handy new function for increasing memory `wp_raise_memory_limit()`.

This function adds a few benefits to the manual `ini_set()` call. Most notably, it will only _increase_ the memory, it will not decrease it even if the filtered value is less than the current memory. Sounds like a corner case, and it is, but I came across it on a site recently where the value was being set to `1024` int, instead of `1024M` string, and all Action Scheduler queues were getting memory exhaustion errors (as well as presumably anything else using the filtered `'admin_memory_limit'` value).

Quite a nasty issue, so worth guarding against.

thenbrent:
This PR reintroduces a loop in `ActionScheduler_QueueRunner::run()`  to increase the rate of action processing. Instead of processing just one batch of actions, with this patch, AS will process as many batches as it can before reaching 90% of the available memory, or it becomes likely processing another action will exceed the available script timeout limit.

Fixes #184

I recommend reviewing individual commit messages for a more in-depth explanation on each change.

---

I also snuck in https://github.com/Prospress/action-scheduler/commit/1a698d04795bac737dfa54769804b5db15c33fd8 to this PR, as it's a relatively small patch, which fixes #188, and it depends on the new `ActionScheduler_Compatibility` class being added here for #184, so I figured it was worth including in the same PR.

I'll move it to a separate PR if it slows down the review/merge of the patches required for #184, as that is a higher priority issue.

### DIFF
--- a/classes/ActionScheduler_Abstract_QueueRunner.php
+++ b/classes/ActionScheduler_Abstract_QueueRunner.php
@@ -132,6 +132,46 @@ abstract class ActionScheduler_Abstract_QueueRunner {
 	}
 
 	/**
+	 * Get memory limit
+	 *
+	 * Based on WP_Background_Process::get_memory_limit()
+	 *
+	 * @return int
+	 */
+	protected function get_memory_limit() {
+		if ( function_exists( 'ini_get' ) ) {
+			$memory_limit = ini_get( 'memory_limit' );
+		} else {
+			$memory_limit = '128M'; // Sensible default, and minimum required by WooCommerce
+		}
+
+		if ( ! $memory_limit || -1 === $memory_limit || '-1' === $memory_limit ) {
+			// Unlimited, set to 32GB.
+			$memory_limit = '32G';
+		}
+
+		return ActionScheduler_Compatibility::convert_hr_to_bytes( $memory_limit );
+	}
+
+	/**
+	 * Memory exceeded
+	 *
+	 * Ensures the batch process never exceeds 90% of the maximum WordPress memory.
+	 *
+	 * Based on WP_Background_Process::memory_exceeded()
+	 *
+	 * @return bool
+	 */
+	protected function memory_exceeded() {
+
+		$memory_limit    = $this->get_memory_limit() * 0.90;
+		$current_memory  = memory_get_usage( true );
+		$memory_exceeded = $current_memory >= $memory_limit;
+
+		return apply_filters( 'action_scheduler_memory_exceeded', $memory_exceeded, $this );
+	}
+
+	/**
 	 * Process actions in the queue.
 	 *
 	 * @author Jeremy Pry
--- a/classes/ActionScheduler_Abstract_QueueRunner.php
+++ b/classes/ActionScheduler_Abstract_QueueRunner.php
@@ -132,6 +132,23 @@ abstract class ActionScheduler_Abstract_QueueRunner {
 	}
 
 	/**
+	 * Check if the host's max execution time is (likely) to be exceeded if processing more actions.
+	 *
+	 * @param int $processed_actions The number of actions processed so far - used to determine the likelihood of exceeding the time limit if processing another action
+	 * @return bool
+	 */
+	protected function time_likely_to_be_exceeded( $processed_actions ) {
+
+		$execution_time        = $this->get_execution_time();
+		$max_execution_time    = $this->get_maximum_execution_time();
+		$time_per_action       = $execution_time / $processed_actions;
+		$estimated_time        = $execution_time + ( $time_per_action * 2 );
+		$likely_to_be_exceeded = $estimated_time > $this->get_maximum_execution_time();
+
+		return apply_filters( 'action_scheduler_maximum_execution_time_likely_to_be_exceeded', $likely_to_be_exceeded, $this, $processed_actions, $execution_time, $max_execution_time );
+	}
+
+	/**
 	 * Get memory limit
 	 *
 	 * Based on WP_Background_Process::get_memory_limit()
@@ -169,6 +186,19 @@ abstract class ActionScheduler_Abstract_QueueRunner {
 		$memory_exceeded = $current_memory >= $memory_limit;
 
 		return apply_filters( 'action_scheduler_memory_exceeded', $memory_exceeded, $this );
+	}
+
+	/**
+	 * See if the batch limits have been exceeded, which is when memory usage is almost at
+	 * the maximum limit, or the time to process more actions will exceed the max time limit.
+	 *
+	 * Based on WC_Background_Process::batch_limits_exceeded()
+	 *
+	 * @param int $processed_actions The number of actions processed so far - used to determine the likelihood of exceeding the time limit if processing another action
+	 * @return bool
+	 */
+	protected function batch_limits_exceeded( $processed_actions ) {
+		return $this->memory_exceeded() || $this->time_likely_to_be_exceeded( $processed_actions );
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


classes/ActionScheduler_Abstract_QueueRunner.php
```diff
@@ -79,6 +92,115 @@ public function get_allowed_concurrent_batches() {
...
+	 * @param int $processed_actions The number of actions processed so far - used to determine the likelihood of exceeding the time limit if processing another action
+	 * @return bool
+	 */
+	protected function time_likely_to_be_exceeded( $processed_actions ) {
+
+		$execution_time      = $this->get_execution_time();
+		$max_execution_time  = $this->get_maximum_execution_time();
+		$time_per_action     = $execution_time / $processed_actions;
+		$estimated_time      = $execution_time + ( $time_per_action * 2 );
+		$likely_to_be_exceed = ( $estimated_time > $this->get_maximum_execution_time() ) ? true : false;
```
JPry: You can drop `? true : false` from the end of this line. This is redundant since the `>` operator (and other comparison operators) will return a boolean value. You can also omit the parentheses around the condition.

One minor thing here is the name of the variable. I'd suggest either `$likely_to_exceed` or `$likely_to_be_exceeded`.

classes/ActionScheduler_Abstract_QueueRunner.php
```diff
@@ -79,6 +92,115 @@ public function get_allowed_concurrent_batches() {
...
+	 *
+	 * Based on WP_Background_Process::memory_exceeded()
+	 *
+	 * @return bool
+	 */
+	protected function memory_exceeded() {
+
+		$memory_limit    = $this->get_memory_limit() * 0.90;
+		$current_memory  = memory_get_usage( true );
+		$memory_exceeded = ( $current_memory >= $memory_limit ) ? true : false;
```
JPry: This is another place where you can drop the `? true : false` portion, as well as the parentheses.

classes/ActionScheduler_Abstract_QueueRunner.php
```diff
@@ -22,6 +32,9 @@
 	 * @param ActionScheduler_QueueCleaner      $cleaner
 	 */
 	public function __construct( ActionScheduler_Store $store = null, ActionScheduler_FatalErrorMonitor $monitor = null, ActionScheduler_QueueCleaner $cleaner = null ) {
+
+		$this->created_time = microtime( true );
+
```
JPry: Minor: We're just setting class properties here, so the extra new lines before and after this new statement can be omitted.

classes/ActionScheduler_Compatibility.php
```diff
@@ -0,0 +1,79 @@
...
+	 * Only allows raising the existing limit and prevents lowering it.
+	 *
+	 * Wrapper for wp_raise_memory_limit(), added in WordPress v4.6.0
+	 *
+	 * @return bool|int|string The limit that was set or false on failure.
+	 */
+	public static function raise_memory_limit() {
+		if ( function_exists( 'wp_raise_memory_limit' ) ) {
+			return wp_raise_memory_limit( 'admin' );
+		} else {
```
JPry: Putting all of this logic in an `else` block isn't necessary, since the function will return early if the WP function exists. It should be removed from the `else` block to help with readability by reducing indentation.

classes/ActionScheduler_Compatibility.php
```diff
@@ -0,0 +1,79 @@
...
+	 * @link https://secure.php.net/manual/en/function.ini-get.php
+	 * @link https://secure.php.net/manual/en/faq.using.php#faq.using.shorthandbytes
+	 *
+	 * @param string $value A (PHP ini) byte value, either shorthand or ordinary.
+	 * @return int An integer byte value.
+	 */
+	public static function convert_hr_to_bytes( $value ) {
+		if ( function_exists( 'wp_convert_hr_to_bytes' ) ) {
+			return wp_convert_hr_to_bytes( $value );
+		} else {
```
JPry: Similar to what was mentioned below, enclosing this section in an `else` block is unnecessary.

classes/ActionScheduler_Compatibility.php
```diff
@@ -0,0 +1,79 @@
...
+					return false;
+				}
+			} elseif ( -1 === $wp_max_limit_int || $wp_max_limit_int > $current_limit_int ) {
+				if ( false !== @ini_set( 'memory_limit', $wp_max_limit ) ) {
+					return $wp_max_limit;
+				} else {
+					return false;
+				}
+			}
+		}
```
JPry: I don't *think* it's possible to get through this block without hitting one of the `return` conditions above, but just to be on the safe side I think we should include a `return false` at the very end of this function as a catch-all.

classes/ActionScheduler_Abstract_QueueRunner.php
thenbrent: @JPry then I'd have to make sure the `=` all align, which makes the diff harder to read. This property also serves a different purpose to the others, so I think it's justified it stands alone.

thenbrent:
Thanks for the review @JPry. I've addressed all of those issues and rebased in the fixes.

> I'm not opposed to this in principle. In relation to the discussion on Prospress/woocommerce-subscriptions#2851, would that change be better placed in a separate helper library?

Yes definitely, but we need this now in AS, so I don't the creation of the library to become a blocker for this to be added. We can migrate to use the library later once it's available.

PR #194, opened 2018-08-16
Fix remove_action() calls

### LINKED ISSUE #193: `remove_action()` can only accept 3 parameters, not 4
`ActionScheduler_FatalErrorMonitor::detach()` has a number of `remove_action()` calls which pass 4 parameters. `remove_action()` can only accept 3 parameters, not 4.

As raised in https://github.com/Prospress/action-scheduler/pull/191#discussion_r210065273

JPry:
Fixes #193.


thenbrent:
LGTM. Thanks for the quick PR. 👍

PR #202, opened 2018-09-13
Fix handling of empty timezone strings

### LINKED ISSUE #179: Manual timezone offsets trigger fatal errors

thenbrent:
This is an extension to #180. It takes all the original work from that PR to improve how we handle timezone strings. The big difference is that when the `'timezone_string'` option value in WordPress is an empty string, it will fallback to use the `'gmt_offset'` option value.

It won't attempt to map the `'gmt_offset'` value to a timezone string, as done in #180. This is necessary because the work on #180 proved that it is impossible to do that mapping reliably in all cases - some offsets do not have a matching timezone string in PHP; some have more than one timezone string, e.g for `UTC-12`, both `'Pacific/Kwajalein'` and `'Pacific/Fiji'` are valid timezone strings.

This also helps avoid falling back to `'UTC'` when it's impossible to find a matching timezone string, but we do have a valid UTC offset, meaning we _can_ determine the accurate local (WordPress) time.

This is the same approach used by WooCommerce core via `wc_string_to_datetime()`, so it's a proven method across a large user base.

Fixes #179.


classes/ActionScheduler_wpCommentLogger.php
```diff
@@ -26,7 +26,7 @@ public function log( $action_id, $message, DateTime $date = NULL ) {
 
 	protected function create_wp_comment( $action_id, $message, DateTime $date ) {
 		$comment_date_gmt = $date->format('Y-m-d H:i:s');
-		$date->setTimezone( ActionScheduler_TimezoneHelper::get_local_timezone() );
+		ActionScheduler_TimezoneHelper::set_local_timezone( $date );
```
JPry: There's a potential issue here:

```
PHP Fatal error:  Uncaught TypeError: Argument 1 passed to ActionScheduler_TimezoneHelper::set_local_timezone() must be an instance of ActionScheduler_DateTime, instance of DateTime given, called in /Users/jpry/Sites/prospress/content/plugins/action-scheduler/classes/ActionScheduler_wpCommentLogger.php on line 29 and defined in /Users/jpry/Sites/prospress/content/plugins/action-scheduler/classes/ActionScheduler_TimezoneHelper.php:16
Stack trace:
#0 /Users/jpry/Sites/prospress/content/plugins/action-scheduler/classes/ActionScheduler_wpCommentLogger.php(29): ActionScheduler_TimezoneHelper::set_local_timezone(Object(DateTime))
#1 /Users/jpry/Sites/prospress/content/plugins/action-scheduler/classes/ActionScheduler_wpCommentLogger.php(23): ActionScheduler_wpCommentLogger->create_wp_comment(2989, 'from shell!!!', Object(DateTime))
#2 phar:///usr/local/Cellar/wp-cli/2.0.1/bin/wp/vendor/wp-cli/shell-command/src/WP_CLI/REPL.php(37) : eval()'d code(1): ActionScheduler_wpCommentLogger->log(2989, 'from shell!!!', Object(DateTime))
#3 phar in /Users/jpry/Sites/prospress/content/plugins/action-scheduler/classes/ActionScheduler_TimezoneHelper.php on line 16
```

I see this as the main blocking issue, as seen in 5843134. While our own code doesn't natively pass an instantiated object, it's possible other code could be doing this.

I added 3f9017200bb73fa2907239186b5202b1c57aecdd as an attempt to bypass this issue. For this most part, this is simply updating docblocks to describe the current behavior. Can you think of any other areas (besides the custom tables extension) where this might be an issue?

JPry:
My comment seems to be hidden by GitHub, but here's the link: 
https://github.com/Prospress/action-scheduler/pull/202#discussion_r217540952

thenbrent:
Good catch @JPry! SHA: 3f9017200bb73 looks good, but I added SHA: b876d8c on top of it to make sure that we maintain backward compatibility for `create_wp_comment()`. That maps a `DateTime` to a `ActionScheduler_DateTime` when needed (and we don't need to worry about the timezone, as its overridden anyway).

> Can you think of any other areas (besides the custom tables extension) where this might be an issue?

There was at least one other instance - the `ActionScheduler_Store::get_scheduled_date_string_local()` method. That's why I went with converting the `DateTime` to a `ActionScheduler_DateTime`  in `ActionScheduler_TimezoneHelper::set_local_timezone()` rather than in `create_wp_comment()`.

`test_add_log_datetime()` is passing still with that patch (and the additions to the test).

Let me know if that looks good.

PR #217, opened 2018-12-05
Display cron expression in Admin schedule recurrence table field

alexminza:
Currently, `Non-repeating` is displayed for `ActionScheduler_CronSchedule` actions scheduled via `as_schedule_cron_action`.


thenbrent:
Thanks for the PR @alexminza. This LGTM and will included in v2.1.2 soon.

PR #221, opened 2018-12-19
Add action_scheduler_pre_init hook

JPry:
This hook will help with the ability to filter core Action Scheduler classes, by ensuring that there is a reliable action to hook into. This ensures that trying to use a hook priority of `0.5` isn't necessary.

Helps with Prospress/action-scheduler-custom-tables#49.


thenbrent:
LGTM. 👍

PR #239, opened 2019-02-26
Add folder structure

rrennick:
See #77 .

Add `WP_CLI` and `data-stores` sub-folders before adding the custom tables classes as most of the files will go into one of those two folders.

Move for abstract classes to `abstracts` folder.


thenbrent:
@rrennick LGTM. A nice improvement. You could also add `/schedules/` and `/actions/` folders to remove another 10 or so classes from the main folder. But I'm not going to block this PR being merged to wait for that.

PR #240, opened 2019-02-26
update register version to 3.0.0

rrennick:
See #77

Bump the AS registered version to ensure loading of this branch.


thenbrent:
Thanks @rrennick. Unit tests were failing, so I added SHA: 9c2e6a4 to update the version being registered to get them to pass.

Issue #222, opened 2019-01-10
Add scheduled action stats to WooCommerce System Status


thenbrent:
The WooCommerce System Status includes a section for `### Post Type Counts ###`, which includes `scheduled-action` posts. However, it would be very helpful to also receive a breakdown of the actions by status, e.g. how many are pending? how many failed?

Action Scheduler doesn't have any other dependency on WooCommerce, so this would have to be added only when WC is found to be active.

crstauf:
Submitted #243 which adds status counts to the WooCommerce system status report, as well as the oldest and newest action dates of each status.

rrennick:
This is in WC core now.

PR #243, opened 2019-03-04
Add scheduled action stats to WooCommerce System Status

crstauf:
PR for #222.

### DIFF
--- a/classes/ActionScheduler_AdminView.php
+++ b/classes/ActionScheduler_AdminView.php
@@ -30,6 +30,7 @@ class ActionScheduler_AdminView extends ActionScheduler_AdminView_Deprecated {
 
 			if ( class_exists( 'WooCommerce' ) ) {
 				add_action( 'woocommerce_admin_status_content_action-scheduler', array( $this, 'render_admin_ui' ) );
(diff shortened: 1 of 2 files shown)


thenbrent:
Thanks for taking the time to submit a PR @crstauf!

This looks good. I've added a few commits to tighten this up (refactor to use multiple smaller functions, add docblocks, remove unused HTML, add column headings etc.). With those tweaks, it's good to merge and will go out in the next minor version.

It now looks like: http://pic.pros.pr/02cded2dd7c2

crstauf:
@thenbrent The changes now prevent the counts from showing in the system status that is counted... is that preferable?

crstauf:
@thenbrent https://github.com/crstauf/action-scheduler/commit/c844ace4ce9d7988d3a046591bdd87486c3c0438 gives the most and cleanest data in both raw and tabular displays, though it's hacky.

thenbrent:
@crstauf ah that's why there was the 2nd cell! The hacky approach looks pretty good to me. I've created a PR for that: https://github.com/Prospress/action-scheduler/pull/252

PR #252, opened 2019-03-08
Fix system status export formatting

thenbrent:
For background see: https://github.com/Prospress/action-scheduler/pull/243#issuecomment-470748600

### DIFF
--- a/classes/ActionScheduler_wcSystemStatus.php
+++ b/classes/ActionScheduler_wcSystemStatus.php
@@ -97,12 +97,13 @@ class ActionScheduler_wcSystemStatus {
 		<table class="wc_status_table widefat" cellspacing="0">
 			<thead>
 				<tr>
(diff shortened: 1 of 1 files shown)


crstauf:
@thenbrent Not sure if intentional, but want to point out this was not merged into `master`, but rather into `issue_222`. `master` continues to have the issue with only the oldest scheduled date showing in the raw system status.

thenbrent:
Ah I didn't see that, thanks for pointing it out @crstauf 🙇

crstauf:
@thenbrent Just checking to see if this merge was corrected?

JPry:
@crstauf It looks like `master` does have these changes:

https://github.com/Prospress/action-scheduler/blob/076630e7a0d5d1fb47f0d77c095e0fb86b99842a/classes/ActionScheduler_wcSystemStatus.php#L97-L124

PR #253, opened 2019-03-08
Version 2.2.1

rrennick:
```
* Fix: Stake claim by group missing some actions when environment time zone not set
* Fix: Text domain on translated strings 
* Fix: Validate schedule meta when retrieving a scheduled action
* Add: Scheduled action stats to WooCommerce System Status
```

### DIFF
--- a/action-scheduler.php
+++ b/action-scheduler.php
@@ -25,21 +25,21 @@
  *
  */
 
-if ( ! function_exists( 'action_scheduler_register_2_dot_2_dot_0' ) ) {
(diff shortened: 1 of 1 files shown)


rrennick:
@thenbrent preparing this ahead of our chat next week.

thenbrent:
I just removed this one:

```
* Fix: Remove unused tooltip column from WC System Status
```

That issue was unreleased, so doesn't need to appear in the changelog.

Issue #175, opened 2018-07-09
WP CLI: Add timestamps to output


thenbrent:
When running Action Scheduler via WP CLI using a `crontab` task or similar automated process, it's handy to log the results into log files so you can keep track of what happened (especially if there are large queues of actions to process).

In these cases, it would be extremely helpful to have date/times displayed alongside the output.

For example, this is a log of errors running AS via WP CLI using Crontab:

![](https://cl.ly/0T1a3o312W3d/[3c7768a5db7cf32612012b01b9037a9b]_Screen%20Shot%202018-07-10%20at%208.47.12%20am.png)

And this is an example of valid logged output:

![](https://cl.ly/3N3o303P3s1k/Screen%20Shot%202018-07-10%20at%208.45.36%20am.png)

**EDIT**: I know it doesn't make a whole lot of sense to display timestamps etc. when there is a progress meter (i.e. WP CLI is being run manually). I'm not sure how to reconcile those different use cases, but I think it's far more likely in production WP CLI will be used via a schedule (i.e. Crontab) than manually.

crstauf:
I've found that when running via WP CLI manually, a timestamp would be helpful when reviewing the output at a later time. Perhaps a flag that can turn **on** timestamp output would be preferable to defaulting its display.

crstauf:
Submitted PR #254 with [notes](https://github.com/Prospress/action-scheduler/pull/254#issue-259715461).

crstauf:
This was merged in #267 (https://github.com/Prospress/action-scheduler/commit/076630e7a0d5d1fb47f0d77c095e0fb86b99842a).

rrennick:
> This was merged in #267

Closing.

PR #254, opened 2019-03-09
WP CLI: Add timestamps to output

crstauf:
PR for #175.

Notes:
1. Adds a flag `--t` to print timestamps.
1. While I attempted to be DRY, likely will need to be adjusted to fit Prospress preferred design.
1. No timestamp on progress bar, because the timestamp doesn't update.
1. Because effort is made to always use UTC, no timezone output on timestamp.
1. Timestamps are added outside of translation functions (ex: `__()`) when possible, but `date_i18n()` is used on the date's format. May need to be adjusted; not sure what's proper here.

### DIFF
--- a/classes/ActionScheduler_DateTime.php
+++ b/classes/ActionScheduler_DateTime.php
@@ -73,15 +73,4 @@ class ActionScheduler_DateTime extends DateTime {
 	public function getOffsetTimestamp() {
 		return $this->getTimestamp() + $this->getOffset();
 	}
-
-	/**
-	 * Get the timestamp for use in WP CLI output.
-	 *
-	 * @return string
-	 */
-	public function getCliTimestamp() {
-		if ( defined( 'AS_CLI_TIMESTAMP' ) && AS_CLI_TIMESTAMP ) {
-			return '[' . date_i18n( AS_CLI_TIMESTAMP, $this->format( 'U' ) ) . '] ';
-		}
-	}
 }
--- a/classes/ActionScheduler_DateTime.php
+++ b/classes/ActionScheduler_DateTime.php
@@ -80,6 +80,8 @@ class ActionScheduler_DateTime extends DateTime {
 	 * @return string
 	 */
 	public function getCliTimestamp() {
-		return '[' . date_i18n( 'Y-m-d H:i:s', $this->format( 'U' ) ) . '] ';
+		if ( defined( 'AS_CLI_TIMESTAMP' ) && AS_CLI_TIMESTAMP ) {
+			return '[' . date_i18n( AS_CLI_TIMESTAMP, $this->format( 'U' ) ) . '] ';
+		}
 	}
 }
--- a/classes/ActionScheduler_DateTime.php
+++ b/classes/ActionScheduler_DateTime.php
@@ -80,6 +80,6 @@ class ActionScheduler_DateTime extends DateTime {
 	 * @return string
 	 */
 	public function getCliTimestamp() {
-		return '[' . $this->format( 'Y-m-d H:i:s' ) . '] ';
+		return '[' . date_i18n( 'Y-m-d H:i:s', $this->format( 'U' ) ) . '] ';
 	}
 }
--- a/classes/ActionScheduler_DateTime.php
+++ b/classes/ActionScheduler_DateTime.php
@@ -73,4 +73,13 @@ class ActionScheduler_DateTime extends DateTime {
 	public function getOffsetTimestamp() {
 		return $this->getTimestamp() + $this->getOffset();
 	}
+
+	/**
+	 * Get the timestamp for use in WP CLI output.
+	 *
+	 * @return string
+	 */
+	public function getCliTimestamp() {
+		return '[' . $this->format( 'Y-m-d H:i:s' ) . '] ';
+	}
 }
--- a/classes/ActionScheduler_WPCLI_QueueRunner.php
+++ b/classes/ActionScheduler_WPCLI_QueueRunner.php
@@ -61,9 +61,9 @@ class ActionScheduler_WPCLI_QueueRunner extends ActionScheduler_Abstract_QueueRu
 		$too_many    = $claim_count >= $this->get_allowed_concurrent_batches();
 		if ( $too_many ) {
 			if ( $force ) {
-				WP_CLI::warning( as_get_datetime_object()->getCliTimestamp() . __( 'There are too many concurrent batches, but the run is forced to continue.', 'action-scheduler' ) );
+				WP_CLI::warning( as_get_formatted_cli_timestamp() . __( 'There are too many concurrent batches, but the run is forced to continue.', 'action-scheduler' ) );
 			} else {
-				WP_CLI::error( as_get_datetime_object()->getCliTimestamp() . __( 'There are too many concurrent batches.', 'action-scheduler' ) );
+				WP_CLI::error( as_get_formatted_cli_timestamp() . __( 'There are too many concurrent batches.', 'action-scheduler' ) );
 			}
 		}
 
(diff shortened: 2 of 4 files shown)


rrennick:
@crstauf There are 3 changes I would make:

- Print the timestamp at the beginning of the messages for readability using demarkation similar to OS logs (eg.`[2019-03-11 06:03:52] `). This would also allow the logic for the timestamp to be moved to a single function.
- Use a full parameter name `-time` instead of the short form. Linux inherited terse parameters from Unix but has added long form parameters to most commands in an effort to move away from them. Most WP CLI parameters are long form.
- Add support for a custom format `-time='m-d H:i:s'`.

crstauf:
@rrennick I understand the last two, but not the first... `ActionScheduler_DateTime::getCliTimestamp()` uses the format you suggest, and is a single function; please explain in more detail the change you'd like.

rrennick:
The format was just an example that I took from the debug log. Having the timestamp at the beginning of all the messages makes the log easier to read/follow.

`[06-Mar-2019 18:41:57 UTC] PHP Notice:  Undefined variable: example in /Sites/ron/wp-content/plugins/plugin-name/classes/data-store.php on line 123`

crstauf:
@rrennick Sorry, I should've mentioned that is the case already as well. Can you please explain further?

classes/ActionScheduler_DateTime.php
```diff
@@ -73,4 +73,13 @@ public function setTimezone( $timezone ) {
 	public function getOffsetTimestamp() {
 		return $this->getTimestamp() + $this->getOffset();
 	}
+
+	/**
+	 * Get the timestamp for use in WP CLI output.
+	 *
+	 * @return string
+	 */
+	public function getCliTimestamp() {
```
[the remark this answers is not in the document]
thenbrent: The method name also implies that it returns a Unix timestamp (i.e. an `int`) when it's actually returning a formatted date/time string, so I think it should be changed to reflect the return type.

classes/ActionScheduler_WPCLI_QueueRunner.php
```diff
@@ -16,6 +16,9 @@ class ActionScheduler_WPCLI_QueueRunner extends ActionScheduler_Abstract_QueueRu
 	/** @var \cli\progress\Bar */
 	protected $progress_bar;
 
+	/** @var bool */
+	protected $print_timestamp = false;
```
thenbrent: Similar to the `getCliTimestamp()` method, this variable is a misnomer. No timestamp is ever printed. It's a date/time string that's prepended, so the variable should be renamed `$print_date` or similar.

classes/ActionScheduler_WPCLI_QueueRunner.php
```diff
@@ -207,4 +213,16 @@ protected function stop_the_insanity( $sleep_time = 0 ) {
...
 		}
 	}
+
+	/**
+	 * Get timestamp if property 'print_timestamp' is truthy.
+	 *
+	 * @uses ActionScheduler_DateTime::getCliTimestamp()
+	 * @return null|string
+	 */
+	protected function maybe_get_timestamp() {
```
thenbrent: Similar to something @rrennick mentioned - prepending the date/time could be done in a separate function like this one to reduce the duplicate code prepending above (and make the approach consistent, e.g. always use placeholders of the prepend operator `.`).

`maybe_get_timestamp()` could be repurposed to something like `maybe_prepend_date( $message )` that both checks `$this->print_timestamp` and then prepends `as_get_datetime_object()->getCliTimestamp()` to the `$message` if needed.

Or ideally, this could be moved into a separate function outside the class like `as_maybe_prepend_date( $message, $print_date )` so that it can also be used by `ActionScheduler_WPCLI_Scheduler_command`.

classes/ActionScheduler_WPCLI_Scheduler_command.php
```diff
@@ -80,11 +84,14 @@ public function run( $args, $assoc_args ) {
...
 	 */
-	protected function print_total_actions( $total ) {
+	protected function print_total_actions( $total, $print_timestamp = false ) {
+		$timestamp = $print_timestamp ? as_get_datetime_object()->getCliTimestamp() : '';
+
 		WP_CLI::log(
 			sprintf(
 				/* translators: %d refers to how many scheduled taks were found to run */
-				_n( 'Found %d scheduled task', 'Found %d scheduled tasks', $total, 'action-scheduler' ),
+				_n( '%sFound %d scheduled task', '%sFound %d scheduled tasks', $total, 'action-scheduler' ),
```
thenbrent: `ActionScheduler_WPCLI_QueueRunner` adds the date/time outside the i18n function. This class is adding them _within_ the i18n functions. I think including them outside is the better approach. But regardless of which approach is taken, it should be consistent.

classes/ActionScheduler_WPCLI_Scheduler_command.php
```diff
@@ -54,23 +58,23 @@ public function run( $args, $assoc_args ) {
...
 			}
 		} catch ( Exception $e ) {
-			$this->print_error( $e );
+			$this->print_error( $e, $print_timestamp );
 		}
 
-		$this->print_total_batches( $batches_completed );
-		$this->print_success( $actions_completed );
+		$this->print_total_batches( $batches_completed, $print_timestamp );
+		$this->print_success( $actions_completed, $print_timestamp );
```
thenbrent: A separate function exclusively for prepending the date string will help reduce a lot of the code being added in the `print_*` methods. Especially if the `$print_timestamp` flag can be stored as a property on the class and accessed via that method.

thenbrent:
> I should've mentioned that is the case already as well. Can you please explain further?

@rrennick did you miss that `getCliTimestamp()` is adding formatting too, i.e. `[]` and the whitespace after the `]`? I missed that first time around.

classes/ActionScheduler_DateTime.php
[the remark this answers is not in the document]
crstauf: Implemented `as_get_formatted_cli_timestamp()` in 42a9506b83cc5043497a52f9663ebdecc50bed79.

classes/ActionScheduler_WPCLI_QueueRunner.php
crstauf: Removed in 814b6b2df56baadaa721061a8b9cbafa182f1a1a.

classes/ActionScheduler_WPCLI_QueueRunner.php
crstauf: Yeah, I would've preferred to have put this in a class that was shared between the runner and the scheduler, but because they didn't share anything, I wasn't sure what was preferred.

I've implemented `as_get_formatted_cli_timestamp()` in 42a9506b83cc5043497a52f9663ebdecc50bed79.

classes/ActionScheduler_WPCLI_Scheduler_command.php
crstauf: Yep, missed this one; fixed.

classes/ActionScheduler_WPCLI_Scheduler_command.php
crstauf: Agreed. I've optimized this in 814b6b2df56baadaa721061a8b9cbafa182f1a1a.

crstauf:
@thenbrent Thanks for the comments; I believe I've tended to them all.

@rrennick During some work tonight, I did not see any date strings anywhere but at the front of the line. When using a `WP_CLI` utility, they come after "Success:", "Warning:", or "Error:", but they're the first to be printed in the specified text.

crstauf:
@thenbrent Had some difficulty with using `date_i18n()`. When used, the time would be in UTC, but the timezone would read as local (I believe WordPress timezone).

```php
date_i18n( AS_CLI_TIMESTAMP, as_get_datetime_object()->format( 'U' ) );
```

Any suggestion?

rrennick:
> did you miss that getCliTimestamp() is adding formatting too, i.e. []

Yes, I had missed that as well.

crstauf:
Closed by #267.

PR #258, opened 2019-03-12
QueueRunner::process_action() - check $action before schedule_next_instance()

jeffstieler:
Only attempt to schedule the next instance of an action if it was successfully fetched.

If an exception is thrown during an action fetch (here: [#](https://github.com/Prospress/action-scheduler/blob/master/classes/ActionScheduler_Abstract_QueueRunner.php#L57)), `$action` will be undefined when `schedule_next_instance()` is called here [#](https://github.com/Prospress/action-scheduler/blob/master/classes/ActionScheduler_Abstract_QueueRunner.php#L66).

This PR adds a check to ensure that `$action` is a `ActionScheduler_Action`. Includes test.

### DIFF
--- a/classes/ActionScheduler_Abstract_QueueRunner.php
+++ b/classes/ActionScheduler_Abstract_QueueRunner.php
@@ -63,7 +63,10 @@ abstract class ActionScheduler_Abstract_QueueRunner extends ActionScheduler_Abst
 			$this->store->mark_failure( $action_id );
 			do_action( 'action_scheduler_failed_execution', $action_id, $e );
 		}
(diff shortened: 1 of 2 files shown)


thenbrent:
Thanks for the PR @jeffstieler. LGTM. 👍 

Out of curiosity, what caused the exception when fetching the action?

jeffstieler:
> Out of curiosity, what caused the exception when fetching the action?

We're directly deleting `scheduled-action` posts here: https://github.com/woocommerce/wc-admin/pull/1770/files#diff-86e3b9f240f17413506d0d741c55cdbaR46

Issue #256, opened 2019-03-10
Recurring actions admin view bug


crstauf:
Okay, I was 50/50 on whether I was doing something wrong or if this is a bug. 

Here's what I've got:

```php
add_action( 'init', function() {
	if ( !as_next_scheduled_action( 'recurring' ) )
		as_schedule_recurring_action( strtotime( '+1 minute' ), MINUTE_IN_SECONDS, 'recurring' );
} );
```

After one run, here are the events ([screenshot](https://user-images.githubusercontent.com/4573033/54089824-9567fa00-4343-11e9-9495-066b5218c84e.png)):

|Hook|Status|Scheduled Date|Log|
|---|---|---|---|
|recurring|Complete|2019-03-10 18:46:36 +0000|**2019-03-10 18:45:36 +0000**<br />action created<br />**2019-03-10 18:46:44 +0000**<br />action started<br />**2019-03-10 18:46:44 +0000**<br />action complete|
|recurring|Pending|2019-03-10 18:46:36 +0000|**2019-03-10 18:46:44 +0000**<br />action created|

As you can see, the scheduled timestamp is the same for both events.

I am relieved to say though, that this appears to be just an admin view bug, and not a functionality bug; after a few minutes, the events do run once a minute, and continue to recur, it's just the scheduled date is the scheduled time of the initial event:

<img width="1488" alt="Screen Shot 2019-03-10 at 2 56 44 PM" src="https://user-images.githubusercontent.com/4573033/54089924-c4cb3680-4344-11e9-9016-4109c1aa341b.png">

crstauf:
## Recurring Events
I'm registering a recurring event like-so:

```php
as_schedule_recurring_event( time(), DAY_IN_SECONDS, 'test' );
```

After one run, here's the list table:

|Hook|Status|Recurrence|Scheduled Date|Log|
|---|---|---|---|---|
|test|Complete|Every 1 day|2019-03-14 00:55:26 +0000|1. **2019-03-14 00:55:16 +0000**<br />action created<br />2. **2019-03-14 00:55:32 +0000**<br />action started<br />3. **2019-03-14 00:55:32 +0000**<br />action created|
|test|Pending|Every 1 day|2019-03-14 00:55:26 +0000|1. **2019-03-14 00:55:32 +0000**<br />action created|

[...]

PR #262, opened 2019-03-16
Recurring actions admin view bug

crstauf:
PR for #256.


thenbrent:
@crstauf I've submitted #333 to fix this one. I started that PR with the same commit you submitted here. You'll see from the changeset that the more I dived into the issue, the more I found needed fixing in the recurring schedules, so the scope went well beyond this one.

Thanks for the contribution here and first steps!

Issue #255, opened 2019-03-09
WP CLI: Schedule new action(s)


crstauf:
Adding an option for devs to easily schedule actions (in bulk) opens up exciting possibilities.

A few uses that come to mind:

1. Schedule infrequent functions (thinking cleanup, or processing tasks)
1. Creating new content
1. Updating WP core or plugins

Granted, there would have to be a corresponding action registered, which maybe makes scheduling events via the CLI irrelevant (because they could be scheduled via PHP). But I'd like to be able to `add_action()` (in a theme or plugin), and then be able to just leave it there, until I schedule it to run at whatever time.

Here's what I'm envisioning for the command:

`wp action-scheduler add <hook name> <scheduled date> <scheduled time> --args=<args in json> --interval=<interval> --limit=<number of actions to schedule>`

- `hook name`: name of the action to run, ex: `woocommerce-webhooks`
- `scheduled date`: date to schedule the action, ex: `2019-03-09`
- `scheduled time`: time of day to schedule the action, ex: `21:08:00`
- `args`: arguments for the action in JSON format, ex: `{"webhook_id":1,"order_id":100}`
- `interval`: seconds between each scheduled action, ex: `3600` (one hour)
- `limit`: number of actions to schedule, ex: `5` (one action every hour for five hours)

I realize there are probably more reasons to not implement than to implement, so I defer to the wisdom of the masses. Just an idea I had that I found potentially exciting and powerful for easing tasks and workflows.

thenbrent:
I can see this being especially useful for testing scheduled actions (especially in bulk). I like the idea and would review/merge a PR for it. 👍

crstauf:
@thenbrent Submitted PR #263. Please review and let me know of changes you'd like to see. 😄

rrennick:
CLI subcommands for `add`, `cancel`, `delete` would be great additions. Some additions to the items listed in the description:

- async actions
- cron schedules
- recurring schedule (in addition to the proposed schedule X of Y every Z)
- group field
- I'd make schedule a single field. WP CLI works with spaces in field names when the parameter is single quoted.

`--args=` would be handled as json automatically with `--format=json`

Konamiman:
Hi, we know that this is an old one but we still think this is an interesting idea and we see that some work has actually been done outside of this repository. We'll be glad to take a look at any submitted pull request that implements this feature in Action Scheduler itself.

crstauf:
[...]

PR #270, opened 2019-03-28
WP CLI: 'add' command

crstauf:
Revised PR for #255.


thenbrent:
I'll leave review of this up to @JPry. Just a quick not to say I've changed the base branch on this to `release/2.3.0`, as it will go out with that version.

crstauf:
Doing some refactoring, which will likely implement this command differently.

Issue #272, opened 2019-03-28
Add deprecated ActionScheduler_WPCLI_Scheduler_command class


thenbrent:
Amidst some very nice refactoring, #267 removed the `ActionScheduler_WPCLI_Scheduler_command` class. This class was a public API and it could be instantiated or extended by 3rd party code. As a result, it needs to continue to exist to maintain backward compatibility unless we release those patches in a major release, e.g. v3.0.0. I'd prefer to ship them sooner than that, as they are good improvements.

A new file for the `ActionScheduler_WPCLI_Scheduler_command` class can be added in the `/deprecated/` directory for this purpose. The `run()` method should still be callable and map to `ActionScheduler_WPCLI_Command_Run:: execute()` as well as calling `_deprecated_function()` to notify any developers that it has been deprecated and will be removed in a future major version.

/CC @JPry @crstauf .

crstauf:
@thenbrent Just pushed https://github.com/crstauf/action-scheduler/commit/03642ee3dd34b1eb076b355da6ecfea9cfd3ec5b, which adds `ActionScheduler_WPCLI_Scheduler_command` into `/deprecated/` directory.

The info you provided in the issue's description certainly helped, though I'm not 100% certain on some things; please review:

1. how the file is included
1. phpDoc blocks
1. `_deprecated_function()` call is correct

I've tested by adding the following code to `ActionScheduler_WPCLI`:

```php
public function run_legacy( $args, $assoc_args ) {
	$command = new ActionScheduler_WPCLI_Scheduler_command();
	$command->run( $args, $assoc_args );
}
```

and running `wp action-scheduler run_legacy`.

thenbrent:
Hey @crstauf, thanks for jumping on that! crstauf@03642ee looks good, but I think `ActionScheduler_WPCLI_Scheduler_command ` will have to extend `ActionScheduler_WPCLI_Command_Run`  to be fully backward compatible.

The main thing we're trying to avoid are fatal errors by anyone extending `ActionScheduler_WPCLI_Scheduler_command ` or calling a method of it, and assuming that class / method will exist. That's why we need it to extend `ActionScheduler_WPCLI_Command_Run` - so that it continues to have all the methods it used to have.

Having `_deprecated_function()` in `run()` is good. I think `_deprecated_function()` should also exist in `ActionScheduler_WPCLI_Scheduler_command::__construct()` so that it's triggered for anything instantiating it too.

crstauf:
[...]

PR #279, opened 2019-03-30
WP CLI: add 'info' command

crstauf:
## Command:
`wp action-scheduler info`

## To do:
- [x] get number of past events
- [ ] count by hook

## Screenshots
Screen Shot 2019-03-30 at 2 40 39 AM

Screen Shot 2019-03-30 at 2 43 18 AM

### DIFF
--- a/classes/ActionScheduler.php
+++ b/classes/ActionScheduler.php
@@ -110,6 +110,7 @@ abstract class ActionScheduler {
 
 		if ( defined( 'WP_CLI' ) && WP_CLI ) {
 			WP_CLI::add_command( 'action-scheduler', 'ActionScheduler_WPCLI' );
+			require_once( self::plugin_path('deprecated/ActionScheduler_WPCLI_Scheduler_command.php') );
 		}
 	}
 
--- a/classes/ActionScheduler_WPCLI_Command_Info.php
+++ b/classes/ActionScheduler_WPCLI_Command_Info.php
@@ -1,26 +1,45 @@
 <?php
 
+/**
+ * Action Scheduler WP CLI command to display system status.
+ */
 class ActionScheduler_WPCLI_Command_Info extends ActionScheduler_Abstract_WPCLI_Command {
+
+	/**
+	 * @var ActionScheduler_Store
+	 */
+	protected $store = null;
 
 	/**
 	 * Execute command.
 	 */
 	public function execute() {
+		$this->store = ActionScheduler::store();
+
 		$this->completed_actions();
 		$this->pending_actions();
 		$this->overdue_actions();
 	}
 
+	/**
+	 * Print count of completed actions.
+	 */
 	protected function completed_actions() {
 		$completed_actions = (int) $this->store->query_actions( array( 'status' => $this->store::STATUS_COMPLETE ), 'count' );
 		$this->log( 'Completed actions: ' . $completed_actions );
 	}
 
+	/**
+	 * Print count of pending actions.
+	 */
 	protected function pending_actions() {
 		$pending_actions = (int) $this->store->query_actions( array( 'status' => $this->store::STATUS_PENDING ), 'count' );
 		$this->log( 'Pending actions: ' . $pending_actions );
 	}
 
+	/**
+	 * Print count of overdue actions.
+	 */
 	protected function overdue_actions() {
 		$overdue_actions = (int) $this->store->query_actions( array( 'status' => $this->store::STATUS_PENDING, 'date' => as_get_datetime_object()->format( 'Y-m-d H:i:s' ) ), 'count' );
 		$this->log( 'Overdue actions: ' . $overdue_actions );
--- a/classes/ActionScheduler_WPCLI_Command_Info.php
+++ b/classes/ActionScheduler_WPCLI_Command_Info.php
@@ -36,7 +36,6 @@ class ActionScheduler_WPCLI_Command_Info extends ActionScheduler_Abstract_WPCLI_
 		$args = array(
 			'status' => 'pending',
 			'date' => as_get_datetime_object()->format( 'Y-m-d H:i:s' ),
-			'date_compare' => '<=',
 		);
 
 		if ( !empty( $hooks ) ) {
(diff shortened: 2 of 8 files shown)


classes/ActionScheduler_WPCLI_Command_Info.php
```diff
@@ -6,25 +6,145 @@ class ActionScheduler_WPCLI_Command_Info extends ActionScheduler_Abstract_WPCLI_
...
+	 * @param string $hooks
+	 */
+	protected function past_due( $hooks ) {
 		$store = ActionScheduler::store();
+		$columns = explode( ',', \WP_CLI\Utils\get_flag_value( $this->assoc_args, 'columns', 'hook,total' ) );
+
+		if ( true === $hooks ) {
+			global $wpdb;
+
+			$query = $wpdb->prepare( "SELECT DISTINCT( post_title ) FROM $wpdb->posts WHERE post_type = %s", ActionScheduler_wpPostStore::POST_TYPE );
```
rrennick: You can't use direct DB queries in WP CLI commands because AS supports custom data stores. This query would only work with the default data store.

crstauf: Thanks, @rrennick! 

https://github.com/Prospress/action-scheduler/pull/279/commits/d418a8681949934614d7d087db7732ba076008a0 introduces `ActionScheduler_wpPostStore::action_hooks()`.

classes/ActionScheduler.php
```diff
@@ -110,6 +110,7 @@ public static function init( $plugin_file ) {
 
 		if ( defined( 'WP_CLI' ) && WP_CLI ) {
 			WP_CLI::add_command( 'action-scheduler', 'ActionScheduler_WPCLI' );
+			require_once( self::plugin_path('deprecated/ActionScheduler_WPCLI_Scheduler_command.php') );
```
rrennick: If you change the class name to `ActionScheduler_WPCLI_Scheduler_Command_Deprecated` then the autoloader will load it automatically if the class is used.

deprecated/ActionScheduler_WPCLI_Scheduler_command.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+/**
+ * Class ActionScheduler_WPCLI_Scheduler_command
+ *
+ * Retain public API for backwards compatibility.
+ */
+class ActionScheduler_WPCLI_Scheduler_command {
```
rrennick: The `c` in `Command` should be upper case.

rrennick:
@thenbrent The new data store functions in this PR should have mock equivalents in the base store class. How has adding new store functions been handled in the past?

classes/ActionScheduler.php
crstauf: @rrennick This change is from #267 (please see https://github.com/Prospress/action-scheduler/pull/267#issuecomment-477174034).

deprecated/ActionScheduler_WPCLI_Scheduler_command.php
crstauf: @rrennick This change is from #267 (please see https://github.com/Prospress/action-scheduler/pull/267#issuecomment-477174034).

classes/ActionScheduler.php
rrennick: The code hasn't been released yet so items that were missed in the first review can be fixed without negative impact.

crstauf: @rrenick The code hasn't even been merged yet, which is why it appears in this PR, even though it's not related to this PR.

Additionally, changing the class name will break backwards compatibility (for direct references from other sources).

deprecated/ActionScheduler_WPCLI_Scheduler_command.php
crstauf: @rrennick The class name in `master` has a lower `c`... this class is being added for backwards compatibility (#272), so matching the class name is needed.

classes/ActionScheduler.php
rrennick: Either the autoloader could be updated or a stub could be added that mitigated the issue. Currently, the only `require`s for classes is in the main plugin file. It would be good to keep it that way.

crstauf: Moving convo to https://github.com/Prospress/action-scheduler/pull/267#issuecomment-479562559.

crstauf:
Started work on `get` command, and I think that this command will likely be closed and merged into `get` command:
- `get` is clearer (verb) than `info` (noun)
- minimize redundancy (DRY)

thenbrent:
> How has adding new store functions been handled in the past?

Unless they are first released in 3.0.0, any new methods need to be backward compatible. That means adding methods to `ActionScheduler_Store` that can perform the required operations (not just stubs of the method name).

It looks like there won't be a problem with `action_dates()`, as it uses existing store methods, so it can be moved to `ActionScheduler_Store`, and inherited by all child classes.

`action_hooks()` is a harder one, because at least in the version included in `ActionScheduler_wpPostStore`, it's doing new queries. A different version of it could be added to `ActionScheduler_Store` that uses existing APIs to maintain backward compatibility, but it wouldn't be very performant (it might have to do something like getting all actions and finding the unique hooks in memory/application level rather than MySQL depending on what `query_actions()` allows).

deprecated/ActionScheduler_WPCLI_Scheduler_command.php
crstauf: Apparently, class names are not case-sensitive. 🤔

crstauf: Capitalized "C" in https://github.com/Prospress/action-scheduler/pull/280/commits/af18ec5a465cebcd3cf4c4028e24e8c93680443b.

crstauf:
Confirmed; `get` will replace the previous version of `info`, and info will be simplified to more of a report... see https://github.com/Prospress/action-scheduler/pull/279/commits/6282619ce645237b95922453f75964a1529b2ac5.

crstauf:
Re https://github.com/Prospress/action-scheduler/pull/279#issuecomment-479700828, `action_dates()` and `action_hooks()` are no longer needed for the `info` command (they've been removed in https://github.com/Prospress/action-scheduler/pull/279/commits/4212b0375a29e3f0a1ea30152d938c858b995e8d).

`action_hooks()` is used in `get`, so will pick the convo up there.

rrennick:
Updated the base for this PR to `version_3_1_0`.

crstauf:
Closing in favor of https://github.com/crstauf/action-scheduler-cli.

PR #282, opened 2019-04-08
WP CLI: add `get` command

crstauf:
`wp action-scheduler get`

### DIFF
--- a/classes/ActionScheduler.php
+++ b/classes/ActionScheduler.php
@@ -110,6 +110,7 @@ abstract class ActionScheduler {
 
 		if ( defined( 'WP_CLI' ) && WP_CLI ) {
 			WP_CLI::add_command( 'action-scheduler', 'ActionScheduler_WPCLI' );
+			require_once( self::plugin_path('deprecated/ActionScheduler_WPCLI_Scheduler_command.php') );
 		}
 	}
 
--- a/classes/ActionScheduler_Abstract_WPCLI_Command.php
+++ b/classes/ActionScheduler_Abstract_WPCLI_Command.php
@@ -27,6 +27,11 @@ abstract class ActionScheduler_Abstract_WPCLI_Command {
 	protected $columns = array();
 
 	/**
+	 * @var ActionScheduler_Store
+	 */
+	protected $store = null;
+
+	/**
 	 * Construct.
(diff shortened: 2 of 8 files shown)


crstauf:
@thenbrent 

See https://github.com/Prospress/action-scheduler/pull/279#issuecomment-479700828.

> `action_hooks()` is a harder one, because at least in the version included in `ActionScheduler_wpPostStore`, it's doing new queries. A different version of it could be added to `ActionScheduler_Store` that uses existing APIs to maintain backward compatibility, but it wouldn't be very performant (it might have to do something like getting all actions and finding the unique hooks in memory/application level rather than MySQL depending on what `query_actions()` allows).

Can you please make a recommendation on what `ActionScheduler_Store::action_hooks()` should look like? I'm not sure how to start, because `ActionScheduler_Store::query_actions()` has no definition.

thenbrent:
> Can you please make a recommendation on what ActionScheduler_Store::action_hooks() should look like? I'm not sure how to start, because ActionScheduler_Store::query_actions() has no definition.

Sure! I was thinking something like this:

```php
	/**
	 * Get all unique action hooks.
	 *
	 * @return string[]
	 */
	public function action_hooks() {

		$all_actions = as_get_scheduled_actions( array(
			'per_page' => -1,
		) );

		$all_hooks = array();
		foreach ( $all_actions as $action ) {
			if ( ! in_array( $action->get_hook(), $all_hooks ) ) {
				$all_hooks[] = $action->get_hook();
			}
		}

		return $all_hooks;
	}
```

I used `as_get_scheduled_actions()` there instead of `$this->query_actions()`, because it saves the step in instantiating the action (or getting the object vars for it).

It would be less memory intensive to use `$this->query_actions()` and instantiate each individually, but also much slower. That would possibly be the better approach, as long as the result is cached (and the cache invalidated whenever a new action is saved/deleted). 

> because ActionScheduler_Store::query_actions() has no definition.

Ah good point. `as_get_scheduled_actions()` has a docblock to fill in the blanks.

rrennick:
Updated the base for this PR to `version_3_1_0`.

crstauf:
Closing in favor of https://github.com/crstauf/action-scheduler-cli.

PR #292, opened 2019-04-18
add exception handler for invalid schedule in make_action_from_post()

rrennick:
This PR adds an exception handler for the WP post store retrieving the schedule from the post content. It's possible for the post content to be json_encoded and not an array.

@mikejolley Can you take a look at this as well?


mikejolley:
@rrennick this looks good and looks like it will mitigate the issue. Just to point out though @JPry was talking about logging these somewhere, and noted that there are try/catches in place in other places which call this method. However, I found a few which don't. 

So the choice seems to be try/catch here, or try/catch higher up when fetching these jobs.

rrennick:
> Just to point out though @JPry was talking about logging these somewhere

I think it would be good to log it in the scheduled action log in addition to adding it to an error log. In that case, it should be done above the data store so it's logged regardless of the data store being used.

rrennick:
@jPry Thanks, I added the hook and an action to the base logger class to log to the action that it failed to fetch.

PR #297, opened 2019-04-24
mark actions with invalid schedules as failures

### LINKED ISSUE #296: Broken schedule causes multiple instances of same hook
This is a followup to #291. While #292 does handle the exception, the action is left in a pending state. If the scheduling function uses `next()` to determine whether to add a recurring schedule of the hook then every time the check runs another instance of the hook will be scheduled.

### To reproduce 

- Disable the SA queue runner by installing https://github.com/Prospress/action-scheduler-disable-default-runner
- Add the following plugin code to a test plugin

```
	if ( ! function_exists( 'as_schedule_recurring_action' ) ) {
		return;
	}

	if ( ! as_next_scheduled_action( 'recurring_test' ) ) {
		as_schedule_recurring_action( time(), WEEK_IN_SECONDS, 'recurring_test', array(), 'test-actions' );
	}
});
```
- Check that the pending action is created
- Update the post_content field of that post to "valid_json_encoded_string" or {"data":"invalid_json"}}.
- Refresh the pending actions screen a few times
- You should see a new recurring action for each refresh

rrennick:
Fixes #296 .

This PR marks actions with invalid recurring schedules as failed so they do not interfere with replacement scheduled actions.

Follow the reproduce steps outlined in the issue description. The action with the invalid schedule should be moved to the failed list and only one replacement action should be created.

### DIFF
--- a/classes/ActionScheduler_Store.php
+++ b/classes/ActionScheduler_Store.php
@@ -188,6 +188,13 @@ abstract class ActionScheduler_Store {
 
 	public function init() {}
 
+	/**
+	 * Mark an action that failed to fetch correctly as failed.
+	 *
+	 * @since 2.2.6
+	 *
+	 * @param int $action_id The ID of the action.
+	 */
 	public function mark_failed_fetch_action( $action_id ) {
 		self::$store->mark_failure( $action_id );
 	}
--- a/classes/ActionScheduler_Store.php
+++ b/classes/ActionScheduler_Store.php
@@ -188,13 +188,36 @@ abstract class ActionScheduler_Store {
 
 	public function init() {}
 
+	public function mark_failed_fetch_action( $action_id ) {
+		self::$store->mark_failure( $action_id );
+	}
+
+	/**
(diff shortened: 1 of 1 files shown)


classes/ActionScheduler_Store.php
```diff
@@ -188,13 +188,43 @@ public function get_status_labels() {
...
+		self::$store->mark_failure( $action_id );
+	}
+
+	/**
+	 * Add base hooks
+	 *
+	 * @since 2.2.6
+	 */
+	protected static function hook() {
+		add_action( 'action_scheduler_failed_fetch_action', array( self::$store, 'mark_failed_fetch_action' ), 20 );
```
JPry: Should this be just `self` rather than `self::$store`? Same question for the `unhook()` method below.

rrennick: I intended for the action to run using the class of the `$store` instance rather than the base class. That allows an inheriting class to override the hook/unhook functions without having to create a second set of handlers.

PR #299, opened 2019-05-01
include actions with no group when sorting actions by group

### LINKED ISSUE #298: Order by group excludes actions with no group
When using `as_get_scheduled_actions` with `orderby` => `group`, the WP Post store excludes scheduled actions that have the default `''` group. 

The `get_query_actions_sql ` function uses `INNER JOIN` https://github.com/Prospress/action-scheduler/blob/master/classes/ActionScheduler_wpPostStore.php#L304:L306 which requires the scheduled action to have a record in the term relationships table to be included in the query results.

### Steps to reproduce
1. Create a test scheduled action with no group
1. Call `as_get_scheduled_actions( array( 'orderby' => 'group', 'order' => 'asc' ) )`
1. Call `as_get_scheduled_actions( array( 'orderby' => 'group', 'order' => 'desc' ) )`
1. The new action won't be included in either result set

rrennick:
Fixes #298 

This PR changes the WP Post Store get actions SQL for `orderby` => `group` queries to use `LEFT JOIN` to allow actions without a group to be included in the query results.

### Steps to test
1. Follow the steps to reproduce outlined in the issue
1. The new action should be returned in the results of step 2.

### DIFF
--- a/classes/ActionScheduler_wpPostStore.php
+++ b/classes/ActionScheduler_wpPostStore.php
@@ -308,11 +308,8 @@ class ActionScheduler_wpPostStore extends ActionScheduler_Store {
 			$sql .= " INNER JOIN {$wpdb->term_relationships} tr ON tr.object_id=p.ID";
 			$sql .= " INNER JOIN {$wpdb->term_taxonomy} tt ON tr.term_taxonomy_id=tt.term_taxonomy_id";
 			$sql .= " INNER JOIN {$wpdb->terms} t ON tt.term_id=t.term_id";
-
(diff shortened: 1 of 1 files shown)


thenbrent:
@rrennick I added SHA: a59011b to remove the redundant `! empty()` check, otherwise this LGTM.

I always think when sorting by a value, the `NULL` results should be at the end of the set, regardless of whether the order is `ASC` or `DESC`, but I don't think that needs to be a blocker here, especially given we're moving to custom tables.

PR #300, opened 2019-05-01
add 1 second delta to unit tests comparing sequential generated timestamps

rrennick:
This PR adds a 1 second delta to the two unit tests that frequently fail due to two sequentially generated timestamps falling across a (time) second boundary.


JPry:
@rrennick This looks good, but oddly there are a couple of tests that are still failing. Would you mind double-checking those?

rrennick:
@JPry Thanks :)

> but oddly there are a couple of tests that are still failing.

The allowed delta value is the 4th parameter vs the 3rd. I've updated the PR.

Issue #303, opened 2019-05-08
Add function_exists( 'add_action' ) check to initialization file


salcode:
## Loading Action Scheduler via Composer

When loading Action Scheduler via Composer, I need to autoload `vendor/prospress/action-scheduler/action-scheduler.php` which Composer supports natively 😃 .

    "autoload": {
        "files": [
            "vendor/prospress/action-scheduler/action-scheduler.php"
        ]
    },

This works great when `action-scheduler.php` is loaded within the context of WordPress.

## The Problem with Autoloading action-scheduler.php

If a project is also using a PHP based development tool (e.g. [PHP_CodeSniffer (phpcs)](https://github.com/squizlabs/PHP_CodeSniffer)), that tool runs outside of the context of WordPress. Because of this when `vendor/prospress/action-scheduler/action-scheduler.php` is autoloaded, a Fatal Error is thrown because `add_action()` is not a defined function ☹️ .

## Suggested Solution

By adding a `function_exists( 'add_action' )` check before calling `add_action()`, we could skip this fatal error when a PHP based development tool is run and `action-scheduler.php` is autoloaded via Composer.

Thanks for this great library (and regardless of how I load it) it is a big help on my current project.

Thanks!

salcode:
For reference: this same problem occurs when loading the CMB2 library as a composer dependency (see CMB2 [Issue 1270](https://github.com/CMB2/CMB2/issues/1270) *Add function_exists( 'add_action' ) check to initialization file*)

johnbillion:
IMO this isn't the correct solution. A WordPress plugin should be able to assume that the WordPress plugin API functions are available, otherwise every plugin in the WordPress ecosystem needs to be updated with `function_exists()` calls to make it compatible with Composer autoloading.

The mistake is that your project is autoloading a WordPress plugin. By its nature a WordPress plugin entrypoint file will almost always have side effects (most commonly by adding actions and filters) instead of simply declaring symbols, which makes it unfit and inappropriate for autoloading in the context of Composer.

Most established Composer-based project frameworks recommend to load WordPress plugins as actual plugins or mu-plugins, instead of autoloading them directly via Composer, for exactly the reason you've identified. For example [Bedrock has its own autoloader](https://roots.io/bedrock/docs/mu-plugins-autoloader/) and any must-use plugins are loaded at the correct point instead of being directly autoloaded in the Composer configuration.

salcode:
[...]

PR #304, opened 2019-05-08
Add check that function add_action() exists

salcode:
By checking that the function add_action() exists, before we call it, we
can avoid throwing a fatal error when a PHP tool is run (e.g. phpcs)
even if we are loading `action-scheduler.php` via Composer's autoload.

See #303

### DIFF
--- a/action-scheduler.php
+++ b/action-scheduler.php
@@ -25,7 +25,7 @@
  *
  */
 
-if ( ! function_exists( 'action_scheduler_register_2_dot_2_dot_5' ) ) {
+if ( ! function_exists( 'action_scheduler_register_2_dot_2_dot_5' ) && function_exists( 'add_action' ) ) {
 
 	if ( ! class_exists( 'ActionScheduler_Versions' ) ) {
 		require_once( 'classes/ActionScheduler_Versions.php' );


thenbrent:
Thanks for the background on this issue in #303 @salcode, and the PR.

This patch fixes the fatal error, but won't it also mean that Action Scheduler is never loaded when using Composer? (Because `action_scheduler_register_2_dot_2_dot_5()` will never be called.)

salcode:
@thenbrent 

Thanks for taking the time to look at this PR and for your response.

For context, I'm working on a WordPress plugin that loads Action Scheduler via Composer. In my tests this works (and I've not been able to come up with a scenario where this would not work but I certainly acknowledge I may not have thought of everything).

## Why  the `add_action()` function (usually) exists

Since the plugin is loaded by WordPress, the Composer autoloader in my plugin (which loads Action Scheduler) is loaded **after** the `add_action()` function has been defined by the WordPress core files.

## Preventing a bad situation

In any scenario where

```
add_action( 'plugins_loaded', 'action_scheduler_register_2_dot_2_dot_5', 0, 0 );
```

is going to execute successfully, our check that `function_exists( 'add_action' )` will be true, which will allow the call to occur.

It is only in the case where calling `add_action()` is going to cause a fatal error (because the function does not exist), that we skip the call. 

This additional conditional should only prevent bad calls to `add_action()`, not good ones.

thenbrent:
Thanks for this PR and your patience in convincing me it is the (only) feasible approach @salcode. Much appreciated!

salcode:
@thenbrent Thanks for all the time you took working with me on this, I really appreciate it.

Issue #293, opened 2019-04-23
Running recurring action scheduled in the past is rescheduled for Now


rrennick:
See https://github.com/woocommerce/woocommerce/issues/23414#issuecomment-485895442 for steps to reproduce.

This affects the WP post store `save_action` function when it's passed a `$scheduled_date` an interval in the future. The existing actions scheduled date is used instead of the passed `$scheduled_date`.

thenbrent:
In line with comments on #327, I'm reopening this so I can investigate it further.

thenbrent:
I've opened #330 to address the root issue here. I'm going to close this one now in lieu of that.

PR #309, opened 2019-05-15
add check for interval_in_seconds method

### LINKED ISSUE #308: Call to undefined method ActionScheduler_CronSchedule::interval_in_seconds()
`PHP Fatal error:  Uncaught Error: Call to undefined method ActionScheduler_CronSchedule::interval_in_seconds() in /var/www/html/wp-content/plugins/test/includes/vendor/prospress/action-scheduler/classes/ActionScheduler_wpPostStore.php:25
Stack trace:
#0 /var/www/html/wp-content/plugins/test/includes/vendor/prospress/action-scheduler/classes/ActionScheduler_Abstract_QueueRunner.php(82): ActionScheduler_wpPostStore->save_action(Object(ActionScheduler_Action), Object(ActionScheduler_DateTime))
#1 /var/www/html/wp-content/plugins/test/includes/vendor/prospress/action-scheduler/classes/ActionScheduler_Abstract_QueueRunner.php(68): ActionScheduler_Abstract_QueueRunner->schedule_next_instance(Object(ActionScheduler_Action))
#2 /var/www/html/wp-content/plugins/test/includes/vendor/prospress/action-scheduler/classes/ActionScheduler_QueueRunner.php(80): ActionScheduler_Abstract_QueueRunner->process_action('105893')
#3 /var/www/html/wp-content/plugins in /var/www/html/wp-content/plugins/test/includes/vendor/prospress/action-scheduler/classes/ActionScheduler_wpPostStore.php on line 25
`

I note that a similar issue was noted/fixed? https://github.com/Prospress/action-scheduler/issues/133

Is this related?

rrennick:
Fixes #308

This PR adds a check that the `interval_in_seconds` method exists which should have been included in https://github.com/Prospress/action-scheduler/commit/88359f0a2bb4438bbe6a1227d9918225f6b14cb6 .

### DIFF
--- a/classes/ActionScheduler_wpPostStore.php
+++ b/classes/ActionScheduler_wpPostStore.php
@@ -21,7 +21,7 @@ class ActionScheduler_wpPostStore extends ActionScheduler_Store {
 			$post_id = $this->save_post_array( $post_array );
 			$schedule = $action->get_schedule();
 
-			if ( ! is_null( $scheduled_date ) && $schedule->is_recurring() ) {
+			if ( ! is_null( $scheduled_date ) && $schedule->is_recurring() && method_exists( $schedule, 'interval_in_seconds' ) ) {
 				$schedule = new ActionScheduler_IntervalSchedule( $scheduled_date, $schedule->interval_in_seconds() );
 			}
 


thenbrent:
With this approach, I think the same bug fixed with #294 (reported in issue #293) will persist for recurring _cron_ actions. That's why I've proposed #327 and instead think we should get to the root of schedule data issue, which I've discussed in #330.

PR #314, opened 2019-06-10
allow queue cleaner to be disabled via filter

### LINKED ISSUE #313: Allow disabling the queue cleaner
While you would not want to permanently disable the queue cleaner there could be situations where a site owner might want to temporarily disable it. Currently, this can only be accomplished by filtering the lifespan to a large time frame.

The use case that I have/had is in testing the migration script in #259, I had a DB backup with a large number of scheduled actions that I was restoring to re-test the migration process. As the backup aged, some scheduled actions were being deleted by the cleaner while the migration was running.

I'll do a PR to add a check on the filtered value and if it's 0 then skip the cleaning process.

rrennick:
Fixes #313

This PR disables the queue cleaner when the `action_scheduler_retention_period` returns a `null`.

### Testing instructions

- Add a filter `add_filter( 'action_scheduler_retention_period', '__return_null' );`
- Modify a completed scheduled action in the database to have a modified date older than 30 days.
- Run the action scheduler through WP CLI `$ wp action-scheduler run`
- Check that the modified action is still in the database.

Note: `__return_zero` is used in the unit tests to clear all the completed/cancelled actions from the queue. The alternative of `null` is used.

### DIFF
--- a/classes/ActionScheduler_QueueCleaner.php
+++ b/classes/ActionScheduler_QueueCleaner.php
@@ -31,7 +31,11 @@ class ActionScheduler_QueueCleaner {
 
 	public function delete_old_actions() {
 		$lifespan = apply_filters( 'action_scheduler_retention_period', $this->month_in_seconds );
-		$cutoff = as_get_datetime_object($lifespan.' seconds ago');
+		if ( null === $lifespan ) {
+			return;
+		}
+
+		$cutoff = as_get_datetime_object( $lifespan . ' seconds ago' );
 
 		$statuses_to_purge = array(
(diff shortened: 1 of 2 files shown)


rrennick:
Due to the imprecision of int -> float conversions there are values where a 1 second difference produces a 1.0000000001 floating point difference which is greater than 1.0. This causes the 1 second delta in https://github.com/Prospress/action-scheduler/pull/314/commits/ef82ebcba3e66a9656949e0df361cd678ca0c6d2 to fail. Bumping to 2 will eliminate this intermittent occurance.

thenbrent:
LGTM. 👍

Issue #319, opened 2019-06-25
Add Support for Async Actions


thenbrent:
In a similar vein to #131, Action Scheduler needs a new async action type. That means a new API for adding async actions to the queue and a new async queue runner for processing them.

These action types can be used to speed up processing of time sensitive tasks (vs. background tasks, like recurring payments) and provide a 3rd alternative for queue processing in addition to WP Cron and WP CLI.

### The Need for Async Actions

The `WC_Action_Queue` class and `WC_Queue_Interface` were both deliberately created with an `add()` method that is designed to:

> Enqueue an action to run one time, as soon as possible

Note that any mention of "scheduling" in that description is replaced with "as soon as possible". That's is the "schedule" for async actions - as close to now as possible.

While Action Scheduler is used by the `WC_Action_Queue::add()` method like this: `as_schedule_single_action( time(), $hook, $args, $group )`, that just means by default, the action will run as soon as a WP Cron request claims a queue, and processes any existing scheduled actions before it.

What should happen is that soon after the action is added, Action Scheduler should initiate a queue to process it. This can be done via a new async request runner.

### Async Runner

The new runner should be triggered via something other than WP Cron, like using the `'shutdown'` hook on admin requests. Researching how Jetpack and other plugins implement async processing outside of WP Cron within a WordPress site will likely save us time running into gotchas here.

We may also wish to include async request chaining in this runner, as discussed in #131. By that I mean initiate a new request to start processing a new batch of async actions once the batch in this request is completed if there are still other async actions to process.

This differs substantially to the process used by the default WP Cron runner, which is to  simply process as many batches as possible within a given time limit during this request, then wait for another WP Cron request before starting another batch.

Other than that, the runner should be able to look very similar to the WP Cron runner, and use many of its existing methods (i.e. `ActionScheduler_AsyncQueueRunner` can probably extend `ActionScheduler_QueueRunner`).

### API Function

To support adding async actions for the new runner, we should also add a new function named something like `as_add_async_action()`. Unlike other API methods, this function will not require a `$timestamp`. Like other functions, it will require a `$hook` and optionally accept `$args` and `$group` for the async action.

### Implementing an Async Flag

In order for the async runner to be able to process only async actions, there needs to be some flag set on these actions that can be used to identify them. This will make it possible for runners to either only process only async actions, required in the case of the Async Runner, or to at least prioritise processing async actions, which may be done via WP CLI runner.

[...]

PR #322, opened 2019-07-03
Add as_enqueue_async_action() and async action APIs

thenbrent:
The PR introduces a new set of APIs for scheduling actions to run "as soon as possible". This is to achieve the async action type required as part of #319.

### Background

Typically, an action is scheduled for a date. In cases where that action needed to run "as soon as possible", that date would have been set to the current time, e.g. `time()`. For an example, see [`WC_Action_Queue::add()`](https://github.com/woocommerce/woocommerce/blob/3.6.5/includes/queue/class-wc-action-queue.php#L22-L32). The problem with "now" is that there is no way to make sure that action is run sooner than any other action scheduled to run now, including those that may have been scheduled to run at that time months in advance.

In practice, an async action should take priority over a scheduled action, because there is an expectation it will be run very soon after being added to the queue. Whereas a scheduled action, like a subscription payment, may have been scheduled months in advance so if it processes a few minutes late, most of the time there's no harm done as no one is really sitting around expecting it to go through.

### Notes on Approach

This patch makes it acceptable to save an action with a `NULL` schedule. This mapps its MySQL datetime string to `0000-00-00 00:00:00`.

The advantage of this approach is that it is fully backward compatible. Existing queries to claim actions claim them by date, meaning actions with the `0000-00-00 00:00:00` scheduled time will always be claimed prior to actions scheduled for a specific date. This makes sure that any async action is given priority in queue processing. This has the added advantage of making sure async actions can be claimed by both the existing WP Cron and WP CLI runners, as well as a new async request runner.

As this makes async APIs backward compatible with both existing datastores and the new custom table data stores coming in #259, and it has none of the issues mentioned in #319 that would stem from using other backward comptaible APIs, like the using the action group, I've chosen this approach.

The downside is that `0000-00-00 00:00:00` is a bit of a hack and will allow for potentially critical bugs in 3rd party code, as seen by the fact that I had to remove existing validation against `NULL` dates for it to be possible. 3rd parties can now accidentally schedule actions with a `NULL` date, instead of receiving an exception to alert them to errors.

### DIFF
--- a/classes/ActionScheduler_ActionFactory.php
+++ b/classes/ActionScheduler_ActionFactory.php
@@ -45,6 +45,28 @@ class ActionScheduler_ActionFactory {
 	}
 
 	/**
+	 * Enqueue an action to run one time, as soon as possible (rather a specific scheduled time).
+	 *
+	 * This method creates a new action with the NULLSchedule. This schedule maps to a MySQL datetime string of
+	 * 0000-00-00 00:00:00. This is done to create a psuedo "async action" type that is fully backward compatible.
+	 * Existing queries to claim actions claim by date, meaning actions scheduled for 0000-00-00 00:00:00 will
+	 * always be claimed prior to actions scheduled for a specific date. This makes sure that any async action is
+	 * given priority in queue processing. This has the added advantage of making sure async actions can be
+	 * claimed by both the existing WP Cron and WP CLI runners, as well as a new async request runner.
+	 *
+	 * @param string $hook The hook to trigger when this action runs
+	 * @param array $args Args to pass when the hook is triggered
+	 * @param string $group A group to put the action in
+	 *
+	 * @return string The ID of the stored action
+	 */
+	public function async( $hook, $args = array(), $group = '' ) {
+		$schedule = new ActionScheduler_NullSchedule();
+		$action = new ActionScheduler_Action( $hook, $args, $schedule, $group );
+		return $this->store( $action );
+	}
+
+	/**
 	 * @param string $hook The hook to trigger when this action runs
 	 * @param array $args Args to pass when the hook is triggered
 	 * @param int $when Unix timestamp when the action will run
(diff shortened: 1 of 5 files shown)


thenbrent:
> but `as_next_scheduled_action()` doesn't find the async actions.

Ah good catch @rrennick! What do you think is the most sane return value for `as_next_scheduled_action()` for an async action?

I see the following options:

1. `0` to represent the absence of a schedule and/or that its scheduled for "the beginning of time" (which is the Unix Epoch in PHP's world). This is risky though, because it will be confused with `false` via weak comparison operators, so it's likely to create bugs with existing 3rd party code.
1. `1`, to represent a truthy value in `int` that is still near enough to the beginning of time to communicate the above messages.
1. `time()` to indicate that the next time any async action should be run is "now" (which is always true for any pending async action because they should be run ASAP). This _could_ cause issues though, if 3rd party code is checking against something like `as_next_scheduled_action() < time()`
1. something else?

It seems like 2 is the best worst option to me.

thenbrent:
Another couple of options:

5. return `NULL`, to indicate there is a scheduled action (i.e. the return value is not `false`), but it doesn't have a scheduled timestamp. This return value has the same issue as `0`.
6. return boolean `true`, to indicate there is a scheduled action, but it doesn't have a scheduled timestamp. This seems semantically correct, backward compatible and relatively robust. It will fail against any checks for `as_next_scheduled_action() > time()` though. I'm not sure if there is anyway around that though.

thenbrent:
> but `as_next_scheduled_action()` doesn't find the async actions.

I've pushed up SHA: 77d2c09 to fix this. It returns `true` for pending async actions. It continues to return `false` if there are only cancelled, completed, failed or in-progress async actions with a matching hook.

thenbrent:
> Can we change the target branch to version_3_0_0 ?

Yes of course! I've cherry-picked the commits from this branch to a new branch based on `version_3_0_0`, force pushed and changed the base branch to `version_3_0_0`.

I originally wrote this using `add/custom-tables` as the base (on SHA: 5af3d174e93199a69), so it should work fine once those branch in merged too (depending on changes from that SHA to merge).

Issue #131, opened 2018-04-16
Add async request chaining to default runner (or a new runner)


thenbrent:
The default Action Scheduler queue runner currently works by scheduling a hook with _WP-Cron_, then processing a batch of X actions when that hook is triggered. This works well for small queues, but not large queues because the lag between WP Cron requests can slow down the batch processing.

This is especially problematic when:

1. running on a host with a short timeout limit, e.g. #108, meaning only small batches can be run at a time; or
1. another plugin or plugins on the site are also using WP Cron which is clogging WP Cron's queue, meaning for example, even if WP Cron is initiated every minute, Action Scheduler may only have its hook called once every 5 minutes

### WP CLI

To workaround these issues, we have a [WP CLI queue runner](https://github.com/Prospress/action-scheduler#wp-cli) which allows the use of WP CLI to avoid any lag between processing batches. This works really well, but also requires WP CLI (and ideally, WP CLI to be scheduled).

### Async Request Chaining

A good workaround that can work on many hosts is to use [async requests to process jobs](https://github.com/techcrunch/wp-async-task).

The basic idea here is to make a request back to the server/site to initiate processing on a new batch once an existing batch has processed or reaches a certain time limit (and been terminated).

### State of Async Request Chaining with WordPress

The trouble with this, and the reason Action Scheduler doesn't do it yet, is because it's difficult to achieve the type of reliability we require to process payments across varied infrastructure.

There are many [background](https://github.com/A5hleyRich/wp-background-processing) [processing](https://github.com/skyverge/wc-plugin-framework/blob/5.1.2/woocommerce/utilities/class-sv-wp-background-job-handler.php) libraries, including one in [WooCommerce core](https://github.com/woocommerce/woocommerce/blob/3.3.5/includes/abstracts/class-wc-background-process.php).

AFAIK, they all have [open issues](https://github.com/skyverge/wc-plugin-framework/issues/195), and despite often those issues being shared between them, the fixes that are developed are not being pushed back to other libraries (or in some cases where they are submitted, they are ignored by the maintainer/s), meaning there is no canonical, flawless way to run background jobs yet.

### Action Scheduler and Async Request Chaining

The above issues don't mean we can't introduce async requests as part of Action Scheduler, but it does mean we can't rely on them (and perhaps even, we shouldn't default to use them).

[...]

PR #323, opened 2019-07-03
Add async queue processing

thenbrent:
This is part two of #319 - the async queue processing. Part one - async action APIs - can be found in #322.

This PR extends the default `ActionScheduler_QueueRunner` class to include async processing. Queues are now run in asynchronous requests, as well as WP Cron requests.

These requests are initiated via the 'shutdown' hook on admin requests when Action Scheduler is not already processing the maximum number of allowed queues. The `'shutdown'` hook is the same approach taken by Jetpack and [Async Task](https://github.com/techcrunch/wp-async-task).

This removes Action Scheduler's dependence on WP Cron. While `ActionScheduler_QueueRunner` still uses WP Cron to process queues, queues will also now be processed anytime an admin request is made and there are pending actions.

Beyond removing the dependency on WP Cron, this has another major advantage - speed.

The approach taken here includes async request chaining (as initially discussed in #131). Whenever one async request processes a batch of actions to the allowed limits (the 20 second time limit is normally the limit that would be hit), a new async request to start processing a new batch of actions will be initiated if there are still other async actions to process.

This reduces the lag between processing each batch, and combined with queue concurrency, it substantially increases throughput when more than one queue is being processed by async requests (especially given the fact that each async request is also an admin request, so it will also initiate a 2nd async request on `'shutdown'` in addition to the async request created to chain processing for the original request).

Importantly, this differs substantially to the process currently used with WP Cron, which is to simply process as many actions/batches as possible within a given time limit during that single request, then wait for another WP Cron request before starting another batch.

#### Keeping WP Cron

I've kept the WP Cron processing, because it's a helpful fallback for when loopback requests are not supported by the site/server.

#### Running Queues via the `'action_scheduler_run_queue'` Action

Importantly, queues are run in the async request by triggering the ` 'action_scheduler_run_queue'` hook, rather than calling `ActionScheduler::runner()->run()` directly.

This maintains compatibility with the WP Cron runner, and helps ensure existing code will apply to this new method for processing queues, like this plugin: https://github.com/Prospress/action-scheduler-disable-default-runner

#### Reduced Default Queue Concurrency

Importantly, given the increased likelihood of concurrent queues now occurring on all sites (instead of just sites tweaking things with Action Scheduler), there is a need to be more conservative with the default concurrency allowed by Action Scheduler.

SHA: f493e7d9914b5ad7e66d5 takes care of this, by reducing the default concurrency to two queues, from five. This prevents consuming too many resources by default. The concurrency can still be [increased](https://actionscheduler.org/perf/#increasing-concurrent-batches) when possible, if needed. I expect the chained async requests will increase throughput enough that concurrency increases will be required less now.

#### Disabling Async Queues with `'action_scheduler_allow_async_runner'`

As mentioned [here](https://github.com/Prospress/action-scheduler/issues/131#issuecomment-391227809), chained async requests without a kill-switch can be a dangerous thing.

To mitigate this, I've included a new `'action_scheduler_allow_async_runner'` filter. This can be used to both:

1. prevent any actions being run via async requests
1. terminate an async request chain part way through

- [ ] this new filter will need to be added to [the docs](https://actionscheduler.org/).

#### Initialising the `ActionScheduler_AsyncRequest_QueueRunner`

I'm injecting the `ActionScheduler_AsyncRequest_QueueRunner` class into the `ActionScheduler_QueueRunner`. That's to match the approach taken for other dependencies by `ActionScheduler_QueueRunner`; however, I don't see much need to swap it.

#### `WP_Async_Request`

I've used the `WP_Async_Request` library used by [WP Background Processing](https://github.com/deliciousbrains/wp-background-processing) for handling the request. This wasn't really necessary, as its mostly just a wrapper around `wp_remote_post()`, but I figured using it was better than starting from scratch.

## Testing

A fun way to test this is with the following snippet and visiting a page on your site with the URL param `?as_add_async_actions=true`. It will require the patch in #322 to get the `as_enqueue_async_action()` method.

```php
function as_test_schedule_async_action() {

	if ( ! isset( $_GET['as_add_async_actions'] ) ) {
		return;
	}

	for( $i = 1; $i < 300; $i++ ) {
		as_enqueue_async_action( 'test_async_action', array( 'timestamp' => time(), 'i' => $i ), 'async' );
	}
}
add_action( 'shutdown', 'as_test_schedule_async_action' );

function as_test_schedule_async_action_sleep() {
	sleep( 1 );
}
add_action( 'test_async_action', 'as_test_schedule_async_action_sleep' );
```

This will add hundreds of actions, and process them at a sustained rate of 120 / minute, as each action takes 1 second to process, and two concurrent queues will be processing them. With WP Cron only, the maximum throughput would be 20 / minute.

### DIFF
--- a/classes/ActionScheduler_AsyncRequest_QueueRunner.php
+++ b/classes/ActionScheduler_AsyncRequest_QueueRunner.php
@@ -58,18 +58,11 @@ class ActionScheduler_AsyncRequest_QueueRunner extends WP_Async_Request {
 	 * dispatch an async request to process them.
 	 */
 	public function maybe_dispatch() {
-		if ( ! $this->allow() ) {
+		if ( ! $this->allow() || ! $this->store->has_pending_actions_due() ) {
 			return;
 		}
 
-		$pending_actions = $this->store->query_actions( array(
-			'date'   => as_get_datetime_object(),
-			'status' => ActionScheduler_Store::STATUS_PENDING,
-		) );
-
-		if ( $pending_actions ) {
-			$this->dispatch();
-		}
+		$this->dispatch();
 	}
 
 	/**
--- a/classes/ActionScheduler_AsyncRequest_QueueRunner.php
+++ b/classes/ActionScheduler_AsyncRequest_QueueRunner.php
@@ -54,9 +54,14 @@ class ActionScheduler_AsyncRequest_QueueRunner extends WP_Async_Request {
 	}
 
 	/**
-	 * If there are pending actions, dispatch an async request to process them.
+	 * If the async request runner is allowed, and there are pending actions,
+	 * dispatch an async request to process them.
 	 */
 	public function maybe_dispatch() {
+		if ( ! $this->allow() ) {
+			return;
+		}
+
 		$pending_actions = $this->store->query_actions( array(
 			'date'   => as_get_datetime_object(),
 			'status' => ActionScheduler_Store::STATUS_PENDING,
@@ -65,5 +70,12 @@ class ActionScheduler_AsyncRequest_QueueRunner extends WP_Async_Request {
 		if ( $pending_actions ) {
 			$this->dispatch();
 		}
+	}
+
+	/**
+	 * Allow 3rd party code to disable running actions via async requets.
+	 */
+	protected function allow() {
+		return apply_filters( 'action_scheduler_allow_async_request_runner', true );
 	}
 }
--- a/classes/ActionScheduler_AsyncRequest_QueueRunner.php
+++ b/classes/ActionScheduler_AsyncRequest_QueueRunner.php
@@ -1,0 +1,69 @@
+<?php
+/**
+ * ActionScheduler_AsyncRequest_QueueRunner
+ */
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * ActionScheduler_AsyncRequest_QueueRunner class.
+ */
+class ActionScheduler_AsyncRequest_QueueRunner extends WP_Async_Request {
+
+	/**
+	 * Data store for querying actions
+	 *
+	 * @var ActionScheduler_Store
+	 * @access protected
+	 */
+	protected $store;
+
+	/**
+	 * Prefix for ajax hooks
+	 *
+	 * @var string
+	 * @access protected
+	 */
+	protected $prefix = 'as';
+
+	/**
+	 * Action for ajax hooks
+	 *
+	 * @var string
+	 * @access protected
+	 */
+	protected $action = 'async_request_queue_runner';
+
+	/**
+	 * Initiate new async request
+	 */
+	public function __construct( ActionScheduler_Store $store ) {
+		parent::__construct();
+		$this->store = $store;
+	}
+
+	/**
+	 * Handle async requests
+	 *
+	 * Run a queue, and maybe dispatch another async request to run another queue
+	 * if there are still pending actions after completing a queue in this request.
+	 */
+	protected function handle() {
+		do_action( 'action_scheduler_run_queue' ); // run a queue in the exact same way as WP Cron
+		$this->maybe_dispatch();
+	}
+
+	/**
+	 * If there are pending actions, dispatch an async request to process them.
+	 */
+	public function maybe_dispatch() {
+		$pending_actions = $this->store->query_actions( array(
+			'date'   => as_get_datetime_object(),
+			'status' => ActionScheduler_Store::STATUS_PENDING,
+		) );
+
+		if ( $pending_actions ) {
+			$this->dispatch();
+		}
+	}
+}
--- a/classes/ActionScheduler_OptionLock.php
+++ b/classes/ActionScheduler_OptionLock.php
@@ -1,0 +1,49 @@
+<?php
+
+/**
+ * Provide a way to set simple transient locks to block behaviour
+ * for up-to a given duration.
+ *
+ * Class ActionScheduler_OptionLock
+ * @since 3.0.0
+ */
+class ActionScheduler_OptionLock extends ActionScheduler_Lock {
+
+	/**
+	 * Set a lock using options for a given amount of time (60 seconds by default).
+	 *
+	 * Using an autoloaded option avoids running database queries or other resource intensive tasks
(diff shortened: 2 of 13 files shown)


rrennick:
This PR needs #322 to be merged before being tested.

rrennick:
After merging in #322, my async actions aren't running on dashboard loads. I'm seeing one of these warnings for each dashboard load:

```
[05-Jul-2019 13:37:55 UTC] PHP Notice:  Undefined property: ActionScheduler_AsyncRequest_QueueRunner::$data in wp-content/plugins/action-scheduler/lib/WP_Async_Request.php on line 107
```

thenbrent:
>  I'm seeing one of these warnings for each dashboard load

Good catch. That was a potentially nasty bug. I had modified the `WP_Async_Request` class while checking how much of it was actually needed (and considering creating a more lightweight fork) and mistakenly committed those modifications under `WP_Async_Request`. That would lead to compatibility issues with any 3rd party using that class, like WC core and many others. I've updated the `WP_Async_Request` to be an exact copy of https://github.com/deliciousbrains/wp-background-processing/blob/fbbc56f2480910d7959972ec9ec0819a13c6150a/classes/wp-async-request.php

I've also changed the base branch of this PR to `version_3_0_0` and cherry-picked the commits to be on top of that branch, instead of `master` for a clean merge.

thenbrent:
I've also just added

* SHA: 2eb84ff54a049 to improve the location of the `'action_scheduler_allow_async_request_runner'` filter; and
* SHA: ee788f5df62 to avoid introducing an issue like https://github.com/woocommerce/woocommerce-admin/issues/2591

rrennick:
@thenbrent This worked great for me in testing. The one thing I noted was that if you load the Scheduled Actions admin screen it includes the pending actions that will be processed when the page load completes. Could we add a notice to the top of that screen indicating that some actions may be run asynchronously?

thenbrent:
Good idea @rrennick. I've added that with SHA: ec1ec246b1072. The locking code required quite a bit of refactoring to be able to provide effective notices.

Here is an example of the notice now displayed when there are pending actions due to be processed and:

* the lock is not set, so a new async runner will be triggered on `'shutdown'`: https://cld.wthms.co/uKKCgS
* the async queue runner lock is set: https://cld.wthms.co/olUH8t

This system isn't perfect, because the `in-progress` view will only show actions with that status at the time of the page load (and it's possible AS will be between actions during that time), but it's still a nice improvement.

classes/ActionScheduler_OptionLock.php
```diff
@@ -0,0 +1,49 @@
...
+	 * For example, ActionScheduler_QueueRunner->maybe_dispatch_async_request() uses a lock to avoid
+	 * calling ActionScheduler_QueueRunner->has_maximum_concurrent_batches() every time the 'shutdown',
+	 * hook is triggered, because that method calls ActionScheduler_QueueRunner->store->get_claim_count()
+	 * to find the current number of claims in the database.
+	 *
+	 * @param string $lock_type A string to identify different lock types.
+	 * @bool True if lock value has changed, false if not or if set failed.
+	 */
+	public function set( $lock_type ) {
+		return update_option( $this->get_key( $lock_type ), time() + $this->get_duration( $lock_type ), $this->get_duration( $lock_type ) );
```
rrennick:
```suggestion
		return update_option( $this->get_key( $lock_type ), time() + $this->get_duration( $lock_type ) );
```

The third parameter of `update_option` is the `autoload` value.

classes/abstracts/ActionScheduler_Lock.php
```diff
@@ -0,0 +1,60 @@
...
+		return ( $this->get_expiration( $lock_type ) >= time() );
+	}
+
+	/**
+	 * Set a lock.
+	 *
+	 * @param string $lock_type A string to identify different lock types.
+	 * @return bool
+	 */
+	public function set( $lock_type ) {}
```
rrennick: Should this be abstract?

classes/abstracts/ActionScheduler_Lock.php
```diff
@@ -0,0 +1,60 @@
...
+
+	/**
+	 * If a lock is set, return the timestamp it was set to expiry.
+	 *
+	 * @param string $lock_type A string to identify different lock types.
+	 * @return bool|int False if no lock is set, otherwise the timestamp for when the lock is set to expire.
+	 */
+	public function get_expiration( $lock_type ) {}
```
rrennick: Also abstract?

rrennick:
@thenbrent Other than the 3 small things I noted, this works great. I checked running AS Admin and the REST API calls will process the pending actions in the background. I may want to add similar notices to it.

thenbrent:
@rrennick good catch on all 3 counts! I've rebased in fixes. :)

classes/abstracts/ActionScheduler_Lock.php
```diff
@@ -0,0 +1,60 @@
+<?php
+
+/**
+ * Class
```
rrennick:
```suggestion
 * Class
```

Remove trailing space. Should we look at getting a code sniffer set up for this repo?

classes/ActionScheduler_AsyncRequest_QueueRunner.php
```diff
@@ -0,0 +1,74 @@
+<?php
+/**
+ * ActionScheduler_AsyncRequest_QueueRunner
+ */
+
+defined( 'ABSPATH' ) || exit;
```
JPry: There's nothing wrong with including this, but I just wanted to note that as long as this file (or any other file) *only* declares a class with no other secondary logic, it's OK to omit this check.

classes/abstracts/ActionScheduler_Lock.php
thenbrent: > Should we look at getting a code sniffer set up for this repo?

That would be great. The question on that is which set of standards should it follow? It definitely doesn't follow WP standards yet.

thenbrent:
> Feel free to merge once the space is removed.

I rebased in a patch to add the class, and also a brief comment on the purpose of the class. I'll merge once Travis passes. 🚀

PR #326, opened 2019-07-11
Fix marking actions with invalid schedules as failed

thenbrent:
This PR reverts Prospress/action-scheduler#297 because it would mark the action as failed every time it was fetched, which included things like every time an admin page loads to display the action. The desired behaviour is to mark the action as failed once, when an attempt to run it is undertaken (and that attempt fails). I will introduce that in a separate PR, because it relies on some other improvements to how exceptions are handled when failing to fetch the action.

Because the code from Prospress/action-scheduler#297 has already been released, I've added a new `ActionScheduler_Store_Deprecated` class to safely deprecate the public and protected methods being reverted.

### DIFF
--- a/deprecated/ActionScheduler_Store_Deprecated.php
+++ b/deprecated/ActionScheduler_Store_Deprecated.php
@@ -1,0 +1,19 @@
+<?php
+
+/**
+ * Class ActionScheduler_Store_Deprecated
+ * @codeCoverageIgnore
+ */
+abstract class ActionScheduler_Store_Deprecated {
+
+	/**
+	 * Get the site's local time.
+	 *
+	 * @deprecated 2.1.0
+	 * @return DateTimeZone
+	 */
+	protected function get_local_timezone() {
+		_deprecated_function( __FUNCTION__, '2.1.0', 'ActionScheduler_TimezoneHelper::set_local_timezone()' );
+		return ActionScheduler_TimezoneHelper::get_local_timezone();
+	}
+}
--- a/deprecated/ActionScheduler_Store_Deprecated.php
+++ b/deprecated/ActionScheduler_Store_Deprecated.php
@@ -7,6 +7,36 @@
 abstract class ActionScheduler_Store_Deprecated {
 
 	/**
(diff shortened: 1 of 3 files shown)


rrennick:
This will create another merge conflict with #259 so, if we can, hold this one until after #259 is merged.

thenbrent:
I've rebased this one on `version_3_0_0` now that #259 has been merged.

deprecated/ActionScheduler_Store_Deprecated.php
```diff
@@ -0,0 +1,49 @@
...
+
+	/**
+	 * Mark an action that failed to fetch correctly as failed.
+	 *
+	 * @since 2.2.6
+	 *
+	 * @param int $action_id The ID of the action.
+	 */
+	public function mark_failed_fetch_action( $action_id ) {
+		_deprecated_function( __METHOD__, '3.0', 'ActionScheduler_Store::mark_failure()' );
```
rrennick:
```suggestion
		_deprecated_function( __METHOD__, '3.0.0', 'ActionScheduler_Store::mark_failure()' );
```

deprecated/ActionScheduler_Store_Deprecated.php
```diff
@@ -0,0 +1,49 @@
...
+		self::$store->mark_failure( $action_id );
+	}
+
+	/**
+	 * Add base hooks
+	 *
+	 * @since 2.2.6
+	 */
+	protected static function hook() {
+		_deprecated_function( __METHOD__, '3.0' );
```
rrennick: 3.0.0

deprecated/ActionScheduler_Store_Deprecated.php
```diff
@@ -0,0 +1,49 @@
...
+	}
+
+	/**
+	 * Remove base hooks
+	 *
+	 * @since 2.2.6
+	 */
+	protected static function unhook() {
+		_deprecated_function( __METHOD__, '3.0' );
```
rrennick: 3.0.0

thenbrent:
Thanks @rrennick! I've fixed up those issues.

PR #328, opened 2019-07-11
Tweak fetch exception handling

thenbrent:
This PR is an extension of work done in a number of prior PRs to handle corrupted data. Specifically, if an action has a corrupted schedule or set of args in the database, fetching those would cause an exception. The first approach to address this was to throw the exception. While this was fine for the purposes of running the queue, other calls to `ActionScheduler::store()->fetch_action()` were not updated to catch exceptions, leading to errors.

The next iteration attempted to catch the exception and log it. This prevents the exceptions halting code execution, but it also lead to other potential bugs, because a valid action was still being returned by `fetch_action()`, when really, once we know data is corrupted, we should be avoiding giving any indication the action is valid, because we really don't know what's happened, and we definitely don't want the action to be run (i.e. `$action->execute()`) because we might be running it on the wrong schedule, or with invalid args (leading to other fatal errors).

This iteration uses a similar approach to that done for an invalid `$action_id` - return an instance of `ActionScheduler_NullAction`. This helps avoid throwing an exception, because we are returnign a valid. But it also avoids harder to catch diagnose errors due to _not_ throwing an exception when the data is corrupted, becuase `ActionScheduler_NullAction::execute()` does nothing, instead of calling `do_action_ref_array()`.

The one thing not being done well with this patch is that on #326 is merged, the action will simply be marked as `complete` after being claimed & processed and then remain as `complete`. It should be marked as `failed`. That can be done by having `ActionScheduler_Abstract_QueueRunner::process_action()` throw an exception if `$this->store->fetch_action( $action_id )` returns a `ActionScheduler_NullAction`.

### DIFF
--- a/classes/ActionScheduler_InvalidActionException.php
+++ b/classes/ActionScheduler_InvalidActionException.php
@@ -10,6 +10,22 @@
 class ActionScheduler_InvalidActionException extends \InvalidArgumentException implements ActionScheduler_Exception {
 
 	/**
+	 * Create a new exception when the action's schedule cannot be fetched.
+	 *
+	 * @param string $action_id The action ID with bad args.
+	 * @return static
+	 */
+	public static function from_schedule( $action_id, $schedule ) {
+		$message = sprintf(
+			__( 'Action [%s] has an invalid schedule: %s', 'action-scheduler' ),
+			$action_id,
+			var_export( $schedule, true )
+		);
+
+		return new static( $message );
+	}
+
+	/**
 	 * Create a new exception when the action's args cannot be decoded to an array.
 	 *
 	 * @author Jeremy Pry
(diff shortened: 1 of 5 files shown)


rrennick:
> The one thing not being done well with this patch is that on #326 is merged, the action will simply be marked as complete after being claimed & processed and then remain as complete. It should be marked as failed. 

The NullAction `execute()` could throw an exception. Then it would be handled by the existing handler for an action hook throwing an exception.

thenbrent:
> The NullAction `execute()` could throw an exception. Then it would be handled by the existing handler for an action hook throwing an exception.

That's a really good idea. 👍 

This PR should probably be put on hold until after #333 is merged now given it has the more critical, underlying fixes to schedules.

thenbrent:
I've rebased this onto `version_3_0_0` and added SHA: 840608165d, to apply similar exceptions and handling for corrupted args or schedules stored in the `DBStore`.

PR #329, opened 2019-07-12
Add context to action processing

### LINKED ISSUE #129: Include which Queue Runner is responsible for claiming/running the action in logs
Now that we have more than one Queue Runner in core (WP-Cron and WP CLI), It will be helpful when tracing an action's history, and generally diagnosing issues on a site, to be able to see which Queue Runner was responsible for running the action.

Hopefully this will be as simple as updating `'action_scheduler_before_execute'` to pass `$this` or some other identifier to callbacks (other hooks in `ActionScheduler_Abstract_QueueRunner::process_action()` may also be updated, but I don't think it will be as necessary, for example, to log which queue runner triggered `'action_scheduler_after_execute'` or `'action_scheduler_failed_execution'`, because that should be theoretically the same as the one on `'action_scheduler_before_execute'`, but it might make sense to pass it to them for consistency).

For background, Prospress team can see: https://secure.helpscout.net/conversation/544426609/6675?folderId=1524429

thenbrent:
Add context to action processing / queue runners so that we can log where the action are run from, e.g. WP CLI, WP Cron the Admin List Table or the upcoming Async Runner.

Very handy for debugging issues after the fact.

Example logs entries: https://cld.wthms.co/fT0T96

Fixes #129

### DIFF
--- a/classes/ActionScheduler_AsyncRequest_QueueRunner.php
+++ b/classes/ActionScheduler_AsyncRequest_QueueRunner.php
@@ -49,7 +49,7 @@ class ActionScheduler_AsyncRequest_QueueRunner extends WP_Async_Request {
 	 * if there are still pending actions after completing a queue in this request.
 	 */
 	protected function handle() {
-		do_action( 'action_scheduler_run_queue' ); // run a queue in the exact same way as WP Cron
+		do_action( 'action_scheduler_run_queue', 'Async Request' ); // run a queue in the same way as WP Cron, but declare the Async Request context
 		$this->maybe_dispatch();
 	}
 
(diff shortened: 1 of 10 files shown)


thenbrent:
I looked at making the context available for translation, but it's a proper noun in most cases - WP CLI, WP Cron - and pretty close to a proper noun in others - Admin List Table and Async Request - so I decided to instead leave it out of translation.

Are there any benefits to using slugs elsewhere? e.g. `wp_cron` instead of `'WP Cron'` (other than avoiding strings in places you might not expect whitespace).

> The context array could be stored so only the slug needed to be passed.

We'd also need to make sure it's got a filter.

rrennick:
> I looked at making the context available for translation, but it's a proper noun in most cases

That sounds good.

Some of the unit tests need to be updated for message string changes.

thenbrent:
> Some of the unit tests need to be updated for message string changes.

@rrennick fixed. 👍

thenbrent:
Thanks @rrennick. I've rebased in a fix for the merge conflict.

Issue #291, opened 2019-04-18
Fatal error on Uncaught Exception


rrennick:
See https://github.com/woocommerce/woocommerce/issues/23373

rrennick:
Fixed in #292

Issue #331, opened 2019-07-17
Add code sniffing


rrennick:
Follow up from https://github.com/Prospress/action-scheduler/pull/323#discussion_r302058476

> The question on that is which set of standards should it follow? It definitely doesn't follow WP standards yet.

[WP coding standards for newer versions of PHP](https://make.wordpress.org/core/2019/07/12/php-coding-standards-changes/)

The WP standards all seemed fine for me except prohibiting the use of short array syntax. With appropriate spacing, I find the short syntax easier to read (partly because it is formatted similar to JSON). 

Since that is the first pass at standards for the newer language features, it may change in the future based on community reaction so we shouldn't feel bound to that.

claudiosanches:
I started a Pull Request using the dev tools that we use on WooCommerce, I hope this may help you all #383

barryhughes:
With the work done in [PR #383](https://github.com/woocommerce/action-scheduler/pull/383) seems like this can be closed (and we can follow-up in further issues if we want to take it further, such as integrating with GH Actions or similar)—feel free to re-open if I'm mistaken!

PR #333, opened 2019-07-19
Fix recurring schedules, e.g. Cron and Interval schedules

### LINKED ISSUE #330: Bug in recurring schedule
This is a good one. This description is mostly thinking out loud to help document my understanding of the issue and record analysis of possible fixes.

There is a (very old) bug in how recurring schedules are handled. Specifically, when a recurring action is run and the next instance is saved by [`QueueRunner::schedule_next_instance()`](https://github.com/Prospress/action-scheduler/blob/2.2.5/classes/ActionScheduler_Abstract_QueueRunner.php#L77-L84), it's scheduled date is updated, but the dates on the schedule object itself are not, i.e. the instance of either `ActionScheduler_IntervalSchedule` or `ActionScheduler_CronSchedule` has the `$start` property equal to the original action's scheduled date. 

Because of that, any calls to `$action->get_schedule()->next()` on the `2...n` instance of a recurring action will return the first instance's scheduled date, not that instance's scheduled date (which is set correctly when it's run, because `QueueRunner::schedule_next_instance()` calculates it using `$schedule->next( as_get_datetime_object() )` and then that is saved on the action via `Store::save_action()`).

This has lead to various issues, including:

* an admin display bug #256 / #262
* unhandled exceptions related to an invalid schedule #291 / #292, #293, #296 and possibly #258

There are root issue is still unresolved though.

### The Root Issue

There is confusion within the code about whether the `$start` property 
of a recurring schedule class, `ActionScheduler_IntervalSchedule` or `ActionScheduler_CronSchedule`, is the _next_ scheduled date, or the _first_ scheduled date. 

A great demonstration of this is seen in the code here to disambiguate that: https://github.com/Prospress/action-scheduler/pull/262/files

This manifests in a few ways, most notable is `as_next_scheduled_action()` using `$action->get_schedule()->next()` to find the "next scheduled date" when really, that just returns the "first scheduled date" for a recurring action.

Another manifestation is the `next()` method itself. This method appears to serve two purposes:

1. return the stored "next run" date for the schedule, if no parameters are passed to it (this maps to the `$start` property, so it would make sense that that should represent the next date based on that internal usage, not the first scheduled date in the sequence)
2. calculate the "next run" date after a given time, if an `$after` parameter is passed to it

## Proposed Fix

One fix is very similar to code already proposed in #262 - add a new method to set the next date for a recurring schedule. This makes it possible to update the schedule object on an existing subscription.

#262 also disambiguates the `$start` property by adding a new `$first` property to store when an action should start running (i.e. the "next date") with when the schedule begins. This makes it possible to store both pieces of data.

#262 would require a few changes to fully fix this bug (rather than just the admin display bug it was designed to fix), namely:

* the `next()` method in #262 should be updated to return `$this->start` instead of `$this->first` when no parameter is passed to insure returns the expected value by `as_next_scheduled_action()`.
* inline with the above, the new `get_start()` method should probably be replaced by a new `get_first()` method
* the `ActionScheduler_CronSchedule` may also need a new set of matching public methods like `set_next()`
* the `__construct()` method would need to be updated to accept both the `$first` and `$start` / `$next` params. Likely, the first parameter would be mapped to `$start` / `$next` still, and the new 3rd parameter would be mapped to `$first`.

This patch with the above modifications seems like relatively simple and suitable as a fix. It's not fully backward compatible as it's adding new public methods to parent classes, but it should be sufficiently safe and it would go out in 3.0.0 anyway.

### Alternative approach to this fix

At the moment, `QueueRunner::schedule_next_instance()` calculates the next scheduled date, and passes that to `Store::save_action()`. This is what conflates the issue with `Schedule::next()` returning the `$start` date. The schedule's `$start` date won't be the next scheduled date because it's not updated.

#262 uses the new `set_next()` to update that. An alternative approach would be to create a new action with a new schedule here with the correct `$start` date param value set to `$next`. `ActionScheduler::factory()->recurring()` and `ActionScheduler::factory()->cron()` can also be used to achieve this quite easily, rather calling `Store::save_action()` directly.

All existing APIs would continue to work with this approach. The only downside is that it would be impossible for a recurring schedule to know when it was first scheduled (vs. when it is next supposed to start). As far as I can tell, that information isn't required or used anywhere in Action Scheduler, but it does seem handy to keep around.

This could be mitigated by adding a 3rd parameter to recurring schedules for defining the `$first` date, and have that default to the `$start` / `$next` date when not set.

## Alternative Fix: `Store::get_scheduled_date_gmt()` method

We already have an API for getting a pending action's scheduled date - `Store::get_date_gmt()`. It would be possible to simply update uses of `$action->get_schedule()->next()` to get the next scheduled date to use `Store::get_date_gmt()` instead.

The gotcha with this is that `Store::get_date_gmt()` will return the run date or modified date for a non-pending actions. To get the scheduled date for a `cancelled`, `failed` or `complete` action, we'd need a new `Store::get_scheduled_date_gmt()` method.

Adding this method doesn't seem ideal given that we effectively already have APIs to find an action's scheduled date - `Schedule::next()`, it's just not working as expected.

### Test Cases

Any proposed fix will need to fix:

- [ ] **Scheduled Date** column value displayed on the admin list table #256
- [ ] `as_next_scheduled_action()` returns the next scheduled date (not the first scheduled date) when called on `2..n` instances of a recurring action
- [ ] A new unit test to test `Schedule::next()` return values are the next scheduled date (not the first scheduled date) against `2..n` instances of a recurring action

thenbrent:
This is a monster, late stage PR for 3.0.0. However, because there are breaking changes, and it resolves a number of discrepancies and gotchas in the recurring and cron actions & schedules, I think it's worth including in 3.0.0 instead of waiting for a 4.0.0, or shipping in a 3.x.x version.

---

Fixes #330 and associated issues - #256, #258, #291, #293 and #296.

---

As outlined in #330, there are a number of issues with recurring action that use either a Cron or Interval schedule.

This PR fixes recurring schedules, and in the process, refactors all schedules to avoid duplicate code and ambiguity of properties and parameters.

This patch introduces two important new base classes:

 * `ActionScheduler_Abstract_Schedule` and
 * `ActionScheduler_Abstract_RecurringSchedule`

These are then used to implement all other schedule types - cron, recurring/interval and simple. This helps unify the data on each schedule. For example, schedules used to have different property names referring to equivalent data like `ActionScheduler_IntervalSchedule::start_timestamp` and `ActionScheduler_SimpleSchedule::timestamp`.

Both of these timestamps actually referring to the date the action was scheduled, but due to ambiguity in naming, like "_start_" or "_date_" or "_first_", the data was sometimes used to refer to the scheduled date, and sometimes the first instance of a recurring schedule. While unifying the properties, their nomenclature has also been changed to avoid ambiguity.

While working on this, I also took the opportunity to:

* harden safeguards when next instances are scheduled to make only recurring action are ever rescheduled
* introduce a new `ActionScheduler_CanceledSchedule` class which takes advantage of the separation of scheduled / next dates on a schedule so that a cancelled action's scheduled date can be derived (and displayed in the admin)
* improve docblocks to better record any gotchas, like time slippage when rescheduling an instance of a recurring action
* add more unit tests, particularly to test cases of 2...n instances of a recurring action
* log issues rescheduling a recurring action

### DIFF
--- a/classes/ActionScheduler_ActionFactory.php
+++ b/classes/ActionScheduler_ActionFactory.php
@@ -22,6 +22,9 @@ class ActionScheduler_ActionFactory {
 				break;
 			case ActionScheduler_Store::STATUS_CANCELED :
 				$action_class = 'ActionScheduler_CanceledAction';
+				if ( ! is_null( $schedule ) && ! is_a( $schedule, 'ActionScheduler_CanceledSchedule' ) ) {
+					$schedule = new ActionScheduler_CanceledSchedule( $schedule->get_date() );
+				}
 				break;
 			default :
 				$action_class = 'ActionScheduler_FinishedAction';
--- a/classes/ActionScheduler_ActionFactory.php
+++ b/classes/ActionScheduler_ActionFactory.php
@@ -107,17 +107,19 @@ class ActionScheduler_ActionFactory {
 	 *
 	 * @param string $hook The hook to trigger when this action runs
 	 * @param array $args Args to pass when the hook is triggered
-	 * @param int $first Unix timestamp for the first run
+	 * @param int $base_timestamp The first instance of the action will be scheduled
+	 *        to run at a time calculated after this timestamp matching the cron
+	 *        expression. This can be used to delay the first instance of the action.
 	 * @param int $schedule A cron definition string
 	 * @param string $group A group to put the action in
 	 *
 	 * @return string The ID of the stored action
 	 */
-	public function cron( $hook, $args = array(), $first = null, $schedule = null, $group = '' ) {
+	public function cron( $hook, $args = array(), $base_timestamp = null, $schedule = null, $group = '' ) {
 		if ( empty($schedule) ) {
-			return $this->single( $hook, $args, $first, $group );
+			return $this->single( $hook, $args, $base_timestamp, $group );
 		}
-		$date = as_get_datetime_object( $first );
+		$date = as_get_datetime_object( $base_timestamp );
 		$cron = CronExpression::factory( $schedule );
 		$schedule = new ActionScheduler_CronSchedule( $date, $cron );
 		$action = new ActionScheduler_Action( $hook, $args, $schedule, $group );
--- a/classes/ActionScheduler_ActionFactory.php
+++ b/classes/ActionScheduler_ActionFactory.php
@@ -82,6 +82,8 @@ class ActionScheduler_ActionFactory {
 	}
 
 	/**
+	 * Create the first instance of an action recurring on a given interval.
+	 *
 	 * @param string $hook The hook to trigger when this action runs
 	 * @param array $args Args to pass when the hook is triggered
 	 * @param int $first Unix timestamp for the first run
@@ -100,8 +102,9 @@ class ActionScheduler_ActionFactory {
 		return $this->store( $action );
 	}
 
-
 	/**
+	 * Create the first instance of an action recurring on a Cron schedule.
+	 *
 	 * @param string $hook The hook to trigger when this action runs
 	 * @param array $args Args to pass when the hook is triggered
 	 * @param int $first Unix timestamp for the first run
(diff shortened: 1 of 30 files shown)


thenbrent:
I've rebased this on `version_3_0_0` now that #259 has been merged. One of the new benefits of this changeset I found while merging in that code is that it's possible to migrate _cancelled_ actions now, because we can access their scheduled date via the new `ActionScheduler_CanceledSchedule`. Previously, they were simply deleted from the old data store instead of being migrated.

rrennick:
@thenbrent This is looking good. Just a couple things:

Related to #331, we should start ensuring any new code merged has consistent spacing. The WP coding standards for spacing provide for good readability. I used the following to find some of the inconsistent spacing (in added code only)

```
$ git show dfb1f11a10fec59c9e6d590d7a2c037ff5eb6a60 | grep ^+ | grep \( | grep -v '( '
+	public function test_next_instance_of_interval_action() {
+		$random    = md5(rand());
+		$date      = as_get_datetime_object('12 hours ago');
+		$store     = ActionScheduler::store();
+		$claim = $store->stake_claim(10, $date);
+		$actions = $claim->get_actions();
+		$this->assertCount(1, $actions);
+		$fetched_action_id = reset($actions);
+		$fetched_action    = $store->fetch_action($fetched_action_id);
+		$store->release_claim($claim);
+		$runner->run();
+		$date = as_get_datetime_object('+1 day');
+		$claim = $store->stake_claim(10, $date);
+		$fetched_action_id = reset($actions);
+		$fetched_action    = $store->fetch_action($fetched_action_id);
+		$store->release_claim($claim);
+		$date = as_get_datetime_object('+1 day');
+		$claim = $store->stake_claim(10, $date);
+		$actions = $claim->get_actions();
+		$this->assertCount(1, $actions);
+		$fetched_action_id = reset($actions);
+		$fetched_action    = $store->fetch_action($fetched_action_id);
```

I'm getting a warning in WP CLI and in the debug log:

```
$ wp option get siteurl
Notice: ActionScheduler_Schedule_Deprecated::next is <strong>deprecated</strong> since version 3.0.0! Use ActionScheduler_Schedule_Deprecated::get_date() instead. in /wp-includes/functions.php on line 4435
https://ron.localhost
```

thenbrent:
> we should start ensuring any new code merged has consistent spacing

@rrennick good call. I've fixed up spacing and rebased in the fixes.

I ran `git show HEAD...a4934fcea90 | grep ^+ | grep \( | grep -v '( '` to review the whole diff.

> I'm getting a warning in WP CLI and in the debug log:

I can't see any remaining calls to `ActionScheduler_Schedule_Deprecated::next()`. Is that notice coming from [Action Scheduler Admin](https://github.com/Prospress/action-scheduler-admin) plugin?

rrennick:
> I can't see any remaining calls to ActionScheduler_Schedule_Deprecated::next(). Is that notice coming from Action Scheduler Admin plugin?

Yes. Other than a rebase this is good to go.

thenbrent:
Rebased, thanks @rrennick!

PR #316, opened 2019-06-13
add help screen to admin list table

rrennick:
Closes #275 

This PR adds a basic help screen to the Tools -> Scheduled Actions screen.

PR #335, opened 2019-07-22
make add_help_tabs static

### LINKED ISSUE #334: non-static method ActionScheduler_AdminHelp::add_help_tabs() should not be called statically
Hi there,

I just wanted to report that I'm getting the following error:

```
call_user_func_array() expects parameter 1 to be a valid callback, non-static method ActionScheduler_AdminHelp::add_help_tabs() should not be called statically
```

Thank you for this library :)

rrennick:
Closes #334 

This is a follow up PR #316 to fix the warning for the function missing the static declaration.

### DIFF
--- a/classes/ActionScheduler_AdminHelp.php
+++ b/classes/ActionScheduler_AdminHelp.php
@@ -7,7 +7,7 @@
 class ActionScheduler_AdminHelp {
 	const SCREEN_ID = 'tools_page_action-scheduler';
 
-	public function add_help_tabs() {
+	public static function add_help_tabs() {
 		$screen = get_current_screen();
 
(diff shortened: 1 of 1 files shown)


thenbrent:
I fixed this in `version_3_0_0` with https://github.com/Prospress/action-scheduler/commit/dcf05dea63ef2344329bbea92b5cf066bb70e950

I haven't added it to `master`, because I think we should lock `master` with a view to getting 3.0.0-beta tagged this week, and `version_3_0_0` being the primary branch until it's officially released.

(I decided to submit the fix without going via PR because it was such a simple issue and I didn't want to add _yet another PR_ to the list, because at the stage, even simple PRs like https://github.com/Prospress/action-scheduler/pull/324 were still awaiting review / merge).

PR #336, opened 2019-07-23
Add AsyncRequest_QueueRunner::get_sleep_seconds()

thenbrent:
To prevent crashing MySQL when processing large queues of actions via async requests.

This is similar to the sleep required by AutomateWoo, as discussed in Prospress/automatewoo#87 (private repo) and see in the code here: https://git.io/fjDvA (also private repo link).


thenbrent:
Rebased. 🚀

PR #339, opened 2019-07-24
Use factory()->get_stored_action() in DBStore

thenbrent:
Instead of duplicating the same logic.


thenbrent:
This patch was required for unit tests to pass in #333, so I've `cherry-pick`ed the commit onto that branch and will close this one.

PR #355, opened 2019-08-29
Fix php 7 deprecated notice for non static method called statically f…

scotteuser:
In php 7+ calling a method statically results in a php notice. WordPress calls the current screen help tabs statically so the help tab calls should be a static function to avoid this notice:

```
Deprecated: Non-static method ActionScheduler_AdminHelp::add_help_tabs() should not be called statically in /var/www/html/wp-includes/class-wp-hook.php on line 286
```
Thanks!

### DIFF
--- a/classes/ActionScheduler_AdminHelp.php
+++ b/classes/ActionScheduler_AdminHelp.php
@@ -7,7 +7,7 @@
 class ActionScheduler_AdminHelp {
 	const SCREEN_ID = 'tools_page_action-scheduler';
 
-	public function add_help_tabs() {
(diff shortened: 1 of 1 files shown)


thenbrent:
This has been fixed in `version_3_0_0` with SHA: dcf05de

For more background, see https://github.com/Prospress/action-scheduler/pull/335#issuecomment-514001428

However, seeing as 3.0.0 is taking longer to release than originally expected, I'll merge this one into `master` for now too. Thanks for the patch @scotteuser!

scotteuser:
Thanks! And good to know about version 3x

PR #371, opened 2019-10-16
Add ActionScheduler_DBStore::json_encode()

### LINKED ISSUE #353: JSON encoding of unicode characters
**Description:**

When using `as_schedule_single_action` and passing it args that contain international characters such as `ü`, etc., on deserialization the characters don't come back out properly encoded as the slashes (`\`) are getting dropped.

For instance, `Glückstagebuch` should get json encoded as `Gl\u00fcckstagebuch`. However, when the action callback finally runs, the slash (`\`) is missing and we just end up with `Glu00fcckstagebuch`.

**Steps to Reproduce:**

1. Make a call to ActionScheduler with international chars:

```php
as_schedule_single_action( time(), 'trigger_single_action', array( 'name' => 'Glückstagebuch' ) );
```
2. Define an action handler for the action:

```php
add_action( 'trigger_single_action', 'handle_single_action', 10, 1 );
```
3. In the action handler function, inspect the passed parameter:

```php
function handle_single_action( $name ) {
    // $name should be `Gl\u00fcckstagebuch` but instead will be `Glu00fcckstagebuch`, breaking the unicode encoding.
    var_dump( $name );
}
```

**Proposed Fix:**

In the [ActionScheduler_wpPostStore@create_post_array](https://github.com/Prospress/action-scheduler/blob/master/classes/ActionScheduler_wpPostStore.php#L35) method, `json_encode` should be called with the `JSON_UNESCAPED_UNICODE` flag like so:

```php
'post_content' => json_encode($action->get_args(), JSON_UNESCAPED_UNICODE),
```

**Comments:**

Currently, I'm having to work around it by not passing in certain params with my call to `as_schedule_single_action` and instead retrieving them in my callback. However, there are scenarios where this state may have changed since the call to `as_schedule_single_action` and I don't necessarily want the latest state, but rather the state at the time of the scheduled action call.

In my case, these are names and email addresses that I'm sending to Mailchimp. The Mailchimp API accepts unicode characters but without the slashes (`\`) or unescaped unicode, the names and email addresses end up garbled in Mailchimp with partial unicode encoding characters.

thenbrent:
Add new `ActionScheduler_DBStore::json_encode()` method to get the JSON representation of a value, with multibyte Unicode characters encoded literally and use it to encode all JSON in `ActionScheduler_DBStore`.

I looked at adding the new method to `ActionScheduler_Store` so it could be used by both `ActionScheduler_DBStore` and `ActionScheduler_wpPostStore`; however, `JSON_UNESCAPED_UNICODE` was only added in PHP 5.4, and only `ActionScheduler_DBStore` requires PHP 5.5+.

@anderly if you have a moment to test this and verify it fixes the issue reported in #353, that would be greatly appreciated.

Fixes #353

### DIFF
--- a/classes/data-stores/ActionScheduler_DBStore.php
+++ b/classes/data-stores/ActionScheduler_DBStore.php
@@ -36,7 +36,7 @@ class ActionScheduler_DBStore extends ActionScheduler_Store {
 				'status'               => ( $action->is_finished() ? self::STATUS_COMPLETE : self::STATUS_PENDING ),
 				'scheduled_date_gmt'   => $this->get_scheduled_date_string( $action, $date ),
 				'scheduled_date_local' => $this->get_scheduled_date_string_local( $action, $date ),
-				'args'                 => json_encode( $action->get_args() ),
+				'args'                 => $this->json_encode( $action->get_args() ),
 				'schedule'             => serialize( $action->get_schedule() ),
 				'group_id'             => $this->get_group_id( $action->get_group() ),
 			];
@@ -184,7 +184,7 @@ class ActionScheduler_DBStore extends ActionScheduler_Store {
 		$args[] = $hook;
 		if ( ! is_null( $params[ 'args' ] ) ) {
 			$query  .= " AND a.args=%s";
-			$args[] = json_encode( $params[ 'args' ] );
+			$args[] = $this->json_encode( $params[ 'args' ] );
 		}
 
 		$order = 'ASC';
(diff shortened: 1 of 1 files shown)


james-allan:
@thenbrent it turns out I was able to reproduce this on  Action Scheduler 2.2.4 but not 3.0. 

Looking into this a little further it appears to be something that's has been fixed now that we've moved to custom tables. 

Posts table (not escaped correctly):
Screen Shot 2019-10-21 at 4 35 44 pm

Our custom table (escaped correctly): 
Screen Shot 2019-10-21 at 4 34 51 pm
 
I'm not sure why that is exactly. I think it has something to do with the way posts are inserted. 🤷‍♂

james-allan:
> I'm not sure why that is exactly. 

Ah, it's because [the post data is passed through `wp_unslash()`](https://core.trac.wordpress.org/browser/tags/5.2/src/wp-includes/post.php#L3715) before being inserted.

thenbrent:
Ah great, I did wonder if would be fixed by using custom tables. Thanks for the confirmation @james-allan!

PR #373, opened 2019-10-16
Support updating from custom tables plugin schema

### LINKED ISSUE #367: Support updating from custom tables plugin schema
When adding the [custom tables plugin](https://github.com/Prospress/action-scheduler/blob/3.0.0-beta-1/classes/abstracts/ActionScheduler_Abstract_Schema.php#L64) to AS core in https://github.com/Prospress/action-scheduler/pull/259, we changed the class name of the action data store from `Action_Scheduler\Custom_Tables\DB_Store_Table_Maker` to `ActionScheduler_StoreSchema` (and `Action_Scheduler\Custom_Tables\DB_Logger_Table_Maker` to `ActionScheduler_LoggerSchema`).

However, `ActionScheduler_Abstract_Schema::schema_update_required()` uses the `static::class` value to [store the schema version](https://github.com/Prospress/action-scheduler/blob/3.0.0-beta-1/classes/abstracts/ActionScheduler_Abstract_Schema.php#L64) in the `wp_options` table. This value is used to determine if a table update is required.

The custom tables plugin has had [one schema change](https://github.com/Prospress/action-scheduler-custom-tables/commit/9c071774b7aa5b092f12b0c54095327e779603f4). We should to make sure Action Scheduler 3.0.0 supports updating the schema if updating from Action Scheduler Custom Tables with `schema_version === 0`.

To support this, we can fallback to check `get_option( 'schema-Action_Scheduler\Custom_Tables\DB_Store_Table_Maker', 0 )` (and `get_option( 'schema-Action_Scheduler\Custom_Tables\DB_Logger_Table_Maker', 0 )`) if the schema for `schema-ActionScheduler_StoreSchema` (or `schema-ActionScheduler_LoggerSchema`) is `0`.

thenbrent:
If no schema value is found (i.e. `0 === $version_found_in_db`), this patch will fallback to check for a schema value set by the Custom Tables plugin. If none is set, or the original schema is set, then `$version_found_in_db` will still be `0`, and the schema update will run; however, the old schema option will be deleted.

One change this patch introduces - if the latest schema from the Custom Tables plugin is set, schema `1`, then no schema update will run. I believe there have been no schema changes in adding the custom tables code to Action Scheduler core, so no schema update would be required, is that correct @rrennick?

If there has, then we should also make the schema used in AS core `2`.

Fixes #367

### DIFF
--- a/classes/abstracts/ActionScheduler_Abstract_Schema.php
+++ b/classes/abstracts/ActionScheduler_Abstract_Schema.php
@@ -64,6 +64,25 @@ abstract class ActionScheduler_Abstract_Schema {
 		$option_name         = 'schema-' . static::class;
 		$version_found_in_db = get_option( $option_name, 0 );
 
+		// Check for schema option stored by the Action Scheduler Custom Tables plugin in case site has migrated from that plugin with an older schema
+		if ( 0 === $version_found_in_db ) {
+
+			$plugin_option_name = 'schema-';
+
+			switch ( static::class ) {
+				case 'ActionScheduler_StoreSchema' :
+					$plugin_option_name .= 'Action_Scheduler\Custom_Tables\DB_Store_Table_Maker';
+					break;
+				case 'ActionScheduler_LoggerSchema' :
+					$plugin_option_name .= 'Action_Scheduler\Custom_Tables\DB_Logger_Table_Maker';
+					break;
+			}
+
+			$version_found_in_db = get_option( $plugin_option_name, 0 );
+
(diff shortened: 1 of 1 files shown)


rrennick:
> I believe there have been no schema changes in adding the custom tables code to Action Scheduler core

The `_local` date indexes were removed in https://github.com/woocommerce/action-scheduler/pull/259/commits/5af3d174e93199a69519e237674f589128ec51de . Those should be okay to leave in place in existing custom tables installs.

PR #382, opened 2019-10-25
Fix text domain and follow WordPress translation guidelines

claudiosanches:
This Pull Request fixes missing text domains, incorrect text domains, also fixing multiple placeholders to help users to proper translate in any language, also following guidelines and including notes for translators.


classes/ActionScheduler_ListTable.php
```diff
@@ -327,8 +339,10 @@ public function display_admin_notices() {
 			// No lock set or lock expired
 			if ( false === $async_request_lock_expiration || $async_request_lock_expiration < time() ) {
 				$in_progress_url       = add_query_arg( 'status', 'in-progress', remove_query_arg( 'status' ) );
-				$async_request_message = sprintf( __( 'A new queue has begun processing. %sView actions in-progress%s', 'action-scheduler' ), '<a href="' . esc_url( $in_progress_url ) . '">', ' &raquo;</a>' );
+				/* translators: %s: process URL */
+				$async_request_message = sprintf( __( 'A new queue has begun processing. <a href="%s">View actions in-progress &raquo;</a>', 'action-scheduler' ), esc_url( $in_progress_url ) );
```
claudiosanches: Some HTML tags are better to just keep in the translatable string, since it's very easy to miss or incorrect translate `%sView` or `s%s`.

classes/ActionScheduler_WPCommentCleaner.php
```diff
@@ -88,16 +88,20 @@ public static function delete_all_action_comments() {
...
 		if ( $next_scheduled_cleanup_hook ) {
-			$notice .= sprintf( __( ' This data will be deleted in %s.' ), human_time_diff( gmdate( 'U' ), $next_scheduled_cleanup_hook ) );
+			/* translators: %s: date interval */
+			$next_cleanup_message = sprintf( __( 'This data will be deleted in %s.', 'action-scheduler' ), human_time_diff( gmdate( 'U' ), $next_scheduled_cleanup_hook ) );
 		}
 
-		$notice .= sprintf( __( ' %sLearn more &raquo;%s' ), '<a href="https://github.com/woocommerce/action-scheduler/issues/368">' , '</a>' );
+		$notice = sprintf(
+			/* translators: 1: next cleanup message 2: github issue URL */
+			__( 'Action Scheduler has migrated data to custom tables; however, orphaned log entries exist in the WordPress Comments table. %1$s <a href="%2$s">Learn more &raquo;</a>', 'action-scheduler' ),
```
claudiosanches: Joins multiple translatable strings isn't good for i18n, better have a long string and join with placeholders.

classes/abstracts/ActionScheduler_Abstract_ListTable.php
```diff
@@ -104,10 +104,12 @@ abstract class ActionScheduler_Abstract_ListTable extends WP_List_Table {
 
 	/**
 	 * Makes translation easier, it basically just wraps
-	 * `_x` with some default (the package name)
+	 * `_x` with some default (the package name).
+	 * 
+	 * @deprecated 3.0.0
 	 */
 	protected function translate( $text, $context = '' ) {
```
claudiosanches: This method doesn't make any sense, it's using the `package` as text domain, but not possible to change, also if changed it, will also change some hooks in this class.

classes/abstracts/ActionScheduler_Abstract_ListTable.php
```diff
@@ -123,7 +125,7 @@ protected function get_bulk_actions() {
 				throw new RuntimeException( "The bulk action $action does not have a callback method" );
 			}
 
-			$actions[ $action ] = $this->translate( $label );
```
claudiosanches: All labels are already translatable, since they come from `ActionScheduler_ListTable`.

classes/abstracts/ActionScheduler_Abstract_ListTable.php
```diff
@@ -199,7 +201,7 @@ public function get_sortable_columns() {
 	public function get_columns() {
 		$columns = array_merge(
 			array( 'cb' => '<input type="checkbox" />' ),
-			array_map( array( $this, 'translate' ), $this->columns )
```
claudiosanches: All columns are already translatable, since they come from `ActionScheduler_ListTable`.

classes/abstracts/ActionScheduler_Abstract_ListTable.php
claudiosanches: Also note that is a bad practice to have dynamic translatable strings like that.

classes/abstracts/ActionScheduler_Abstract_ListTable.php
```diff
@@ -561,7 +563,8 @@ public function column_default( $item, $column_name ) {
 	protected function display_header() {
 		echo '<h1 class="wp-heading-inline">' . esc_attr( $this->table_header ) . '</h1>';
 		if ( $this->get_request_search_query() ) {
-			echo '<span class="subtitle">' . esc_attr( $this->translate( sprintf( 'Search results for "%s"', $this->get_request_search_query() ) ) ) . '</span>';
```
claudiosanches: This code is very bad for i18n, since will send the string already with the placeholder filled into the `translate` method.

claudiosanches:
I left some notes explains the reason for some of those changes, I hope that this may help you.

thenbrent:
Thanks for the patches @claudiosanches, and explanations on things so I understand why the changes are needed. This is LGTM. 👍

PR #388, opened 2019-11-05
restrict class autoloader to AS namespaces on namespaced classes

### LINKED ISSUE #385: "Cannot declare class" error
I am running 3.0.0-RC-2 and getting the following error:

`An error of type E_COMPILE_ERROR was caused in line 19 of the file [root directory]/wp-content/plugins/action-scheduler-master/classes/migration/Config.php. Error message: Cannot declare class Action_Scheduler\Migration\Config, because the name is already in use`

I think I've narrowed it down to a conflict with the "[Admin Columns Pro](https://www.admincolumns.com/)" plugin -- both plugins work fine when the other is deactivated but not when both are activated. I notice that Admin Columns also has a Config class, but I would have thought that the namespace would prevent conflicts. [Here's a link](https://github.com/codepress/admin-columns/blob/fd80f71741a245bcd6d446e3fccb9314f5cbbaa1/classes/Config.php#L5) to the class declaration in the free version, but note that the error does not seem to occur in the free version, so that might be a red herring. Also worth noting that the issue does not occur with Action Scheduler 2.2.5.

Entirely possible that this is ACP's fault, in which case feel free to disregard.

Steps to reproduce:

1. Activate both Action Scheduler 3.0.0-RC-2 and Admin Columns Pro 4.7.2
2. Navigate to any of: 
`/wp-admin/upload.php`
`/wp-admin/edit.php`
`/wp-admin/users.php`

rrennick:
Closes #385 

PHP calls autoloader functions in the order they are registered. WordPress loads plugins in alphabetical order so AS is one of the first plugins loaded. The AS autoloader function did not check to see if a requested namespaced class was under the `Action_Scheduler` namespace. This PR adds that check and fixes a PHPDoc comment typo.

### DIFF
--- a/classes/ActionScheduler_DataController.php
+++ b/classes/ActionScheduler_DataController.php
@@ -95,7 +95,7 @@ class ActionScheduler_DataController {
 	}
 
 	/**
-	 * Set the tick count reuired for freeing memory.
+	 * Set the tick count required for freeing memory.
 	 *
 	 * @param integer $free_ticks The number of ticks to free memory on.
 	 */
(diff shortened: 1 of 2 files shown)


thenbrent:
LGTM 👍

PR #397, opened 2019-11-22
$wpdb->prepare get_items_query_search

korndev:
Do a $wpdb->prepare on esc_like for get_items_query_search


thenbrent:
LGTM. FWIW, this code isn't actually being used anywhere because `get_items_query_search()` is only called in `ActionScheduler_Abstract_ListTable::prepare_items()` and `ActionScheduler_ListTable` overrides that methods.

PR #399, opened 2019-12-02
mark action as migrated if delete fails

### LINKED ISSUE #398: Add migration failure warning
If for some reason the post store is unable to delete an action after it has been migrated to the custom tables then the migration process continues indefinitely adding copies of the action to the custom tables store.

Reproduce using the following:
- Activate WooCommerce to have AS 2.2.5 active
- Ensure there is at least one existing action in the post store
- Add this filter
```
add_filter( 'pre_delete_post', function( $delete, $post ) {
	if ( ! class_exists( 'ActionScheduler_wpPostStore' ) ) {
		return $delete;
	}
	if ( $post->post_type === ActionScheduler_wpPostStore::POST_TYPE ) {
		return false;
	}
	return $delete;
}, 10, 2 );
```
- Activate AS 3.0.0RC

Alternatives we could consider
- stop the migration
- mark the action as migrated
- add a warning to the dashboard

rrennick:
Closes #398 

This PR changed the post status of an action to `migrated` when deleting the post after migration fails. This is a simpler fix than adding post meta as the status change removes the action from the migration queries. 

Using post meta would have added complexity to both the data store and the queries as the data store doesn't currently have any support for meta.

### DIFF
--- a/classes/data-stores/ActionScheduler_wpPostStore.php
+++ b/classes/data-stores/ActionScheduler_wpPostStore.php
@@ -12,9 +12,6 @@ class ActionScheduler_wpPostStore extends ActionScheduler_Store {
 	protected $local_timezone = NULL;
 
 	/** @var int */
-	protected $last_deleted_action = null;
-
-	/** @var int */
 	private static $max_index_length = 191;
 
 	public function save_action( ActionScheduler_Action $action, DateTime $scheduled_date = NULL ){
@@ -474,14 +471,9 @@ class ActionScheduler_wpPostStore extends ActionScheduler_Store {
 		}
 		do_action( 'action_scheduler_deleted_action', $action_id );
 
-		$this->last_deleted_action = $action_id;
-		add_action( 'action_scheduler/migrate_action_incomplete', array( $this, 'mark_migrated') );
-
 		if ( ! wp_delete_post( $action_id, TRUE ) ) {
 			$this->mark_migrated( $action_id );
 		}
-
-		remove_action( 'action_scheduler/migrate_action_incomplete', array( $this, 'mark_migrated') );
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


thenbrent:
> This is a simpler fix than adding post meta as the status change removes the action from the migration queries.

I really like this idea and approach. The implementation needs some changes though.

Specifically, I think the migration status should be set by `ActionMigrator::migrate()`, not `ActionScheduler_wpPostStore::delete_action()` to keep the migration logic encapsulated.

Ideally, `mark_migrated()` should also be defined in `ActionScheduler_Store` so we can always rely on it (that would be a breaking change if it's defined `abstract` though, which I'd like to avoid, so we might need to have some kind of fallback).

rrennick:
> I think the migration status should be set by ActionMigrator::migrate(), not ActionScheduler_wpPostStore::delete_action() to keep the migration logic encapsulated.

The reason I did it this was is that currently the action data stores do not have setters for status. Stores support marking as complete, canceled, or failed. Do we want to introduce setting a status in the store architecture this close to release?

thenbrent:
> The reason I did it this way is that currently the action data stores do not have setters for status. Stores support marking as complete, canceled, or failed. Do we want to introduce setting a status in the store architecture this close to release?

This PR's scope shouldn't expand to provide a generic `set_status()` method if that's what you mean. The existing `mark_migrated()` method proposed here is all I had in mind using. It's already `public`, so can be called by `ActionMigrator::migrate()`. We'd just need to make sure it's available and/or define it in the base `ActionScheduler_Store` class (that is the ideal implementation, but I wouldn't block the PR if it simply checked and called `ActionScheduler_wpPostStore::mark_migrated()` from within `ActionMigrator::migrate()` instead, because there are other issues that would stem from defining `ActionScheduler_Store::mark_migrated()`).

classes/data-stores/ActionScheduler_wpPostStore.php
```diff
@@ -470,7 +470,10 @@ public function delete_action( $action_id ) {
 			throw new InvalidArgumentException(sprintf(__('Unidentified action %s', 'action-scheduler'), $action_id));
 		}
 		do_action( 'action_scheduler_deleted_action', $action_id );
-		wp_delete_post($action_id, TRUE);
+
+		if ( ! wp_delete_post( $action_id, TRUE ) ) {
+			$this->mark_migrated( $action_id );
```
thenbrent: @rrennick do we want to mark this migrated here? It could mean the status is changed to migrated when non-migration related activities occur. Especially now that we may have `wpPostStore` running for a while after updating to 3.0 if an action with args > 191 chars is found, we can't assume migration will happen soon after updating anymore.

rrennick: @thenbrent Good catch. I updated the PR to add a check in the action migrator.

thenbrent: Checking for a null action is a good approach. 👍

Issue #401, opened 2019-12-08
Actions with args more than 191 characters long prevent migration in AS 3.0.0 from completing


wvega:
If an action cannot be saved to the destination data store because the action's args no longer fit the maximum number of characters in the column, the **Action Scheduler migration in progress. The list of scheduled actions may be incomplete** notice will continue to be shown indefinitely.

You can reproduce this If you create a recurring action using the code below in AS 2.2.5 and then activate AS 3.0.0.

```php
as_schedule_recurring_action(
    strtotime( '+2 hours' ),
    DAY_IN_SECONDS,
    'as_test_hook',
    array_fill( 0, 100, 9999 )
);
```

I don't think there is much that AS can do about those actions directly. The code that registered the action is likely the only one that would know how to split it into smaller actions that achieve the same result.

I saw this issue once with a task from TaxJar, but I'm not sure how common it can really be. So I guess the purpose of this issue is to raise awareness.

As a suggestion, the Action Scheduler section in the WooCommerce > System Status could show the hooks of actions that couldn't be migrated for the reason mentioned above. That would make it easier to understand why the migration never finishes if the issue comes up and would allow support requests to be redirected to the owners of the plugins that registered those actions to handle the migration.

wvega:
As a related note, AS 2.2.5 currently shows a notice when you try to create such an action, but the action is still created.

>ActionScheduler_Action::$args was called <strong>incorrectly</strong>. To ensure the action args column can be indexed, action args should not be more than 191 characters when encoded as JSON. Support for strings longer than this will be removed in a future version.

On AS 3.0.0, trying to create the same action throws a RuntimeException:

>Error saving action: Error saving action: Database error.

thenbrent:
This is a tricky one. Thanks for reporting @wvega!

> I saw this issue once with a task from TaxJar

Do you happen to know which TaxJar plugin? We could submit a patch directly. That's the only known offender atm, so I'd prefer to fix it directly ASAP rather than making more widespread changes to AS 3.0, like allowing > 191 chars again...

> As a suggestion, the Action Scheduler section in the WooCommerce > System Status could show the hooks of actions that couldn't be migrated for the reason mentioned above.

We can also:

1. make sure the action is marked as `failed` during migration; or
1. block the migration and only begin it once we know for sure there are no actions with `args` > 191 characters (the safest approach, that will ensure no data corruption, but also hacky)

@rrennick @james-allan thoughts on those two options?

[...]

PR #404, opened 2019-12-18
Provide better error message when saving an action with $args > 191 char

thenbrent:
Part of #401 following #402.

Previous error message:

> Error saving action: Database error.

New error message:

> Error saving action: ActionScheduler_Action::$args too long. To ensure the args column can be indexed, action args should not be more than 191 characters when encoded as JSON.

### DIFF
--- a/classes/data-stores/ActionScheduler_wpPostStore.php
+++ b/classes/data-stores/ActionScheduler_wpPostStore.php
@@ -12,9 +12,6 @@ class ActionScheduler_wpPostStore extends ActionScheduler_Store {
 	/** @var DateTimeZone */
 	protected $local_timezone = NULL;
 
-	/** @var int */
-	private static $max_index_length = 191;
-
 	public function save_action( ActionScheduler_Action $action, DateTime $scheduled_date = NULL ){
 		try {
 			$this->validate_action( $action );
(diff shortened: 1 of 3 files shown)


classes/data-stores/ActionScheduler_wpPostStore.php
```diff
@@ -820,8 +817,11 @@ public function migration_dependencies_met( $setting ) {
 	 * @param ActionScheduler_Action $action
 	 */
 	protected function validate_action( ActionScheduler_Action $action ) {
-		if ( strlen( json_encode( $action->get_args() ) ) > self::$max_index_length ) {
-			_doing_it_wrong( 'ActionScheduler_Action::$args', sprintf( 'To ensure the action args column can be indexed, action args should not be more than %d characters when encoded as JSON. Support for strings longer than this will be removed in a future version.', self::$max_index_length ), '2.1.0' );
+		try {
+			parent::validate_action( $action );
+		} catch ( Exception $e ) {
+			$message = sprintf( __( '%s Support for strings longer than this will be removed in a future version.', 'action-scheduler' ), $e->getMessage() );
+			_doing_it_wrong( 'ActionScheduler_Action::$args', $e->getMessage(), '2.1.0' );
```
rrennick: The second parameter should be `$message`.

PR #407, opened 2019-12-31
update FAQ for 3.0

### LINKED ISSUE #405: Update FAQ docs to reflect 3.0
The FAQ readme hasn't been updated with changes we have made in 3.0. The PR for this should be held until near the time of release.

rrennick:
Closes #405

Updated FAQ is https://github.com/woocommerce/action-scheduler/blob/update/faq/docs/faq.md.

Should the first question/paragraph reference use in themes throughout?


thenbrent:
Thanks @rrennick. This looks good. I added a couple of commits. The biggest change is to add a whole new FAQ to explain how data is stored (preceding the existing FAQ on swapping out data and replacing the FAQ on CPTs as that is less relevant now).

> Should the first question/paragraph reference use in themes throughout?

It's fine as it is now IMO. Plugins are the primary use case so it's better to focus on those in the FAQ.

PR #423, opened 2020-01-15
needs refresh update composer.json

### LINKED ISSUE #422: Minimum PHP version?
There is no PHP version defined in `composer.json` file, and from the code, I see that the library requires at least PHP 5.4, maybe even more modern.

I know that WC is PHP 5.6 and soon PHP 7+ soon, but what about the lib?

In `tests/phpunit.xml.dist` I see PHP 5.6, but that's for tests.
In `classes/ActionScheduler_DataController.php` I see this:
```php
/** Migration minimum required PHP version. */
const MIN_PHP_VERSION = '5.5';
```

rrennick:
Fixes #422 

This PR updates `composer.json` to

- specify minimum PHP version
- specify default PHP version
- update PHPUnit to 7.5
- update woocommerce sniffs

### Testing

- Run `composer install`
- Run unit tests
- Run `vendor/bin/phpcs classes` 
- phpcs should generate lots of messages

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,11 +6,19 @@
   "license": "GPL-3.0-or-later",
   "prefer-stable": true,
   "minimum-stability": "dev",
-  "require": {},
+  "require": {
+    "php": ">=5.3|>=7.0"
+  },
   "require-dev": {
-    "phpunit/phpunit": "^5.6",
+    "php": ">=7.1",
+    "phpunit/phpunit": "7.5.18",
     "wp-cli/wp-cli": "~1.5.1",
-    "woocommerce/woocommerce-sniffs": "0.0.8"
+    "woocommerce/woocommerce-sniffs": "0.0.9"
+  },
+  "config": {
+    "platform": {
+      "php": "7.1"
+    }
   },
   "scripts": {
     "test": [
--- a/composer.lock
+++ b/composer.lock
@@ -4,21 +4,21 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "fe700435a00b5cdde47170ab9811f5ae",
+    "content-hash": "7e3add449dfb60940d79207c25cbd9ac",
     "packages": [],
     "packages-dev": [
         {
             "name": "composer/ca-bundle",
-            "version": "1.2.4",
+            "version": "1.2.6",
             "source": {
                 "type": "git",
                 "url": "https://github.com/composer/ca-bundle.git",
-                "reference": "10bb96592168a0f8e8f6dcde3532d9fa50b0b527"
+                "reference": "47fe531de31fca4a1b997f87308e7d7804348f7e"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/composer/ca-bundle/zipball/10bb96592168a0f8e8f6dcde3532d9fa50b0b527",
-                "reference": "10bb96592168a0f8e8f6dcde3532d9fa50b0b527",
+                "url": "https://api.github.com/repos/composer/ca-bundle/zipball/47fe531de31fca4a1b997f87308e7d7804348f7e",
+                "reference": "47fe531de31fca4a1b997f87308e7d7804348f7e",
                 "shasum": ""
             },
             "require": {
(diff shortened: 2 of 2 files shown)


composer.json
```diff
@@ -6,11 +6,18 @@
   "license": "GPL-3.0-or-later",
   "prefer-stable": true,
   "minimum-stability": "dev",
-  "require": {},
+  "require": {
+    "php": ">=5.3|>=7.0"
```
slaFFik: 5.3 is not the correct version.
`AS_DBLogger`, `BatchFether`, `AS_DBStore` use  short array `[]` syntax, which is PHP 5.4+.

slaFFik: Also, don't forget about the Migration classes, there is PHP 5.5 checks there.

composer.json
```diff
@@ -6,11 +6,18 @@
...
+  },
   "require-dev": {
-    "phpunit/phpunit": "^5.6",
+    "phpunit/phpunit": "7.5.18",
     "wp-cli/wp-cli": "~1.5.1",
-    "woocommerce/woocommerce-sniffs": "0.0.8"
+    "woocommerce/woocommerce-sniffs": "0.0.9"
```
slaFFik: 0.0.9 version of WC sniffs require PHP 7.0, so when installing AS lib on PHP 5.6, I get this error:
```
> .\vendor\bin\phpcs classes

Parse error: syntax error, unexpected ':', expecting ';' or '{' in C:\Users\slaFFik\Sites\libs\woocommerce\action-scheduler\vendor\woocommerce\woocommerce-sniffs\src\WooCommerce-Core\Sniffs\Commenting\CommentTagsSniff.php on line 32
```

slaFFik:
So basically I can't run tests/phpcs checks locally while being on PHP 5.6.
phpunit:
> .\vendor\bin\phpunit
> This version of PHPUnit is supported on PHP 7.1, PHP 7.2, and PHP 7.3.
> You are using PHP 5.6.38 (c:\openserver\modules\php\PHP-5.6-x64\php.exe).

phpcs:
> .\vendor\bin\phpcs classes
> Parse error: syntax error, unexpected ':', expecting ';' or '{' in C:\Users\slaFFik\Sites\libs\woocommerce\action-scheduler\vendor\woocommerce\woocommerce-sniffs\src\WooCommerce-Core\Sniffs\Commenting\CommentTagsSniff.php on line 32

composer.json
rrennick: > AS_DBLogger, BatchFether, AS_DBStore use short array [] syntax, which is PHP 5.4+.

The migration classes and tables data stores are only loaded if the `MIN_PHP_VERSION` is met. AS runs in deprecated mode using the post store on 5.3-4. The automated unit testing still runs against 5.3 & 5.4.

rrennick:
> With platform.php=7.1 and strict exact versions of phpunit/sniffs everything is installed as I'm on 7.1, which is not the case.

Thanks, nice catch. 

@thenbrent How do you feel about requiring 7.1 for dev and if a developer has a need to develop with an older version of PHP, they could use PHPUnit installed separately?

composer.json
slaFFik: So if I want to use custom tables to store the queue I'm required to be on PHP 5.5+, correct? On older PHP everything will be stored in wp_posts table.

thenbrent: @slaFFik yes that's correct.

thenbrent:
> @thenbrent How do you feel about requiring 7.1 for dev and if a developer has a need to develop with an older version of PHP, they could use PHPUnit installed separately?

That's only when using Composer right? If so, that's fine. If people are using other methods (e.g. the subtree in Subscriptions), then I don't want to enforce that.

composer.json
slaFFik: @thenbrent Does the lib handles the migration of a live site from `wp_posts` to custom tables when PHP version is changed on a server from 5.4 to 5.5?

rrennick: Does the lib handles the migration of a live site from wp_posts to custom tables when PHP version is changed on a server from 5.4 to 5.5?

@slaFFik Thanks for your questions. You highlighted that we needed to add some migration documentation. I created #424 with a v3.0 specific FAQ.

rrennick:
> So basically I can't run tests/phpcs checks locally while being on PHP 5.6.

@slaFFik In my setup I have PHPUnit 6 set up globally with its path added to my `.bash_profile`. I can run either the global or project version.

```
# run global PHPUnit
$ phpunit
# run project PHPUnit
$ vendor/bin/phpunit
```

> That's only when using Composer right?

That's the intention. @james-allan could you sub-tree in this branch and verify it doesn't interfere with running unit tests in WC Subscriptions?

rrennick:
@barryhughes This PR is probably safe to close.

PR #439, opened 2020-01-27
add check that demarkation ID was set in actions table

### LINKED ISSUE #437: Demarkation ID not set in actionscheduler_actions
We have had 2 reports of customer sites having fatal errors querying action ID `2` in the post store. I investigated a staging site for one of the two customers and the `actionscheduler_actions.action_id` field starts at `2`. 

We need to add a check to verify that the IDENTITY was set and if not attempt to use `ALTER TABLE` to set it. If both fail the migration should be disabled & the site administrator notified of an issue.

rrennick:
Fixes #437

This PR adds a check that the next identity value gets set on the `_actions` table. 

I'm not sure what conditions could be used to prevent the identity field being set on `INSERT` so I don't have steps to reproduce the issue.

### DIFF
--- a/classes/data-stores/ActionScheduler_HybridStore.php
+++ b/classes/data-stores/ActionScheduler_HybridStore.php
@@ -85,6 +85,24 @@ class ActionScheduler_HybridStore extends Store {
 					'status'    => '',
 				]
 			);
+			// Verify the correct action_id was set.
+			$table_name = $wpdb->{ActionScheduler_StoreSchema::ACTIONS_TABLE};
+			$action_id = $wpdb->get_var( "SELECT MAX(action_id) FROM {$table_name}" ); // phpcs:ignore WordPress.DB.PreparedSQL.InterpolatedNotPrepared
+			// If it's lower use an update to set it.
+			if ( $action_id < $this->demarkation_id ) {
+				$wpdb->update(
+					$wpdb->{ActionScheduler_StoreSchema::ACTIONS_TABLE},
+					[
+						'action_id' => $this->demarkation_id,
+					],
+					[
+						'action_id' => $action_id,
+					],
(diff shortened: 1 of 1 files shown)


classes/data-stores/ActionScheduler_HybridStore.php
```diff
@@ -85,6 +85,24 @@ public function set_autoincrement( $table_name, $table_suffix ) {
 					'status'    => '',
 				]
 			);
+			// Verify the correct action_id was set.
+			$table_name = $wpdb->{ActionScheduler_StoreSchema::ACTIONS_TABLE};
+			$action_id = $wpdb->get_var( "SELECT MAX(action_id) FROM {$table_name}" ); // phpcs:ignore WordPress.DB.PreparedSQL.InterpolatedNotPrepared
+			// If it's lower use an update to set it.
+			if ( $action_id < $this->demarkation_id ) {
```
thenbrent: This issue could occur due to `$this->demarkation_id` being `0`, or something else less than the actual `MAX(ID) from wp_posts` value. In that case, even with this change, the issue would persist.

rrennick: Since we don't have any data to go on it's 50/50 that the ID was 0 or there was an issue with inserting the identity. I think what I will look at is adding a check for the ID > 2 or block the migration to this PR.

thenbrent: > what I will look at is adding a check for the ID > 2 or block the migration to this PR.

That's a good approach. 👍

rrennick:
Let's hold off on this as the source of this issue could have been #449.

rrennick:
Closing as this was addressed in #503

PR #440, opened 2020-01-28
move action_execute hook to after status check

rrennick:
This PR replaces the existing `action_scheduler_before_execute` with `action_scheduler_before_process`. It moves `action_scheduler_before_execute` to after the action status has been checked which ensures that the `$action_id` is the ID of a valid action. 

It also makes the `action_scheduler_failed_execution` hook conditional on `$action_id` referencing a valid pending action.

In master, when invalid action IDs are passed to the queue runner, two log entries are written to the action log table. If there is a conflict where the queue runner is consistently being passed an invalid action ID, over time the async runner will insert large numbers of log entries into the table.

The approach of moving the timing of the hooks should maintain compatibility with custom data stores while eliminating the default behaviour of writing log entries for invalid action IDs.

### DIFF
--- a/classes/abstracts/ActionScheduler_Abstract_QueueRunner.php
+++ b/classes/abstracts/ActionScheduler_Abstract_QueueRunner.php
@@ -49,6 +49,7 @@ abstract class ActionScheduler_Abstract_QueueRunner extends ActionScheduler_Abst
 	 */
 	public function process_action( $action_id, $context = '' ) {
 		try {
+			$valid_action = false;
 			do_action( 'action_scheduler_before_execute', $action_id, $context );
 
 			if ( ActionScheduler_Store::STATUS_PENDING !== $this->store->get_status( $action_id ) ) {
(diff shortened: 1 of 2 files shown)


rrennick:
The failed test on this is one of the timing tests that is off by 1 second.

thenbrent:
Moving the order the hook is called in can be considered a breaking change. Could we introduce the new `action_scheduler_before_process` in the new position instead? The `if ( $valid_action )` check and corresponding logic and new failed hook could remain in their current positions.

rrennick:
> Could we introduce the new action_scheduler_before_process in the new position instead?

Yes, we could do that. The log entries are attached to the 2 hooks that affected by the change. We would need to switch to start log entry to the _process hook. Does that sound like the better choice?

thenbrent:
@rrennick yeah I think it's fine to switch _when_ the log entry is made.

rrennick:
@thenbrent This one has been updated.

PR #442, opened 2020-01-28
add spacing for text domain replacement when used as a composer dependency

### LINKED ISSUE #435: Add consistent spacing on text domains
The composer script in https://github.com/woocommerce/woocommerce-admin/blob/try/package-as/bin/composer-package-update.sh misses several text domains because some text domains do not have the spacing between the `'` and the `)`.

We'll have the same issue in WC https://github.com/woocommerce/woocommerce/pull/25404#issuecomment-577209545

rrennick:
Fixes #435.

This PR adds white space around text domain strings to allow composer dependency scripts to replace text domains. See https://github.com/woocommerce/woocommerce-admin/pull/3557 as an example.


rrennick:

PR #443, opened 2020-01-28
add composer autoloading

### LINKED ISSUE #434: Investigate Composer PSR-4 autoloading
Per https://github.com/woocommerce/woocommerce/pull/25404#pullrequestreview-345255964 we'll need to look at switching to using the composer autoloading to allow AS to be added to WooCommerce (and other plugins) as an external.

rrennick:
Fixes #434 

This PR adds composer autoloading. The existing autoloading code is the fallback when `composer install` has not been run which should avoid any backward compatibility issues.

### Testing

1. Add an `error_log()` to `ActionScheduler::autoload()` where a class file is included.
1. Test the following matrix

#### Steps
1. Schedule an action with `as_schedule_single_action()`
1. Schedule an action with `as_schedule_cron_action()`
1. Load the Admin screen
1. Test WP CLI `wp action-scheduler run`

__Against__
 
#### Setup
1. Remove the `vendor` folder
  - This should log to the debug log as classes are loaded
1. Run `composer install --no-dev`
  - This should not log to the debug log
1. Run `composer install`
  - This should not log to the debug log

### DIFF
--- a/action-scheduler.php
+++ b/action-scheduler.php
@@ -40,7 +40,14 @@ if ( ! function_exists( 'action_scheduler_register_3_dot_0_dot_1' ) ) {
 	}
 
 	function action_scheduler_initialize_3_dot_0_dot_1() {
-		require_once( 'classes/abstracts/ActionScheduler.php' );
+		$autoloader = __DIR__ . '/vendor/autoload.php';
+		if ( is_readable( $autoloader ) ) {
+			require_once( $autoloader );
+			define( 'AS_COMPOSER_AUTOLOADING', true );
+		} else {
+			define( 'AS_COMPOSER_AUTOLOADING', false );
+			require_once( 'classes/abstracts/ActionScheduler.php' );
+		}
 		ActionScheduler::init( __FILE__ );
 	}
 
--- a/classes/ActionScheduler_AdminView.php
+++ b/classes/ActionScheduler_AdminView.php
@@ -104,12 +104,13 @@ class ActionScheduler_AdminView extends ActionScheduler_AdminView_Deprecated {
 			return;
 		}
 
+		$as_version = ActionScheduler_Versions::instance()->latest_version();
 		$screen->add_help_tab(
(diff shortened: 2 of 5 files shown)


rrennick:

action-scheduler.php
```diff
@@ -40,7 +40,14 @@ function action_scheduler_register_3_dot_0_dot_1() {
 	}
 
 	function action_scheduler_initialize_3_dot_0_dot_1() {
-		require_once( 'classes/abstracts/ActionScheduler.php' );
+		$autoloader = __DIR__ . '/vendor/autoload.php';
+		if ( is_readable( $autoloader ) ) {
+			require_once( $autoloader );
+			define( 'AS_USING_AUTOLOADING', true );
```
thenbrent: Using `spl_autoload_register()` is also using autloading, so this constant could be confusing. What about something like `AS_COMPOSER_AUTOLOADING` instead?

rrennick: > will this be backward compatible? e.g. what happens when someone has AS 2.2

It's compatible with previous versions. Previous versions of AS initialize the autoloader in `ActionScheduler::init( __FILE__ );`. The PR adds the Composer autoloader immediately before that call. 

> could we notify admins (including HEs), likely somewhere in WP Admin, that Composer autoloading is being used (and perhaps the version of AS being loaded)?

We can add these to the help screen & SSR report.

thenbrent: > The PR adds the Composer autoloader immediately before that call.

Ah of course. So the newest version of this method still always gets called, in which case, the newest version of the autoloader will always work.

rrennick: @thenbrent I added the AS version to both the help screen and the SSR. I thought it might be confusing to users to see the autoloading info on the help screen so it's only on the SSR.

thenbrent: @rrennick yeah you're right, good idea to hide it on the help screen.

PR #455, opened 2020-02-10
Allow alternate storage engine

### LINKED ISSUE #418: Migrations fails because args too long
How are we suppose to migrate actions to 3.0 with "long" arguments?

None of our actions can be migrated because:

```
Error saving action: Error saving action: ActionScheduler_Action::$args too long. To ensure the args column can be indexed, action args should not be more than 191 characters when encoded as JSON.
```

An example of our args:

```
'data' => array ('opf' => '300', 'firstname' => 'John Doe', 'email' => 'user@example.com', 'page' => 'https://example.com/programs/foobar/', 'user-agent' => 'Mozilla/5.0 (Linux; Android 5.1; A1601 Build/LMY47I) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/61.0.3163.98 Mobile Safari/537.36', 'ip' => '110.54.151.154', )
```

rrennick:
Fixes #418 

This PR 
- Updates the base store class to allow a datastore to set the maximum `$args` length on an action
- Updates the table store to support a maximum of 8000 characters encoded. The maximum record length is 65535 bytes (excluding `*text` fields). 8000 * 4 (for utf8mb4) = 32000 bytes. 8000 is the largest we should go without switching to a `*text` field.
- It does not remove the deprecated warning on the args length in the post store

### Testing Procedure

- Switch to AS tag 2.2.5
- Deactivate WC 4.0 beta and WC Subscriptions 3.0 if on those versions
- Add the following snippet 
```
	$action_id = as_schedule_single_action( time() + 50, 'test_action', range( 100, 150 ) );
	error_log( 'test_action: ' . $action_id );
	$found_action = as_next_scheduled_action( 'test_action', range( 100, 150 ) );
	error_log( 'Schedule: ' . gmdate( 'Y-m-d H:i:s O', $found_action ) );
} );
```
- This will log to the debug log
```
[10-Feb-2020 19:48:04 UTC] PHP Notice:  ActionScheduler_Action::$args was called <strong>incorrectly</strong>. To ensure the action args column can be indexed, action args should not be more than 191 characters when encoded as JSON. Support for strings longer than this will be removed in a future version. Please see <a href="https://wordpress.org/support/article/debugging-in-wordpress/">Debugging in WordPress</a> for more information. (This message was added in version 2.1.0.) in /Sites/ron/wp-includes/functions.php on line 4986
[10-Feb-2020 19:48:04 UTC] test_action: 11084
[10-Feb-2020 19:48:04 UTC] Schedule: 2020-02-10 19:48:48 +0000
```
- If your migration was already complete `wp option delete action_scheduler_migration_status`
- Switch to this branch
- Refresh the dashboard then disable the snippet above
- Debug log should have
```
[10-Feb-2020 19:48:04 UTC] test_action: 12345
[10-Feb-2020 19:48:04 UTC] Schedule: 2020-02-10 19:48:48 +0000
```
- After allowing the migration to run 
- Check that test actions created on 2.2.5 are migrated (the original action ID can be found in the migration entry on the action log).
- Check that test actions created on this branch are there 

- go to Tools -> Scheduled actions
- enter `149` in the search box
- `test_action` actions should be found

### DIFF
--- a/classes/abstracts/ActionScheduler_Store.php
+++ b/classes/abstracts/ActionScheduler_Store.php
@@ -16,7 +16,7 @@ abstract class ActionScheduler_Store extends ActionScheduler_Store_Deprecated {
 	private static $store = NULL;
 
 	/** @var int */
-	private static $max_index_length = 191;
+	protected static $max_args_length = 191;
 
 	/**
 	 * @param ActionScheduler_Action $action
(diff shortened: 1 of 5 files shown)


rrennick:
@peterfabian Can you give this one a check with webhooks?

peterfabian:
Thanks for the ping, I'll check it out a bit later today.

peterfabian:
I tested this by creating a bunch of actions with AS 2.2.5 + WC 3.9.1, some with long and some with short args. All the actions have been migrated correctly. Also new actions with long args seem to work correctly, webhooks are firing, filtering and searching in the Scheduled Actions list works, too. 
Great work!

rrennick:
Thanks @peterfabian

tillkruss:
@rrennick: Are the `extended_args` searchable?

rrennick:
> Are the extended_args searchable?

Yes, with the `search` parameter. When looking for the next scheduled action, the passed args hash is compared to the DB args hash.

PR #458, opened 2020-02-11
unhook async request dispatch after dispatching request

### LINKED ISSUE #457: Possible deadlock diagnosis - shutdown and loopback requests
We've seen a few issues reporting deadlocks, namely #438 and #441. I think I've found the root cause.

The async runner is attached to the `'shutdown'` hook in `ActionScheduler_QueueRunner::maybe_dispatch_async_request()`. That's the way AS starts to process an async queue.

However, the async queue runner also uses loopback requests to chain processing of actions. This is done in `ActionScheduler_AsyncRequest_QueueRunner::maybe_dispatch()`, which is called from `ActionScheduler_AsyncRequest_QueueRunner::handle()`.

While `ActionScheduler_AsyncRequest_QueueRunner::maybe_dispatch()` only dispatches a loopback request when there isn't already a queue processing, it doesn't actually stop _a second request being made by `ActionScheduler_QueueRunner::maybe_dispatch_async_request()` on `'shutdown'` (`ActionScheduler_AsyncRequest_QueueRunner::maybe_dispatch()` runs in the context of admin ajax, so `is_admin()` would return `true` in `ActionScheduler_AsyncRequest_QueueRunner::maybe_dispatch()`, and the `async-request-runner` is only for requests initiated via the `'shutdown'` hook, not those initiated via `ActionScheduler_AsyncRequest_QueueRunner`).

### Proposed Solution

There's a couple of ways we could fix this:

1. Remove `ActionScheduler_QueueRunner::maybe_dispatch_async_request()` from `shutdown`  in `ActionScheduler_AsyncRequest_QueueRunner::maybe_dispatch()` after calling `$this->dispatch()`
2. call `ActionScheduler::lock()->set( 'async-request-runner' )` in `ActionScheduler_AsyncRequest_QueueRunner::maybe_dispatch()` to set the lock used by `ActionScheduler_QueueRunner::maybe_dispatch_async_request()`
3. or both to be safe?

rrennick:
Fixes #457 

This PR unhooks `maybe_dispatch_async_request()` after dispatching an async request. This will prevent a second thread from attempting to disptch another async request.

### DIFF
--- a/classes/ActionScheduler_AsyncRequest_QueueRunner.php
+++ b/classes/ActionScheduler_AsyncRequest_QueueRunner.php
@@ -69,6 +69,7 @@ class ActionScheduler_AsyncRequest_QueueRunner extends WP_Async_Request {
 		}
 
 		$this->dispatch();
+		ActionScheduler_QueueRunner::instance()->unhook_dispatch_async_request();
 	}
 
 	/**
--- a/classes/ActionScheduler_QueueRunner.php
+++ b/classes/ActionScheduler_QueueRunner.php
@@ -65,8 +65,21 @@ class ActionScheduler_QueueRunner extends ActionScheduler_Abstract_QueueRunner {
 		}
 
 		add_action( self::WP_CRON_HOOK, array( self::instance(), 'run' ) );
+		$this->hook_dispatch_async_request();
+	}
 
-		add_filter( 'shutdown', array( $this, 'maybe_dispatch_async_request' ) );
+	/**
+	 * Hook check for dispatching an async request.
+	 */
+	public function hook_dispatch_async_request() {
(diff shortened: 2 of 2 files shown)


rrennick:

thenbrent:
@rrennick thanks for getting a PR up quickly. I'm curious why you chose to go with unhooking instead of `ActionScheduler::lock()->set( 'async-request-runner' )`? (Or both options)

Setting the lock _might_ be a slightly safer option, because it covers all requests not just the current request.

thenbrent:
LGTM. As discussed, the plan is to start with unhooking the queue runner and see how that goes on the up-coming WC Admin release. Then if needed, we can also set the lock.

Issue #464, opened 2020-02-13
Proper composer autoloading


slaFFik:
I decided to merge several issues in one and list them all here and provide a solution.
I'm writing here from point of view of a developer, who already uses composer in their own plugin and wants to include AS as yet another package.

Right now AS doesn't properly use composer autoload. Perhaps, there are some reasons for that, I don't know them.
Because of the way AS 3.1.0 is loading ([using classmap](https://github.com/woocommerce/action-scheduler/blob/400725e8f80086f9f584a17b17b5352e27df5b52/composer.json#L37), implemented in #434, + manual file require), it's not possible to use a correct way to install packages and use the ordinary composer autoload functionality.

So, this doesn't work:
```php
require_once __DIT__ . 'vendor/autoload.php';
```
You need to explicitly require AS main file, and do it BEFORE composer autoload, like this:
```php
require_once __DIR__ . '/vendor/woocommerce/action-scheduler/action-scheduler.php';
require_once __DIR__ . '/vendor/autoload.php';
```
This is only relevant for AS 3.1.0 (because of #434).
AS 3.0.1 works fine when included via composer and initiated directly like this:
```php
require_once __DIR__ . '/vendor/autoload.php';
require_once __DIR__ . '/vendor/woocommerce/action-scheduler/action-scheduler.php';
```

With this v3.1.0 approach, composer is not used at all to load AS, everything is done in AS and its `classmap` is useless, afaik.

My investigation in #463 gave me that extra information, and understanding, that perhaps this situation can be improved because I'm not the only one (see #445 as an example).

So, I suggest to remove `"classmap"` and add `"files"`:
```
"autoload": {
    "files": ["action-scheduler.php"]
},
```

This way a proper way to include packages, using
```php
require_once __DIT__ . 'vendor/autoload.php';
```
will be possible, without the need to require `action-scheduler.php` in own plugin.

rrennick:
@slaFFik Thanks for the explanation. The way Action Scheduler currently works when there are multiple plugins active that include AS is

- `aa` has AS `2.2.5`
- `bb` has AS `3.1.0`
- `cc` has AS `3.0.1`

When WP loads each plugin (in alphabetical order) 
- `action-scheduler.php` registers its version with a callback for initializing that version
- on `plugins_loaded` AS runs the callback of the latest version
- the callback registers the autoloader for that version

In this example, only the autoloader of `bb` would be registered. 

[...]

PR #468, opened 2020-02-18
always hook initialize_latest_version()

rrennick:
See #464 

This PR is to address 

> You need to explicitly require AS main file, and do it BEFORE composer autoload, like this:

```
require_once __DIR__ . '/vendor/woocommerce/action-scheduler/action-scheduler.php';
require_once __DIR__ . '/vendor/autoload.php';
```

If a Composer autoloader is already active in the containing plugin then the `class_exists( 'ActionScheduler_Versions' )` check returns `true`. Because that check returns true, `initialize_latest_version` does not get hooked. Before adding a hook, WP checks whether it has already been hooked so adding a hook multiple times will not cause it to be executed multiple times.

### Testing Instructions

- Check out https://github.com/woocommerce/woocommerce-admin/pull/3557
- `npm install`
- `composer install`
- deactivate all plugins except WC & WC Admin
- add an action
```
	$action_id = as_schedule_single_action( time() + 50, 'function_check' );
} );
```
- load the dashboard should produce a fatal error
- copy `action-scheduler.php` from this branch to the `woocommerce-admin/external/action-scheduler` folder
- dashboard should load
- activate AS in this branch
- there should be no errors in the debug log

### DIFF
--- a/action-scheduler.php
+++ b/action-scheduler.php
@@ -29,9 +29,9 @@ if ( ! function_exists( 'action_scheduler_register_3_dot_1_dot_0' ) ) {
 
 	if ( ! class_exists( 'ActionScheduler_Versions' ) ) {
 		require_once( 'classes/ActionScheduler_Versions.php' );
-		add_action( 'plugins_loaded', array( 'ActionScheduler_Versions', 'initialize_latest_version' ), 1, 0 );
 	}
 
+	add_action( 'plugins_loaded', array( 'ActionScheduler_Versions', 'initialize_latest_version' ), 1, 0 );
 	add_action( 'plugins_loaded', 'action_scheduler_register_3_dot_1_dot_0', 0, 0 );
 
 	function action_scheduler_register_3_dot_1_dot_0() {


rrennick:
I want to do some more testing with this ti verify it doesn't create the problem I outlined in https://github.com/woocommerce/action-scheduler/issues/464#issuecomment-586284949

rrennick:
I used 3 plugins where two had composer autoloading and version bumps as follows

- AS - this branch
- `as-test` version 3.2.1
- `woocommerce-admin` version 3.1.2

I added `error_log( __FILE__ )` in both the `as_schedule_single_action` and `ActionScheduler_ActionFactory` with this test snippet:

```
	error_log( ActionScheduler_Versions::instance()->latest_version() );
	$action_id = as_schedule_single_action( time() + 50, 'which_one' );
} );
```

The result was

```
[18-Feb-2020 17:19:17 UTC] 3.2.1
[18-Feb-2020 17:19:17 UTC] /Users/ronrennick/Sites/ron/wp-content/plugins/as-test/vendor/woocommerce/action-scheduler/functions.php
[18-Feb-2020 17:19:17 UTC] /Users/ronrennick/Sites/ron/wp-content/plugins/woocommerce-admin/external/action-scheduler/classes/ActionScheduler_ActionFactory.php
```

So, I'll leave this on hold for now pending further investigation.

Issue #460, opened 2020-02-13
[3.0] Actions lost when migrating if `action_scheduler_migration_status` option is already set


vedanshujain:
Option `action_scheduler_migration_status` can be set by either updating to 4.0 beta, or more realistically by installing WooCommerce Services plugin which ships AS 3.0.

The problem is, if you have previously installed, but then disabled the WooCommerce Services plugin, the option still remains set, which means that when you upgrade to WC 4.0, all your current actions will lost because AS 3.0 will think that migration is already completed. 

We should perhaps use a different method for checking migration status, like count(post_type = scheduled_action) > 0.

rrennick:
@vedanshujain Thanks for reporting this. We already encountered this when WC Subscriptions  3.0 was released and WooCommerce Memberships had unintentionally shipped a version with the AS 3.0 beta. 

For the affected customers, the happiness team has instructions for deleting the option via the dashboard (which reactivates the migration). From my perspective, based on the number of reports from above, this is an edge case that doesn't warrant adding a frequent query indefinitely to all WC installs. But, I'll leave this open for discussion.

peterfabian:
I _think_ it might be better to check for posts, as there is a default WP `KEY type_status_date (post_type,post_status,post_date,ID),`, it shouldn't be too costly to check. 

Especially since some cleanup actions (`\ActionScheduler_WPCommentCleaner::init`) depend only on this option being set, potentially deleting action logs data?

rrennick:
As an alternative, would deleting the option in the WC 4.0 upgrade routine cover most of the edge cases? If there aren't any actions in the posts table it will be marked as complete again within a couple minutes.

vedanshujain:
Deleting the option in upgrade routine, as well as in plugins activate/deactivate hooks sounds like a good approach to me. Although I would say that since we are shipping AS 3.0 with WC Service for some time now, its probably not an edge case anymore.

rrennick:
Addressed in #470.

jaredatch:
> For the affected customers, the happiness team has instructions for deleting the option via the dashboard (which reactivates the migration). From my perspective, based on the number of reports from above, this is an edge case that doesn't warrant adding a frequent query indefinitely to all WC installs

Just a heads up. WPForms 1.5.9 includes AS 3.1. We've seen this issue come up at least a dozen times that have been reported.

[...]

PR #470, opened 2020-02-19
Unmark migration when a plugin is deactivated to prevent action loss.

vedanshujain:
If a plugin that loads AS 3.x is deactivated, flag for migration completed still remains. This means that if remaining plugins have highest active AS version of 2.x, then in future if a plugin with AS 3.x is installed, migration routine will not kick in because migration flag was already set.

An example of this is when WC Services with loads AS 3.x is disabled and WC 3.9 is enabled on the site, then this effectively downgrades AS from 3.x to 2.x. Then if WC is updated from 3.x to 4.x, action migration will not kick in because migration was marked completed by WC services back when it was updated/installed.

Also, since we cannot reliably determine what AS version will remain after deactivation, hook is triggered for all plugin deactivations.

### DIFF
--- a/classes/ActionScheduler_DataController.php
+++ b/classes/ActionScheduler_DataController.php
@@ -65,6 +65,15 @@ class ActionScheduler_DataController {
 	}
 
 	/**
+	 * Unmark migration when a plugin is de-activated. Will not work in case of silent activation, for example in an update.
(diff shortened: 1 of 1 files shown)


vedanshujain:
Fixes #460

PR #472, opened 2020-02-21
remove composer autoloading

### LINKED ISSUE #471: Remove composer autoloading
After an initial report in #463, testing for #464 and a followup PR #468, I've concluded that composer autoloading needs to be completely removed from AS.

### From #464:

> So, I suggest to remove "classmap" and add "files":

```
"autoload": {
    "files": ["action-scheduler.php"]
},
```

If you have two plugin active that are both configured to autoload a file named `example.php` only  the `example.php` in the plugin that has its autoloader registered first will be autoloaded. Composer does not compare the files to determine whether they contain the same code. Instead, it assumes the files are duplicates.

### Current AS configuration:

```
  "autoload": {
    "classmap": [
      "classes/",
      "deprecated/",
      "lib/"
    ]
```

If all plugins using AS updated to this configuration via composer then AS would not get initialized. The autoloader in the containing plugins would already have a class map for all the AS classes and the `class_exists( 'ActionScheduler_Versions' )` check returns true

### Code change in #468:

While this PR does correct the issue of initializing AS, the classmap(s) used are in the order that the plugin autoloaders are registered (thereby breaking the latest version logic). Since the `functions.php` isn't autoloaded, the `functions.php` used is from the latest registered version.

rrennick:
Fixes #471 

This PR reverts all of https://github.com/woocommerce/action-scheduler/commit/12daab2e6a71d270673926e74423491e70bcdfca plus the related code in https://github.com/woocommerce/action-scheduler/commit/e496d28ce85ad619905b179ff98ab1c793630fcd.

### Testing

- Verify that AS loads correctly in the dashboard
- Change the registered version to `3.1.1`
- Check the version being loaded in the WC system status report
- Schedule a test action and allow it to process

### DIFF
--- a/action-scheduler.php
+++ b/action-scheduler.php
@@ -40,14 +40,7 @@ if ( ! function_exists( 'action_scheduler_register_3_dot_1_dot_0' ) ) {
 	}
 
 	function action_scheduler_initialize_3_dot_1_dot_0() {
-		$autoloader = __DIR__ . '/vendor/autoload.php';
-		if ( is_readable( $autoloader ) ) {
-			require_once( $autoloader );
-			define( 'AS_COMPOSER_AUTOLOADING', true );
-		} else {
-			define( 'AS_COMPOSER_AUTOLOADING', false );
-			require_once( 'classes/abstracts/ActionScheduler.php' );
-		}
+		require_once( 'classes/abstracts/ActionScheduler.php' );
 		ActionScheduler::init( __FILE__ );
 	}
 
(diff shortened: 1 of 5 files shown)


action-scheduler.php
```diff
@@ -40,14 +40,7 @@ function action_scheduler_register_3_dot_1_dot_0() {
 	}
 
 	function action_scheduler_initialize_3_dot_1_dot_0() {
```
timmyc: Does this method name need to be changed at all to match the next version? 3.1.1?

rrennick: > Does this method name need to be changed at all to match the next version? 3.1.1?

Yes, we do the version bump in its own PR #473.

PR #474, opened 2020-02-28
Various documentation and phpDoc fixes

### LINKED ISSUE #453: Usage -> Scheduling an Action: incorrect example in strtotime()
Right now in `docs/usage.md` in "Scheduling an Action" section you provide an example of recurring action, and for the timestamp you write this:

```php
strtotime( 'midnight tonight' )
```

This is an incorrect argument according to the PHP docs: there is no such [relative format](https://www.php.net/manual/en/datetime.formats.relative.php) as `tonight`.
We have either `midnight` (which is today's midnight) or `tomorrow` (which is tomorrow's midnight). 
This function returns `false` and that means the action is incorrectly scheduled (in my case I'm using `(int)` and thus I have `0` - the start of the Unix epoch).

I assume you meant `tomorrow`.

I'm not creating a PR because the whole Usage doc should be hugely modified, so there is no sense in small subsequent improvements when a general overhaul is planned (see #432).

### LINKED ISSUE #469: Typo in as_next_scheduled_action() description
`as_next_scheduled_action()` phpdoc description:
> An action in the queue could be pending, in-progress or aysnc...

Should be `async`.

crstauf:
Corrects several mistakes in the documentation, as well as in phpDoc blocks.

Fixes #453.
Fixes #469.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -40,7 +40,7 @@ as_enqueue_async_action( $hook, $args, $group )
 
 - **$hook** (string)(required) Name of the action hook. Default: _none_.
 - **$args** (array) Arguments to pass to callbacks when the hook triggers. Default: _`array()`_.
-- **$group** (array) The group to assign this job to. Default: _''_.
+- **$group** (string) The group to assign this job to. Default: _''_.
 
 ### Return value
 
@@ -64,7 +64,7 @@ as_schedule_single_action( $timestamp, $hook, $args, $group )
 - **$timestamp** (integer)(required) The Unix timestamp representing the date you want the action to run. Default: _none_.
 - **$hook** (string)(required) Name of the action hook. Default: _none_.
 - **$args** (array) Arguments to pass to callbacks when the hook triggers. Default: _`array()`_.
-- **$group** (array) The group to assign this job to. Default: _''_.
+- **$group** (string) The group to assign this job to. Default: _''_.
 
 ### Return value
 
@@ -89,7 +89,7 @@ as_schedule_recurring_action( $timestamp, $interval_in_seconds, $hook, $args, $g
 - **$interval_in_seconds** (integer)(required) How long to wait between runs. Default: _none_.
 - **$hook** (string)(required) Name of the action hook. Default: _none_.
 - **$args** (array) Arguments to pass to callbacks when the hook triggers. Default: _`array()`_.
-- **$group** (array) The group to assign this job to. Default: _''_.
+- **$group** (string) The group to assign this job to. Default: _''_.
 
 ### Return value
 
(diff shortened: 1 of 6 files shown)


functions.php
```diff
@@ -7,10 +7,10 @@
 /**
  * Enqueue an action to run one time, as soon as possible
  *
- * @param string $hook The hook to trigger.
- * @param array  $args Arguments to pass when the hook triggers.
- * @param string $group The group to assign this job to.
- * @return string The action ID.
+ * @param string $hook The hook to trigger
```
rrennick: The trailing periods should be left for this block to pass the sniffs.

docs/api.md
```diff
@@ -158,12 +158,12 @@ as_unschedule_action( $hook, $args, $group )
...
 - **$hook** (string)(required) Name of the action hook.
-- **$args** (array) Arguments to pass to callbacks when the hook triggers. Default: _`array()`_.
-- **$group** (array) The group to assign this job to. Default: _''_.
+- **$args** (array) Arguments passed to callbacks when the hook triggers. Default: _`array()`_.
+- **$group** (string) The group the job is assigned to. Default: _''_.
 
 ### Return value
 
-(null)
+(string|null) The scheduled action ID if a scheduled action was found, or null if no matching action found.
```
rrennick: I'm okay with changing the Action ID type to int so long as it's updated to that wherever we make changes.

functions.php
```diff
@@ -24,7 +24,7 @@ function as_enqueue_async_action( $hook, $args = array(), $group = '' ) {
  * @param array $args Arguments to pass when the hook triggers
  * @param string $group The group to assign this job to
  *
- * @return string The job ID
+ * @return int The job ID
```
rrennick: Let's update these to be `action ID` for consistency.

rrennick (changes requested):
@crstauf Thanks for submitting this. I've noted a few things to change before merging.

docs/api.md
crstauf: I'm mostly interested in getting the documentation corrected in this PR. I would certainly prefer for this function to return `int` rather than `string`, and also prefer to leave that for another PR.

functions.php
crstauf: There's no consistency on phpDoc blocks: some blocks [have periods](https://github.com/woocommerce/action-scheduler/blob/be0a27922b6ca34f5f5e924cb7265b78062a2918/functions.php#L10), [some do not](https://github.com/woocommerce/action-scheduler/blob/be0a27922b6ca34f5f5e924cb7265b78062a2918/functions.php#L22). 

If a decision is made on which should be used, do not mind making that happen.

crstauf:
@rrennick Responded to feedback, and pushed one commit.

functions.php
rrennick: > If a decision is made on which should be used, do not mind making that happen.

The current standard is ending period. But, the way PHPDocs formatting changes are addressed in any project I've worked on is to update a file when making other changes in that file.

crstauf: @rrennick I've added periods in cdfbd42.

rrennick (approved):
@crstauf Thanks for cleaning these up.

crstauf:
Welcome! 🙇

PR #476, opened 2020-02-28
Ability to hide "Tools > Scheduler Actions" page

### LINKED ISSUE #431: Ability to hide "Tools > Scheduler Actions" page
Not always a page "Scheduled Actions" is needed to be public in the admin area.

I would like to be able to hide it using a built-in filter in `\ActionScheduler::init()`, similar to what is done with deprecated functions (using `action_scheduler_load_deprecated_functions` filter).

Of course, I can hide it like this:
```php
add_action( 'admin_menu', function() {
	remove_submenu_page( 'tools.php', 'action-scheduler' );
}, PHP_INT_MAX );
```
with own conditions (when to actually hide it) in that closure, but a native filter, `action_scheduler_load_admin_view`, would be preferrable.

crstauf:
Closes #431.

### DIFF
--- a/classes/ActionScheduler_AdminView.php
+++ b/classes/ActionScheduler_AdminView.php
@@ -36,7 +36,9 @@ class ActionScheduler_AdminView extends ActionScheduler_AdminView_Deprecated {
 				add_filter( 'woocommerce_admin_status_tabs', array( $this, 'register_system_status_tab' ) );
 			}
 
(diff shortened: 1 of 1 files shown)


crstauf:
Can someone please re-run the build? Failure is from two timestamps not being equal (one second difference).

crstauf:
To whomever re-ran the build, thank you.

rrennick (approved):
@crstauf Looks great. Thanks for the PR.

PR #478, opened 2020-03-02
Update docs

### LINKED ISSUE #432: docs/Usage page sections order is incorrect
Steps at the top of the page:
> 1. installing the library
> 1. scheduling and action
> 1. attaching a callback to that action

But later in the doc, you start with Scheduling first, then Installation. 
And the 3rd step doesn't make sense, as there is no such section on a page at all.

So, imo, the text on a page should be re-arranged, and steps should be tweaked.

### LINKED ISSUE #438: Add async runner filter to docs
We are seeing some reports of deadlocks & 503s from the aync runner. We should add a section to the 3.0 page with instructions for using the `'action_scheduler_async_request_sleep_seconds'` filter to reduce the async runner load frequency or `'action_scheduler_allow_async_request_runner'` to disable it completely.

rrennick:
Update documentation in the docs folder:

- Remove `Default: None` on required parameters.
- Added `;` where appropriate in sample code.
- Fixed parameter type for group.
- Remove extra back ticks in api docs.
- Add the CLI `migrate` command.
- Other wording fixes.

Fixes #438 
Fixes #432

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -9,37 +9,6 @@ Using Action Scheduler requires:
 1. scheduling an action
 1. attaching a callback to that action
 
-## Scheduling an Action
-
-To schedule an action, call the [API function](/api/) for the desired schedule type passing in the required parameters.
-
-The example code below shows everything needed to schedule a function to run at midnight, if it's not already scheduled:
-
-```php
-require_once( plugin_dir_path( __FILE__ ) . '/libraries/action-scheduler/action-scheduler.php' );
-
-/**
- * Schedule an action with the hook 'eg_midnight_log' to run at midnight each day
- * so that our callback is run then.
- */
-function eg_schedule_midnight_log() {
-	if ( false === as_next_scheduled_action( 'eg_midnight_log' ) ) {
-		as_schedule_recurring_action( strtotime( 'midnight tonight' ), DAY_IN_SECONDS, 'eg_midnight_log' );
-	}
-}
-add_action( 'init', 'eg_schedule_midnight_log' );
-
-/**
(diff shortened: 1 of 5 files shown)


docs/version3-0.md
```diff
@@ -12,14 +12,30 @@ Yes! The Action Scheduler Custom Tables plugin code is now part of Action Schedu
 
 By default, Action Scheduler will only initiate a migration from the internal `WPPostStore` data store. To enable migration from any custom datastore add the following filter `add_filter( 'action_scheduler_migrate_data_store', '__return_true' );`.
 
-### I'm currently on PHP <5.5. When I update PHP will the migration start automatically?
+### I'm currently on PHP >=5.5. When I update PHP will the migration start automatically?
```
slaFFik: This change is imo incorrect.
The initial question I raised:
> When I'm on PHP 5.4, what will happen when I will upgrade PHP to a newer version (5.5+)?

Right now `I'm currently on PHP >=5.5.` doesn't make sense, because that's the default behavior of AS lib to start migration on PHP 5.5, because that's the requirement for custom tables.

docs/usage.md
```diff
@@ -9,37 +9,6 @@ Using Action Scheduler requires:
 1. scheduling an action
 1. attaching a callback to that action
```
slaFFik: I assume this 3rd ToC item (`attaching a callback to that action`) can be removed, as it's covered in "Scheduling an Action" section.

rrennick: @slaFFik Thanks for the suggestion. PR updated.

PR #480, opened 2020-03-04
Doc clarification. Fixes #479

### LINKED ISSUE #479: Implementation queries / doc ommissions
Now that action scheduler has moved to custom tables, it's appropriate for a need I have in a public plugin. I'm currently porting my current implementation (based on https://github.com/deliciousbrains/wp-background-processing) to Action Scheduler, but I have a few implementation questions that I can't find answers to in the docs. 

If someone can quickly answer these, happy to pull a docs PR together. 

1. The docs mention that action scheduler initialises itself on init/1, but it doesn't indicate at what point in the WordPress hook lifecycle job events are fired. It would be helpful to understand what point all of my job handlers need to be hooked. Currently I'm doing it on init/2 but it might be helpful if I could do them later as it would clean up a few things in my codebase.

2. [Somewhat related] Clarify what happens if a task is due to run, but there is nothing attached to the relevant hook - does the job fail? will it re-try?

Thanks

leewillis77:
Clarifies the point in the request lifecycle that action callbacks need to have been registered.

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -40,6 +40,8 @@ add_action( 'eg_midnight_log', 'eg_log_action_data' );
 
 For more details on all available API functions, and the data they accept, refer to the [API Reference](/api/).
 
+_Note:_ You should ensure that your callbacks are attached to the relevant hook before priority 10 of WordPress' `init` hook is fired. 
+
 ## Installation
 
 There are two ways to install Action Scheduler:
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -40,7 +40,7 @@ add_action( 'eg_midnight_log', 'eg_log_action_data' );
 
 For more details on all available API functions, and the data they accept, refer to the [API Reference](/api/).
 
-_Note:_ You should ensure that your callbacks are attached to the relevant hook before priority 10 of WordPress' `init` hook is fired. 
+_Note:_ You should ensure that your callbacks are attached to the relevant hook before WordPress' `wp_loaded` hook is fired.
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


rrennick:
@leewillis77 Thanks for submitting this. WP CRON does not run inside a hook so any priority can be used. Cron runs after `wp_loaded`.

leewillis77:
Hi; 

The issue is that developers using this library need to know the *latest* point at which they can register their callbacks and have them reliably called. Clearly there is a point in the lifecycle where it is "too late".  

From the discussion I had on #479 I understood that that was init (priority 9) as confirmed by Brent, but you're now saying that's not correct? Happy to update the PR with a revision, but need to understand your thoughts on what that point is. 

From your response, and a bit of experimentation, it seems that callbacks can be registered up until `parse_request` - is that correct?

rrennick:
> but you're now saying that's not correct?

Yes, the last hook in [wp-setting.php](https://github.com/WordPress/WordPress/blob/master/wp-settings.php#L562) is `wp_loaded`. Cron runs after that so any priority is okay.

rrennick:
> From your response, and a bit of experimentation, it seems that callbacks can be registered up until parse_request - is that correct?

Actions can be run multiple ways (WP CLI, cron, dashboard, and async). `wp_loaded` is a safe hook to use in all those cases.

leewillis77:
Updated the PR accordingly. Thanks

thenbrent:
> From the discussion I had on #479 I understood that that was init (priority 9) as confirmed by Brent

When `ALTERNATE_WP_CRON` is defined, WordPress runs WP-Cron on the `'init'` hook with the default priority of `10`. That means anything hooked to `'init'` with priority `10` or higher priority will not run for alternative WP-Cron requests, including scheduled actions (even though they will run for normal WP-Cron requests).

This is a known issue. Reported in trac here: https://core.trac.wordpress.org/ticket/24160

leewillis77:
Reverted to back to suggesting init (pre-10) as per Brent's comment - thanks!

PR #487, opened 2020-03-12
restore action handling to WC status, use data store to delete actions

rrennick:
Closes #485 

This PR restores the action handling to the Scheduled Actions screen under WooCommerce -> Status -> Scheduled Actions. The handling was removed in #354. 

While investigating this I noticed that the bulk delete action attempts to delete the actions from the default data store. The PR updates that function to use the data store to delete the actions.

### Testing

1. In master confirm that neither the bulk delete action or the action links work in the WC status page
1. Switch to this branch
1. Update `action-scheduler.php` to artificially bump the version
```
-if ( ! function_exists( 'action_scheduler_register_3_dot_1_dot_2' ) ) {
+if ( ! function_exists( 'action_scheduler_register_3_dot_1_dot_3' ) ) {
 
        if ( ! class_exists( 'ActionScheduler_Versions' ) ) {
                require_once( 'classes/ActionScheduler_Versions.php' );
                add_action( 'plugins_loaded', array( 'ActionScheduler_Versions', 'initialize_latest_version' ), 1, 0 );
        }
 
-       add_action( 'plugins_loaded', 'action_scheduler_register_3_dot_1_dot_2', 0, 0 );
+       add_action( 'plugins_loaded', 'action_scheduler_register_3_dot_1_dot_3', 0, 0 );
 
-       function action_scheduler_register_3_dot_1_dot_2() {
+       function action_scheduler_register_3_dot_1_dot_3() {
                $versions = ActionScheduler_Versions::instance();
-               $versions->register( '3.1.2', 'action_scheduler_initialize_3_dot_1_dot_2' );
+               $versions->register( '3.1.3', 'action_scheduler_initialize_3_dot_1_dot_3' );
        }
 
-       function action_scheduler_initialize_3_dot_1_dot_2() {
+       function action_scheduler_register_3_dot_1_dot_3() {
                $versions = ActionScheduler_Versions::instance();
-               $versions->register( '3.1.2', 'action_scheduler_initialize_3_dot_1_dot_2' );
+               $versions->register( '3.1.3', 'action_scheduler_initialize_3_dot_1_dot_3' );
        }
 
-       function action_scheduler_initialize_3_dot_1_dot_2() {
+       function action_scheduler_initialize_3_dot_1_dot_3() {
                require_once( 'classes/abstracts/ActionScheduler.php' );
                ActionScheduler::init( __FILE__ );
        }
 
        // Support usage in themes - load this version if no plugin has loaded a version yet.
[...]

PR #488, opened 2020-03-12
fix initialization order on activation

### LINKED ISSUE #484: Init order
Not an issue, just a question as I was looking at the code today and this comment 

https://github.com/woocommerce/action-scheduler/blob/59dbd87de0c8a826b08142fd122279409dffadf7/classes/abstracts/ActionScheduler.php#L153

got me thinking if the order of init functions is correct?
https://github.com/woocommerce/action-scheduler/blob/59dbd87de0c8a826b08142fd122279409dffadf7/classes/abstracts/ActionScheduler.php#L157-L160

rrennick:
Closes #484 

This PR fixes the initialization order in the block of code that runs the initialization after `init` (effectively during activation) to be consistent with the mainline initialization order.

I did some code clean up to address spacing inconsistencies and improve the code readability.

Test by verifying that the plugin still functions normally: Admin screens, actions run, etc.

### DIFF
--- a/classes/abstracts/ActionScheduler.php
+++ b/classes/abstracts/ActionScheduler.php
@@ -137,31 +137,29 @@ abstract class ActionScheduler {
 		 */
 		do_action( 'action_scheduler_pre_init' );
 
-		require_once( self::plugin_path('functions.php') );
+		require_once( self::plugin_path( 'functions.php' ) );
 		ActionScheduler_DataController::init();
 
-		$store = self::store();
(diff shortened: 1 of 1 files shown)


thenbrent:
This should be tested with #487 applied too, because that PR updates bulk deletion to use `ActionScheduler::store()`. `ActionScheduler::store()->init()` will need to have been called before that's usable.

rrennick:
> This should be tested with #487 applied too

This tested fine in both Tools -> Scheduled Actions and WooCommerce -> Status -> Scheduled Actions.

PR #492, opened 2020-03-13
add table missing notice with action link

### LINKED ISSUE #491: Add dashboard method to recreate tables
There are enough reports of AS tables being missing (including being deleted by the site owner via the WP Optimize plugin https://wordpress.org/support/topic/can-not-update-my-database/#post-12534740) that we should add a method to recreate the tables via the dashboard.

WooCommerce Reference: https://github.com/woocommerce/woocommerce/issues/25895

rrennick:
Closes #491 

As reported in the issue, [WP Optimize](https://en-ca.wordpress.org/plugins/wp-optimize/) is prompting some users to delete the action scheduler tables. In #455, we bumped the schema version to ensure that tables would be created when updating to WooCommerce 4.0 if they had previously been created and deleted.

This PR adds a check to verify the AS tables exist to the Scheduled Actions screen. The check is conditional on the site using either the hybrid or internal table data store. If any of the 4 tables are missing. A dashboard notice is added with an action link to create the tables.

### Testing

- Ensure the migration is completed
- Rename `wp_actionscheduler_actions`
- Go to Tools -> Scheduled Actions
- Verify there are no actions listed, the missing tables message appears 
- Deactivate AS, activate WC 3.9.3
- Create an action with `as_schedule_single_action()`
- Drop `wp_actionscheduler_actions`
- Deactivate WC, activate AS
- Go to Tools -> Scheduled Actions same message should appear
- Action created above should be listed
- Restore original `wp_actionscheduler_actions`

### DIFF
--- a/classes/ActionScheduler_ListTable.php
+++ b/classes/ActionScheduler_ListTable.php
@@ -512,8 +512,8 @@ class ActionScheduler_ListTable extends ActionScheduler_Abstract_ListTable {
 
 		$store_schema  = new ActionScheduler_StoreSchema();
 		$logger_schema = new ActionScheduler_StoreSchema();
-		$store_schema->register_tables( 'yes' );
-		$logger_schema->register_tables( 'yes' );
+		$store_schema->register_tables( true );
+		$logger_schema->register_tables( true );
 
 		remove_action( 'action_scheduler/created_table', array( $store, 'set_autoincrement' ), 10 );
 	}
--- a/classes/ActionScheduler_ListTable.php
+++ b/classes/ActionScheduler_ListTable.php
@@ -325,7 +325,7 @@ class ActionScheduler_ListTable extends ActionScheduler_Abstract_ListTable {
 	public function display_admin_notices() {
 		global $wpdb;
 
-		if ( is_a( $this->store, 'ActionScheduler_HybridStore' ) || is_a( $this->store, 'ActionScheduler_DBStore' ) ) {
+		if ( ( is_a( $this->store, 'ActionScheduler_HybridStore' ) || is_a( $this->store, 'ActionScheduler_DBStore' ) ) && apply_filters( 'action_scheduler_enable_recreate_data_store', true ) ) {
 			$table_list = array(
 				'actionscheduler_actions',
 				'actionscheduler_logs',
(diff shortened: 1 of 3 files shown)


classes/ActionScheduler_ListTable.php
```diff
@@ -318,10 +318,43 @@ protected function maybe_render_actions( $row, $column_name ) {
...
+			$table_list = array(
+				'actionscheduler_actions' => 'action_id',
+				'actionscheduler_logs' => 'action_id',
+				'actionscheduler_groups' => 'group_id',
+				'actionscheduler_claims' => 'claim_id',
+			);
+
+			$last_error = $wpdb->last_error;
+			foreach ( $table_list as $table_name => $field_name ) {
+				$id = $wpdb->get_var( "SELECT {$field_name} FROM {$wpdb->prefix}{$table_name} LIMIT 1" ); // phpcs:ignore WordPress.DB.PreparedSQL.InterpolatedNotPrepared
```
jeffstieler: Rather than deliberately causing an SQL error, can we use `SHOW TABLES LIKE` instead?

rrennick: @jeffstieler Thanks for the suggestion. I was thinking of performance but since this only occurs in the dashboard on two low use screens, cleaner code is better for maintainability. This is ready for another review.

rrennick:
@jeffstieler I implemented the feedback from https://github.com/woocommerce/action-scheduler/issues/491#issuecomment-599317614 for review.

classes/ActionScheduler_ListTable.php
```diff
@@ -473,6 +499,24 @@ protected function row_action_run( $action_id ) {
...
+			$store = $this->store;
+		} else {
+			$store = new ActionScheduler_HybridStore();
+		}
+		add_action( 'action_scheduler/created_table', array( $store, 'set_autoincrement' ), 10, 2 );
+
+		$store_schema  = new ActionScheduler_StoreSchema();
+		$logger_schema = new ActionScheduler_StoreSchema();
+		$store_schema->register_tables( 'yes' );
+		$logger_schema->register_tables( 'yes' );
```
jeffstieler: Nitpick: can we make this argument easier to grok from reading the call, or at least switch this to using a boolean?

See: [Self-Explanatory Flag Values for Function Arguments](https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#self-explanatory-flag-values-for-function-arguments)

rrennick: I update the instructions & flag type.

patrikalienus:
I'm trying to follow the above bullet list of actions to take as updating from 3.9.3 -> 4.0.1 created the same issue as updating to 4.0 would have. The above list doesn't do a great job at explaining though. "Create an action with as_schedule_single_action()" doesn't say much - create what action, for instance?

rrennick:
@patrikalienus The instructions in the description were for testing this pull request. You do not need to take any additional steps to update to WC 4.0.1. When updating, WooCommerce runs all of its update routines between the original version and updated version.

patrikalienus:
@rrennick Well that's strange, because I'm running into the exact same issues when updating to 4.0.1...

rrennick:
> I'm running into the exact same issues when updating to 4.0.1...

This PR was to fix an issue created by another plugin vs updating WooCommerce. If you are having an issue where tables aren't being created then you'll need to look into your error logs.

LGx0:
wp_action_error
wp_woo_error

jhlouw86:
My issue got resolved but following these steps (Slava Abakumov):

1. Open /wp-admin/options.php page on your site.
2. On this page find options schema-ActionScheduler_LoggerSchema, schema-ActionScheduler_StoreSchema, and action_scheduler_migration_status and remove their values (leave them empty), save options changes.
3. Activate on the Plugins page the Woocommerce plugin again.

https://wordpress.org/support/topic/plugin-is-not-working-after-upgrade-please-fix-asap/

PR #503, opened 2020-03-30
add date values to set demarkation query

### LINKED ISSUE #437: Demarkation ID not set in actionscheduler_actions

rrennick:
Closes #437 

This PR adds date strings to the four date fields `actionscheduler_actions` to the set demarkation ID query. 

In the WP schema, `0000-00-00 00:00:00` is the default date in the `posts` table but WP doesn't write any records to the table relying on the default value. So, the MySQL configuration outlined in #437 wouldn't be detected in a default WP install.

### Testing

- Update your `sql_mode` to include `STRICT_TRANS_TABLES,NO_ZERO_DATE` and restart MySQL 5.7
- Test the configuration with 
```mysql> insert into wp_actionscheduler_actions (hook,status) values ('test_zero_date','completed');```
- MySQL is configured correctly when the query returns an invalid date error

- Rename `wp_actionscheduler_actions`
- Visit Tools -> Scheduled Actions
- Note the value of the `action_scheduler_hybrid_store_demarkation` option
- Check the IDENTITY value of the `wp_actionscheduler_actions` to ensure that it is at least 1 greater than the demarkation value
- restore your original `wp_actionscheduler_actions` table
- Test that async actions can be enqueued with 
```
	as_enqueue_async_action( 'as-async-test' );
} );
```
- Check debug log for warnings/errors & pending action is created

### DIFF
--- a/classes/data-stores/ActionScheduler_HybridStore.php
+++ b/classes/data-stores/ActionScheduler_HybridStore.php
@@ -77,18 +77,33 @@ class ActionScheduler_HybridStore extends Store {
 			}
 			/** @var \wpdb $wpdb */
 			global $wpdb;
-			$wpdb->insert(
+			/**
+			 * A default date of '0000-00-00 00:00:00' is invalid in MySQL 5.7 when configured with 
+			 * sql_mode including both STRICT_TRANS_TABLES and NO_ZERO_DATE.
+			 */
+			$default_date = new DateTime( 'tomorrow' );
+			$null_action  = new ActionScheduler_NullAction();
+			$date_gmt     = $this->get_scheduled_date_string( $null_action, $default_date );
+			$date_local   = $this->get_scheduled_date_string_local( $null_action, $default_date );
+
+			$row_count = $wpdb->insert(
 				$wpdb->{ActionScheduler_StoreSchema::ACTIONS_TABLE},
 				[
-					'action_id' => $this->demarkation_id,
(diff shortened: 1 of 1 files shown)


thenbrent:
Nice find & fix @rrennick. I'll leave this up for others to review, but one important point:

> WP doesn't write any records to the table relying on the default value.

I'm pretty sure I've seen `0000-00-00 00:00:00` in the posts table before. Maybe for `auto-draft` posts?

The reason that matters, is because while this patch is a good solution for the demarkation ID, having `STRICT_TRANS_TABLES`, `NO_ZERO_DATE` we cause problems after that still, because we use `0000-00-00 00:00:00` for async actions' scheduled date (to make sure they run before anything else).

rrennick:
> because we use 0000-00-00 00:00:00 for async actions' scheduled date

@thenbrent This is okay. The error only occurs if the query relies on the MySQL default value of the field and the default is `0000-00-00 00:00:00`. I've updated the PR testing instructions to verify that async actions can still be enqueued.

@jeffstieler This is ready for review.

PR #509, opened 2020-04-01
Warn the user whenever they have past-due actions

### LINKED ISSUE #91: Warn the user whenever they have past-due actions
While I was doing support I had this idea. I believe it is fairly easy for Action-Scheduler to warn the user whenever they have past-due actions. It can even wait for a threshold (like a day or two).

It be built on top of the [admin notices API](https://codex.wordpress.org/Plugin_API/Action_Reference/admin_notices) and have some link to our documentation which explains what may be wrong and how to fix it. 

It should also give the option to run the past-due actions manually. Although these things are possible (detect there are past-due actions, and run actions manually) I believe it is worth to make them obvious and super simple to diagnose problems, offer a temporarily (run the actions now) and a permanent solution (run the crons)

crstauf:
Closes #91.

- [ ] Ping `thenbrent` when merged ([ref](https://github.com/woocommerce/action-scheduler/pull/509#discussion_r434306702)).

### DIFF
--- a/classes/ActionScheduler_AdminView.php
+++ b/classes/ActionScheduler_AdminView.php
@@ -120,7 +120,7 @@ class ActionScheduler_AdminView extends ActionScheduler_AdminView_Deprecated {
 	public function maybe_check_pastdue_actions() {
 
 		# Filter to prevent checking actions (ex: inappropriate user).
-		if ( !apply_filters( 'action_scheduler_check_pastdue_actions', true ) )
+		if ( ! apply_filters( 'action_scheduler_check_pastdue_actions', true ) )
 			return;
 
 		# Get last check transient.
@@ -166,7 +166,7 @@ class ActionScheduler_AdminView extends ActionScheduler_AdminView_Deprecated {
 		}
 
 		# If check failed, set transient and abort.
-		if ( !boolval( $check ) ) {
+		if ( ! boolval( $check ) ) {
 			$interval = apply_filters( 'action_scheduler_pastdue_actions_check_interval', round( $threshold_seconds / 4 ), $threshold_seconds );
 			set_transient( 'action_scheduler_last_pastdue_actions_check', time(), $interval );
 
--- a/classes/ActionScheduler_AdminView.php
+++ b/classes/ActionScheduler_AdminView.php
@@ -137,6 +137,7 @@ class ActionScheduler_AdminView extends ActionScheduler_AdminView_Deprecated {
 	/**
 	 * Check past-due actions, and print notice.
 	 *
+	 * @todo update $link_url to "Past-due" filter when released (see issue #510, PR #511)
 	 */
 	protected function check_pastdue_actions() {
 
@@ -172,8 +173,14 @@ class ActionScheduler_AdminView extends ActionScheduler_AdminView_Deprecated {
 			return;
 		}
 
+		$link_url = add_query_arg( array(
+			'page'   => 'action-scheduler',
+			'status' => 'pending',
+			'order'  => 'asc',
+		), admin_url( 'tools.php' ) );
+
 		# Print notice.
-		printf( __( '<div class="%s"><p>%d past-due actions found; something may be wrong.</p></div>', 'action-scheduler' ), 'notice notice-warning', $num_pastdue_actions );
+		printf( __( '<div class="%s"><p><strong>Action Scheduler:</strong> %d past-due actions found; something may be wrong. <a href="%s">View pending actions &raquo;</a></p></div>', 'action-scheduler' ), 'notice notice-warning', $num_pastdue_actions, esc_url( $link_url ) );
 
 		# Facilitate third-parties to evaluate and print notices.
 		do_action( 'action_scheduler_pastdue_actions_extra_notices', $query_args );
--- a/classes/ActionScheduler_AdminView.php
+++ b/classes/ActionScheduler_AdminView.php
@@ -137,7 +137,6 @@ class ActionScheduler_AdminView extends ActionScheduler_AdminView_Deprecated {
 	/**
 	 * Check past-due actions, and print notice.
 	 *
-	 * @todo set alert message
 	 */
 	protected function check_pastdue_actions() {
 
--- a/classes/ActionScheduler_AdminView.php
+++ b/classes/ActionScheduler_AdminView.php
@@ -40,7 +40,7 @@ class ActionScheduler_AdminView extends ActionScheduler_AdminView_Deprecated {
 			}
 
 			add_action( 'admin_menu', array( $this, 'register_menu' ) );
-
+			add_action( 'admin_notices', array( $this, 'maybe_check_past_actions' ) );
 			add_action( 'current_screen', array( $this, 'add_help_tabs' ) );
 		}
 	}
@@ -108,6 +108,65 @@ class ActionScheduler_AdminView extends ActionScheduler_AdminView_Deprecated {
 		}
 
 		return $this->list_table;
+	}
+
+	/**
+	 * Maybe check past actions, and display alert.
+	 *
+	 * @todo break into pieces
+	 * @todo test
+	 * @todo add alert message
+	 * @todo confirm and set seconds past
+	 * @todo confirm and set min past actions count
+	 */
+	public function maybe_check_past_actions() {
+
+		# Filter to prevent checking actions (ex: inappropriate user).
+		if ( apply_filters( 'action_scheduler_check_past_actions', false ) )
+			return;
+
+		# Use transient to prevent checking on every admin page load.
+		$transient_key = 'action_scheduler_last_past_actions_check';
+		$last_check = get_transient( $transient_key );
+
+		# If transient exists, we're inside of the interval, so abort.
+		if ( !empty( $last_check ) )
+			return;
+
+		# Set thresholds.
+		$past_seconds = ( int ) apply_filters( 'action_scheduler_past_actions_seconds', DAY_IN_SECONDS );
+		$min_past_actions = ( int ) apply_filters( 'action_scheduler_alert_past_actions_count', 20 );
+
+		# Allow third-parties to preempt the default check logic.
+		$check = apply_filters( 'action_scheduler_past_action_check_pre', null );
+
+		$query_args = array(
+			'status'   => ActionScheduler_Store::STATUS_PENDING,
+			'per_page' => $min_past_actions,
+			'date'     => time() - $past_seconds,
+		);
+
+		# If no third-party preempted, run default check.
+		if ( is_null( $check ) ) {
+			$num_past_actions = $store->query_actions( $query_args, 'count' );
+			$check = apply_filters( 'action_scheduler_past_actions_check', ( $num_past_actions < $min_past_actions ), $num_past_actions, $past_seconds, $min_past_actions );
+
+			# Set transient to only check on designated interval.
+			$interval = ( int ) apply_filters( 'action_scheduler_check_past_actions_interval', round( $past_seconds / 4 ), $past_seconds );
+			set_transient( $transient_key, 1, $interval );
+		}
+
+		$check = ( bool ) $check;
+
+		# If check failed, do not display notice.
+		if ( !$check )
+			return;
+
+		# Print notice.
+		printf( __( '<div class="%s"><p>%s</p></div>', 'notice notice-warning', '{{ ALERT MESSAGE }}', 'action-scheduler' ) );
+
+		# Facilitate third-parties to evaluate and print notices.
+		do_action( 'action_scheduler_past_actions_notices', $query_args );
 	}
 
 	/**
--- a/classes/ActionScheduler_AdminView.php
+++ b/classes/ActionScheduler_AdminView.php
@@ -118,6 +118,7 @@ class ActionScheduler_AdminView extends ActionScheduler_AdminView_Deprecated {
 	 * @todo add alert message
 	 * @todo confirm and set seconds past
 	 * @todo confirm and set min past actions count
+	 * @todo add error_log() call
 	 */
 	public function maybe_check_past_actions() {
 
--- a/classes/ActionScheduler_AdminView.php
+++ b/classes/ActionScheduler_AdminView.php
@@ -148,9 +148,11 @@ class ActionScheduler_AdminView extends ActionScheduler_AdminView_Deprecated {
 		# Allow third-parties to preempt the default check logic.
 		$check = apply_filters( 'action_scheduler_pastdue_actions_check_pre', null );
 
+		# Scheduled actions query arguments.
 		$query_args = array(
-			'status' => ActionScheduler_Store::STATUS_PENDING,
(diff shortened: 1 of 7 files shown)


rrennick:
@crstauf Sorry, I should have left a comment here. The `In-Progress` admin notice has a link to the `In-Progress` status filter. My suggestion in Slack was that it might be worth doing the same in this PR. I'm going to close #510 as a duplicate.

crstauf:
> the `In-Progress` status filter

@rrennick If I'm understanding "In-Progress" correctly, this would only be actions that are in the current batch, not all past-due actions. Possible a view for "Past-due" actions could be helpful?

> it might be worth doing the same in this PR

Can you please describe further what might be worth doing in this PR? Linking to the "In-Progress" view?

rrennick:
> If I'm understanding "In-Progress" correctly

If you load the Scheduled Actions screen while there are actions in progress by one of the queue runners there is a notice added https://github.com/woocommerce/action-scheduler/blob/master/classes/ActionScheduler_ListTable.php#L374

crstauf:
@rrennick Okay, that's exactly what I was thinking: do you mind re-opening #510? I'd prefer to keep them separate.

crstauf:
Below are the defined settings; please review and let me know if any changes needed:

- Threshold for past-due action if scheduled time before: `DAY_IN_SECONDS`
- Threshold for minimum number of past-due actions: `20`
- Check interval: One quarter of threshold seconds (ex: `DAY_IN_SECONDS / 4`)

thenbrent:
This is great @crstauf thanks for the PR!

> Threshold for minimum number of past-due actions: 20

I don't think we want a threshold on number of actions. Even small sites can have WP Cron problems, and if they have 10 scheduled payments not being processed, it's a big problem.

The other suggestion I have is to either:

* constrain this notice to only show in Action Scheduler pages (and let plugins handle notices for other pages); or
* add a new doc on actionscheduler.org, explaining what this past due action notice means, and ways to fix it.

Action Scheduler is a developer tool used in plugins. Developers know what it is. Site owners are less likely to know what it is. That means most site owners won't have any idea what "past-due actions found" means.

crstauf:
> I don't think we want a threshold on number of actions

@thenbrent Is it okay if I set the minimum to one, and leave it filterable?

> constrain this notice to only show in Action Scheduler pages

With #510 a "Past-due" view will be added to the list table, so having the alert only show on that page would be redundant: having the notice show throughout the admin, or as dictated by the `action_scheduler_check_pastdue_actions` filter, is ideal.

> add a new doc on actionscheduler.org

I can get this started, but the full language will need to be provided or added by someone with more technical support knowledge.

thenbrent:
> Is it okay if I set the minimum to one, and leave it filterable?

That sounds like the ideal approach, as some sites may prefer a higher threshold. 👍 

> I can get this started, but the full language will need to be provided or added by someone with more technical support knowledge than me.

Sounds good. Once the PR is reviewed / merged in its final form, I can write that up and include a link in the alert.

crstauf:
@rrennick @thenbrent This PR is ready for review.

classes/ActionScheduler_AdminView.php
```diff
@@ -110,6 +110,89 @@ protected function get_list_table() {
...
+	 * Action: admin_notices
+	 *
+	 * Maybe check past-due actions, and print notice.
+	 *
+	 * @uses $this->check_pastdue_actions()
+	 */
+	public function maybe_check_pastdue_actions() {
+
+		# Filter to prevent checking actions (ex: inappropriate user).
+		if ( !apply_filters( 'action_scheduler_check_pastdue_actions', true ) )
```
rrennick, line 123: We follow the WP coding standard of `if ( ! apply....`. There's a second one of these below.

classes/ActionScheduler_AdminView.php
```diff
@@ -110,6 +110,89 @@ protected function get_list_table() {
...
+			$store = ActionScheduler_Store::instance();
+			$num_pastdue_actions = ( int ) $store->query_actions( $query_args, 'count' );
+
+			# Check if past-due actions count is greater than or equal to threshold.
+			$check = ( $num_pastdue_actions >= $threshhold_min );
+			$check = ( bool ) apply_filters( 'action_scheduler_pastdue_actions_check', $check, $num_pastdue_actions, $threshold_seconds, $threshhold_min );
+		}
+
+		# If check failed, set transient and abort.
+		if ( !boolval( $check ) ) {
```
rrennick, line 169: Another `!` without the spacing here.

docs/faq.md
```diff
@@ -105,3 +105,9 @@ It requires no setup, and won't override any WordPress APIs (unless you want it
 Action Scheduler is designed to manage the scheduled actions on a single site. It has no special handling for running queues across multiple sites in a multisite network. That said, because its storage and Queue Runner are completely swappable, it would be possible to write multisite handling classes to use with it.
 
 If you'd like to create a multisite plugin to do this and release it publicly to help others, [open a new issue to let us know](https://github.com/woocommerce/action-scheduler/issues/new), we'd love to help you with it.
+
+### My site has past-due actions; what can I do?
+
+Actions that are past-due have missed their scheduled date; because of [how WP Cron works](https://developer.wordpress.org/plugins/cron/), it is normal to have some past-due actions. 
+
+If there are several past-due actions more than one day old, there may be something wrong with your site. If you need help determining the issue, please [contact us](https://woocommerce.com/contact-us/).
```
rrennick, line 113: Can you add something along the lines of `and are a WooCommerce.com customer`?  There are quite a few non-WC plugins that use AS. Those users won't be able to get support via our contact form.

classes/ActionScheduler_AdminView.php
crstauf: @rrennick My apologies; fixed in b3e80b5.

classes/ActionScheduler_AdminView.php
crstauf: Fixed in b3e80b5.

docs/faq.md
crstauf: @rrennick Actually, @thenbrent indicated he'd provide the final write-up here: https://github.com/woocommerce/action-scheduler/pull/509#issuecomment-608184598.

I can still make the change if desired, just noting Brent's intent.

thenbrent: I'm happy to write that up still. Please ping me once this is merged. :)

classes/ActionScheduler_AdminView.php
```diff
@@ -110,6 +110,89 @@ protected function get_list_table() {
...
+
+			# Check if past-due actions count is greater than or equal to threshold.
+			$check = ( $num_pastdue_actions >= $threshhold_min );
+			$check = ( bool ) apply_filters( 'action_scheduler_pastdue_actions_check', $check, $num_pastdue_actions, $threshold_seconds, $threshhold_min );
+		}
+
+		# If check failed, set transient and abort.
+		if ( ! boolval( $check ) ) {
+			$interval = apply_filters( 'action_scheduler_pastdue_actions_check_interval', round( $threshold_seconds / 4 ), $threshold_seconds );
+			set_transient( 'action_scheduler_last_pastdue_actions_check', time(), $interval );
```
rrennick, line 171: @crstauf Is this transient used anywhere?

crstauf: @rrennick Yes, here:
https://github.com/woocommerce/action-scheduler/blob/8df980217252f37e510c3051068951de27c131d0/classes/ActionScheduler_AdminView.php#L120-L135

crstauf: Although it occurs to me now that if a site is using object caching, a transient may not survive the eviction algorithm. Thoughts?

crstauf: @rrennick 

> Although it occurs to me now that if a site is using object caching, a transient may not survive the eviction algorithm. Thoughts?

rrennick: > Although it occurs to me now that if a site is using object caching, a transient may not survive the eviction algorithm. Thoughts?

I'd expect that possibility applies to any transient. I haven't seen any code that attempts to mitigate it elsewhere.

crstauf: > haven't seen any code that attempts to mitigate it elsewhere

@rrennick Simplest fix is to use an option instead of a transient. Thoughts?

rrennick: @crstauf I would expect that if eviction were a significant problem then developers would use them as little as possible. If a site had an issue with eviction then it would probably manifest itself in multiple places. 

The advantage of transients is that in a non-persistent cache scenario they are only loaded when needed. In 99% of installs this transient would rarely be accessed.

crstauf: @rrennick Agreed. I only know this because I previously managed an eCommerce site that had issues with object cache evictions.

Is there any opposition against using an option that is not autoloaded?

rrennick: > Is there any opposition against using an option that is not autoloaded?

A transient is an option that's not autoloaded and WP will load a transient from the DB if it's not in cache. I would leave it as a transient since the expiry can be customized.

crstauf: > leave it as a transient

Okay; resolving this conversation.

docs/faq.md
crstauf: Added todo item to PR; resolving this conversation.

rrennick:
> constrain this notice to only show in Action Scheduler pages (and let plugins handle notices for other pages)

I appreciate that this was discussed above. More than half of all existing WooCommerce installs are still running WC 3.X.  With the notice being universal in the dashboard, all of those sites will see this notice when they update to WC 4.X (whichever version the were added to). If they were an established store with a significant amount of data they will see that number rise as the historical import generates new actions over the import process. One of the test imports I ran created 7000+ actions.

Can we add a filter to conditionally disable the notice? That would allow plugins using AS to filter where the notice appears.

crstauf:
> add a filter to conditionally display the notice

@rrennick `action_scheduler_show_pastdue_actions_notice` introduced in 92c92a8.

classes/ActionScheduler_AdminView.php
```diff
@@ -112,6 +113,94 @@ protected function get_list_table() {
...
+
+			return;
+		}
+
+		# Allow third-parties to conditionally display notice(s).
+		if ( ! apply_filters( 'action_scheduler_show_pastdue_actions_notice', true ) ) {
```
rrennick, line 180: @crstauf This filter will work great. It can be moved to the beginning of the function as there is no need to check for past actions if the notice isn't going to be shown.

classes/ActionScheduler_AdminView.php
```diff
@@ -112,6 +113,94 @@ protected function get_list_table() {
...
+
+		# Scheduled actions query arguments.
+		$query_args = array(
+			'date'     => as_get_datetime_object( time() - $threshold_seconds ),
+			'status'   => ActionScheduler_Store::STATUS_PENDING,
+			'per_page' => $threshhold_min,
+		);
+
+		# If no third-party preempted, run default check.
+		if ( is_null( $check ) ) {
```
rrennick, line 162: This condition could be changed to another short circuit return and moved to before assigning `$query_args`.

rrennick (changes requested):
@crstauf This looks good and tested fine for me. It will be ready for merge with the two changes I noted.

classes/ActionScheduler_AdminView.php
crstauf: @rrennick I thought about that, but then considered that perhaps a third-party may want `action_scheduler_pastdue_actions_check_pre` to run, and it be logged elsewhere.

Thoughts?

classes/ActionScheduler_AdminView.php
crstauf: @rrennick If a third-party has hooked on to `action_scheduler_pastdue_actions_check_pre` and is using built-in notice, moving this check up would not allow the notice to print.

crstauf:
> two changes I noted.

@rrennick Thanks! Replied.

classes/ActionScheduler_AdminView.php
```diff
@@ -112,6 +113,94 @@ protected function get_list_table() {
...
+	 * Action: admin_notices
+	 *
+	 * Maybe check past-due actions, and print notice.
+	 *
+	 * @uses $this->check_pastdue_actions()
+	 */
+	public function maybe_check_pastdue_actions() {
+
+		# Filter to prevent checking actions (ex: inappropriate user).
+		if ( ! apply_filters( 'action_scheduler_check_pastdue_actions', true ) )
```
rrennick, line 126: Sorry, I just noticed. This should have a `current_user_can( 'manage_options` )` as only admins can access the SA screen.

classes/ActionScheduler_AdminView.php
rrennick: Looking above we can remove this filter and use the `action_scheduler_check_pastdue_actions` to limit where the notice is shown. Sorry for the misdirection.

classes/ActionScheduler_AdminView.php
rrennick: See comment above.

classes/ActionScheduler_AdminView.php
crstauf: Resolved in a17cf37.

classes/ActionScheduler_AdminView.php
crstauf: @rrennick It's not clear to me what you're wanting. Can you please explain in more detail?

rrennick: > Can we add a filter to conditionally disable the notice?

This is redundant. The `action_scheduler_check_pastdue_actions` filter can be used to disable the check and notice.

crstauf: @rrennick Ahh gotcha. Yeah, I thought it may be redundant, but then I thought you were differentiating between checking for past-due actions and displaying a notice.

Will remove the filter that disables the notice.

Nothing to do on `L162` then?

crstauf: `action_scheduler_show_pastdue_actions_notice` filter removed in 5d26cc3.

rrennick: > Nothing to do on L162 then?

The `$query_args` array is only used inside the `is_null()` check so it could be moved to that conditional block.

crstauf: @rrennick It’s used later as well, in the action for third-parties to print their own notices.

jorgeatorres:
This has been rebased to fix conflicts. Assigning to you @Konamiman, per p1658501004449729-slack-C0E1AV8T0.

Konamiman:
Hi @rrennick, this PR looks good to me, do you think it needs more changes?

rrennick:
@Konamiman Thanks for the ping. No objections here.

Konamiman:
Great. Merging then.

Konamiman (approved):

PR #511, opened 2020-04-01
Add a "Past-due" view for scheduled actions

### LINKED ISSUE #510: Add a "Past" view for pending scheduled actions
Add a view that shows only actions that are pending and have past their scheduled date. This would be especially helpful for the message shown in #91.

Suggested by @rrennick in Slack DM.

crstauf:
Closes #510.

- [ ] After #509 and this PR have been merged, add a link to the view on the past-due actions admin notice (intro'd in #509).

### Changelog

> Enhancement - Adds a new "Past Due" view to the scheduled actions list table.

### DIFF
--- a/classes/abstracts/ActionScheduler_Store.php
+++ b/classes/abstracts/ActionScheduler_Store.php
@@ -132,6 +132,27 @@ abstract class ActionScheduler_Store extends ActionScheduler_Store_Deprecated {
 	abstract public function action_counts();
 
 	/**
+	 * Get additional action counts.
+	 *
+	 * - add past-due actions
+	 *
+	 * @return array
+	 */
+	public function extra_action_counts() {
+		$extra_actions = array();
+
+		$pastdue_action_counts = ( int ) $this->query_actions( array(
+			'status' => self::STATUS_PENDING,
+			'date'   => as_get_datetime_object(),
+		), 'count' );
+
+		if ( 0 !== $pastdue_action_counts )
+			$extra_actions['past-due'] = $pastdue_action_counts;
+
+		return apply_filters( 'action_scheduler_extra_action_counts', $extra_actions );
+	}
+
+	/**
 	 * @param string $action_id
 	 */
 	abstract public function cancel_action( $action_id );
--- a/classes/abstracts/ActionScheduler_Store.php
+++ b/classes/abstracts/ActionScheduler_Store.php
@@ -146,8 +146,9 @@ abstract class ActionScheduler_Store extends ActionScheduler_Store_Deprecated {
 			'date'   => as_get_datetime_object(),
 		), 'count' );
 
-		if ( 0 !== $pastdue_action_counts )
+		if ( $pastdue_action_counts ) {
 			$extra_actions['past-due'] = $pastdue_action_counts;
+		}
 
 		return apply_filters( 'action_scheduler_extra_action_counts', $extra_actions );
 	}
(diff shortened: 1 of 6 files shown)


crstauf:
@rrennick Job #1772.21 failed because of a unix timestamp comparison failure; can you please trigger a rebuild?

rrennick:
@crstauf triggered & passed.

crstauf:
@rrennick How can the PR alert custom data stores to support the "Past-due" view?

In the PR, I've adjusted the built-in data stores, which means that custom data stores will not have the necessary code to support the "Past-due" view.

Suggestions?

crstauf:
@rrennick @thenbrent This PR is ready for review.

rrennick:
@crstauf Thanks again. On initial look on this looks good. I'll hold it until the past due notice is merged.

jorgeatorres:
Re-assigning to @barryhughes per p1659719396500269-slack-C0E1AV8T0.

classes/abstracts/ActionScheduler_Store.php
```diff
@@ -131,6 +131,28 @@ public function query_action( $query ) {
...
+		$pastdue_action_counts = ( int ) $this->query_actions( array(
+			'status' => self::STATUS_PENDING,
+			'date'   => as_get_datetime_object(),
+		), 'count' );
+
+		if ( $pastdue_action_counts ) {
+			$extra_actions['past-due'] = $pastdue_action_counts;
+		}
+
+		return apply_filters( 'action_scheduler_extra_action_counts', $extra_actions );
```
barryhughes, line 159: Add a docblock?

barryhughes (commented):
Observation: as currently defined, past-due actions also include *async* actions, and because async actions have a nominal scheduled date of something like `0000-00-00 00:00:00` they are also listed first (and while it's true they might legitimately be past-due, sometimes nothing will actually be wrong and they will simply be waiting on the next queue runner starting).

https://user-images.githubusercontent.com/3594411/183159488-633cc8d2-74a1-4f19-9529-500d11b90746.mov

☝🏽 This probably won't be too much of a problem, most of the time, but I wanted to surface it. It also means that occasionally the **Past-Due** view filter will be displayed when actually all we have are some async actions that are nothing to worry about:

past-due-view

Lastly, and though I know it's not technically new as of this PR, the past-due admin notice may appear for similar reasons (even if there are no past-due *scheduled* actions):

warning-past-due

Again, I realize some of the code pre-dates this PR, but I wanted to get some thoughts before merging.

classes/abstracts/ActionScheduler_Store.php
jorgeatorres: Done.

jorgeatorres:
Hi @barryhughes!

Thanks for the review! I've added the docblock per your request.

As for the async actions, I totally get what you mean. For those, "past due" means something different, but I'm not sure if there's an easy way to get the correct count using the query args that the AS datastore currently supports.
I guess we could get "creative": for instance, we could ignore async actions when checking whether there are any actions past due (maybe by subtracting actions with a scheduled date <= `0000-00-00 ...`). Would that make sense?

That's not ideal as async actions can be "past due" in a way that makes sense in this check but we don't have the information nor query args to support that so we could at least try to reduce noise.

Please let me know if you think of a better way to do this check or if I'm missing something. Thanks!

barryhughes:
Yep, I think you're right: unless we also want to make a schema change, there isn't really a clean and efficient way to handle this. I think we could ship this as it is (and we can make further enhancements in the future if feedback suggests it is needed). wdyt?

barryhughes (approved):

jorgeatorres:
>  I think we could ship this as it is (and we can make further enhancements in the future if feedback suggests it is needed). wdyt?

Sounds good to me. If necessary, we could then do something along the lines I described above or revisit [this issue](https://github.com/woocommerce/action-scheduler/issues/686) on adding a way to search for async actions.

PR #524, opened 2020-04-17
Fix default value for free-memory-on flag

samimakela:
The default value for the free-memory-on should be 50 as documented. Currently, if no value is given, no memory freeing attempts will be made. This might unintentionally cause problems with larger batches of e.g. Subscriptions payments.


rrennick:
@samimakela Thanks for reporting this and the pull request. I added a second commit to your branch to add validation to the two integer parameters and created #526. I'll close this PR as you commit is included in #526.

PR #538, opened 2020-04-28
Deprecate use of shutdown hook

### LINKED ISSUE #536: WP shutdown hook runs on fatal error
Action Scheduler registers the autoloader in `plugins_loaded` but doesn't initialize the data stores until `init`. If a fatal error occurs between those two hooks, the WP `shutdown` hook still runs. 

I found this while investigating #516 and triggered an error in the WooCommerce initialization. Both the WC error and the AS error were logged. #498 attempted to mitigate a conflict with WP Rocket where the Action data store appeared to not have been initialized.

This wouldn't have been a problem with the WP Post store as WP initializes the core database variables before loading any plugins. Data stores that rely on custom tables should not be used in the `shutdown` hook.

Scheduling the migration hook should be moved out of the `shutdown` hook and the documentation should be updated to reflect the limitation.

rrennick:
Closes #536

This PR moves schedule the migration from the `shutdown` hook to the `wp_loaded` hook and schedules it for 1 minute in the future. It updates the unit tests to account for the migration action existing when the unit tests run.

### Testing

- delete the `action_scheduler_migration_status` setting
- Check Scheduled Actions -> Tools -> Pending for the migration hook scheduled to < 60 seconds
- Allow a minute to pass
- Check that the migration action is removed
- Check that the `action_scheduler_migration_status` setting is `complete`

### DIFF
--- a/classes/abstracts/ActionScheduler.php
+++ b/classes/abstracts/ActionScheduler.php
@@ -186,7 +186,7 @@ abstract class ActionScheduler {
 	 * @param string $function_name The name of the function being called.
 	 */
 	public static function check_shutdown_hook( $function_name ) {
-		if ( 'shutdown' == current_filter() ) {
+		if ( 'shutdown' === current_filter() ) {
 			$message = sprintf(
 				/* translators: $1: open code tag, $2: function name, $3: close code tag. */
 				__( 'Action Scheduler function %1$s%2$s%3$s should not be used within the WordPress %1$sshutdown%3$s hook.', 'action-scheduler' ),
--- a/classes/abstracts/ActionScheduler.php
+++ b/classes/abstracts/ActionScheduler.php
@@ -181,6 +181,24 @@ abstract class ActionScheduler {
 	}
 
 	/**
+	 * Issue deprecated warning if an Action Scheduler function is called in the shutdown hook.
+	 *
+	 * @param string $function_name The name of the function being called.
+	 */
+	public static function check_shutdown_hook( $function_name ) {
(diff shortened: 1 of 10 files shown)


classes/migration/Scheduler.php
```diff
@@ -83,7 +83,7 @@ public function schedule_migration( $when = 0 ) {
 		}
 
 		if ( empty( $when ) ) {
-			$when = time();
+			$when = time() + 60;
```
jeffstieler, line 86: Nitpick - could use `MINUTE_IN_SECONDS` here.

classes/abstracts/ActionScheduler.php
```diff
@@ -180,6 +180,24 @@ public static function init( $plugin_file ) {
 		add_action( 'action_scheduler/migration_complete', 'ActionScheduler_WPCommentCleaner::maybe_schedule_cleanup' );
 	}
 
+	/**
+	 * Issue deprecated warning if an Action Scheduler function is called in the shutdown hook.
+	 *
+	 * @param string $function_name The name of the function being called.
+	 */
+	public static function check_shutdown_hook( $function_name ) {
+		if ( 'shutdown' == current_filter() ) {
```
jeffstieler, line 189: Nitpick - linters in our other projects would complain about not using strict equals here.

rrennick:
> two small nitpicks if you want to change them.

@jeffstieler Thanks, fixed in https://github.com/woocommerce/action-scheduler/commit/fdfb055285f4cb9f8ee49635f0c3279ca4c4238b Will merge once Travis passes.

PR #559, opened 2020-05-25
Fire do_action() inside ActionScheduler_DBStore::delete_action() BEFORE deleting action

EricBusch:
Currently, the `do_action( 'action_scheduler_deleted_action', $action_id );` hook is fired **_after_** the action has been deleted. This makes it impossible to hook into the `action_scheduler_deleted_action`, lookup the action and then do something based on the type of action it is.

It's also inconsistent with `wpPostStore::delete_action()` which fires the `do_action( 'action_scheduler_deleted_action', $action_id );` hook **_before_** performing the deletion. The methods should execute the same code in the same order.

My PR checks to ensure the `$action_id` exists then fires the `action_scheduler_deleted_action` if it does, then finally performs the delete.

### DIFF
--- a/classes/data-stores/ActionScheduler_DBStore.php
+++ b/classes/data-stores/ActionScheduler_DBStore.php
@@ -543,11 +543,13 @@ class ActionScheduler_DBStore extends ActionScheduler_Store {
 	public function delete_action( $action_id ) {
 		/** @var \wpdb $wpdb */
 		global $wpdb;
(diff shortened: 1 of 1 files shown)


thenbrent:
@EricBusch thanks for the PR. Changing when a hook is fired is breaking change, so this patch couldn't be released outside of a major version. You can add a new hook instead though and that may be included in a minor version or patch release.

EricBusch:
@thenbrent Good point. 

@rrennick Thanks for the suggestion.

I followed @rrennick's suggestion and created a new PR: https://github.com/woocommerce/action-scheduler/pull/561

rrennick:
Superseded by #561

PR #563, opened 2020-05-27
Query actions with partial match

alexmigf:
Hello,

I belong to the WP Overnight team and we have a plugin that uses the Action Scheduler for scheduling emails.

The `args` column inside the `wp_actionscheduler_actions` table has data saved has JSON array, and with the current query is not possible to query for matches inside a multidimensional array.

Because our plugin could greatly benefit from this addition, we decided to extend your `ActionScheduler_DBStore` class and perform additional changes to the `get_query_actions_sql` method. Our plugin is a WooCommerce plugin, and since WC bundles its own version of Action Scheduler, loading a modified class using `action_scheduler_store_class` is a potential source of conflicts. This solution also could benefit other plugins too, that's why we think it makes sense to implement this into the Action Scheduler core.

We added a new query argument `args_partial_match` to indicate if we want to use the partial match for the `args` or keep using the default. For the partial match we are using the MySQL `JSON_EXTRACT` function if the database version is superior to 5.7, and using the `LIKE` operator if the version is inferior.

```php
// Default query (we have to match all key value pairs)
$args = array(
	'per_page' => -1,
	'hook' => 'wpo_wcsre_send_emails',
	'args' => array(
		'order_id' => 80,
		'email_id' => 81,
	)
);
$query = as_get_scheduled_actions($args);

// With the new partial match query (we can match only one key value pair)
$args = array(
	'per_page' => -1,
	'hook'=> 'wpo_wcsre_send_emails',
	'args'=> array(
		'email_id'=> 81,
	),
	'args_partial_match'=> true
);
$query = as_get_scheduled_actions($args);
```

Feel free to comment or suggest modifications.


thenbrent:
> since WC bundles its own version of Action Scheduler, loading a modified class using action_scheduler_store_class is a potential source of conflicts. This solution also could benefit other plugins too, that's why we think it makes sense to implement this into the Action Scheduler core.

@alexmigf thank you for both recognizing the risk of conflicts shipping this outside of AS itself, and contributing upstream to the library. Both are greatly appreciated to help improve everyone's experience with AS.

After a quick review, this looks fine to me. The concerns I have around adding this type of functionality have already been accounted for in the patch by using the new `args_partial_match` param.

I'll let @rrennick do proper review though before it's merged.

rrennick:
@alexmigf Could you provide an example `$args` with the code snippet you are using that doesn't work with the current datastore but does work with this PR?

rrennick:
@alexmigf Nice work on this one. I do see two issues

- This only works in the custom table data store. If the partial match logic was moved to the base data store then any datastore using `json_encode()` on `$args` could take advantage of it.
- It only works when the length of json_encoded `$args` is 191 characters or less in the custom table data store.

classes/data-stores/ActionScheduler_DBStore.php
```diff
@@ -305,9 +307,42 @@ protected function get_query_actions_sql( array $query, $select_or_count = 'sele
 			$sql          .= " AND a.hook=%s";
 			$sql_params[] = $query[ 'hook' ];
 		}
+
 		if ( ! is_null( $query[ 'args' ] ) ) {
-			$sql          .= " AND a.args=%s";
-			$sql_params[] = $this->get_args_for_query( $query[ 'args' ] );
+			if( isset($query[ 'args_partial_match' ]) && $query[ 'args_partial_match' ] === true && $mysql_version >= '5.7' ) {
```
abhishek-pokhriyal, line 312: 1. > When referring to array items, only include a space around the index if it is a variable.
1. > Yoda Conditions
```suggestion
			if ( isset( $query['args_partial_match'] ) && true === $query['args_partial_match'] && $mysql_version >= '5.7' ) {
```

classes/data-stores/ActionScheduler_DBStore.php
```diff
@@ -305,9 +307,42 @@ protected function get_query_actions_sql( array $query, $select_or_count = 'sele
...
+				foreach ( $query[ 'args' ] as $key => $value ) {
+					$supported_types = [ 'integer', 'boolean', 'double', 'string' ];
```
abhishek-pokhriyal, line 314: https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#declaring-arrays
```suggestion
					$supported_types = array( 'integer', 'boolean', 'double', 'string' );
```

pokhiii (commented):
@alexmigf Some WPCS suggestions that can be used throughout.

alexmigf:
> @alexmigf Nice work on this one. I do see two issues
> 
>     * This only works in the custom table data store. If the partial match logic was moved to the base data store then any datastore using `json_encode()` on `$args` could take advantage of it.
> 
>     * It only works when the length of json_encoded `$args` is 191 characters or less in the custom table data store.

I just keep the same structure already in place in this array `$query = wp_parse_args( $query, array(...) );`, and also the database table column `data_type` remains the same. I'm assuming this is a limitation for this feature.

rrennick:
> If the partial match logic was moved to the base data store then any datastore using json_encode() on $args could take advantage of it.

@alexmigf The base data store is https://github.com/woocommerce/action-scheduler/blob/master/classes/abstracts/ActionScheduler_Store.php . The current implementation in the PR only applies to the included tables data store. If you aren't comfortable with splitting the logic across the two data stores, let us know.

alexmigf:
@rrennick I'm not quite sure what you mean by:

> splitting the logic across the two data stores

Are you referring to this? https://github.com/woocommerce/action-scheduler/blob/master/classes/data-stores/ActionScheduler_wpPostStore.php

Because i believe **wpPostStore** was deprecated for core usage.

Let me know your thoughts.

rrennick:
> Because i believe wpPostStore was deprecated for core usage.

The tables data store requires PHP 5.5. Also, there are sites using custom data stores. This is a great enhancement but the base store needs to support the parameter with fallback behaviour. Then both the included stores and custom stores can take advantage of that logic.

alexmigf:
@rrennick The tables data store correspond to the **DBStore** and the base store to **wpPostStore**, is that correct?

rrennick:
> and the base store to wpPostStore

Base store class is https://github.com/woocommerce/action-scheduler/blob/master/classes/abstracts/ActionScheduler_Store.php.

Spreeuw:
@rrennick I'm not sure I'm missing something here, but the class that you're linking is the abstract, and as such doesn't implement anything related to the processing of the parameters itself (nor mentions any of the existing parameters).
This pull request only adds some additional logic to the `get_query_actions_sql` method (something that might not even be relevant to custom data stores, depending on how they handle/store the action parameters), it doesn't change anything to how the existing parameters are processed.
I'd love to hear your thoughts on this!

dan-lutd:
@rrennick and @alexmigf   I will test this for my simple use case.

Konamiman:
Hi @rrennick, this one is quite old, do you think it would be ready to merge once the conflicts are resolved or is there anything else that should be changed?

Also @alexmigf, could you please fix the conflicts and add some tests for the added functionality?

Edit: I see now that @rrennick requested a change that hasn't been implemented, and which [apparently requires further clarification](https://github.com/woocommerce/action-scheduler/pull/563#issuecomment-659574441). To start with, @alexmigf are you still interested in this being eventually merged?

rrennick:
@Konamiman Thanks for the ping. The enhancement in this PR looks like it could benefit quite a few use cases. Feel free to follow through with this one as you see fit.

jorgeatorres:
Hi @alexmigf!

Thanks for your contribution. We understand it's been a while since you last worked on this, so please let us know if you're still interested in seeing this through (reiterating the question [asked here](https://github.com/woocommerce/action-scheduler/pull/563#issuecomment-1220439575)). We would love to merge this PR, but it would require a a little work before we can do that.

- First of fall, a rebase is needed to fix conflicts.
- There's also some feedback I've added in the code itself.
- Finally, we're concerned about the `LIKE` implementation working differently from the JSON-based one (see example below) and wonder whether we should drop the LIKE implementation altogether and throw an exception or directly require MySQL 5.7 in AS. Alternatively, we could support different types of matching, for example, by allowing different values for `args_partial_match` such as 'json' and 'like' instead of it being a boolean. This would at least make the behavior you get explicit (and not depend on which MySQL version you have installed). What are your thoughts on this?
   
   <details>
   <summary>Example: LIKE vs JSON_EXTRACT differences in behavior</summary>
   
   An entry with, say, JSON args such as `{"test":{"email":"myemail"}}`, would not be picked up by the `JSON_EXTRACT` implementation if passed `'email' => 'myemail'` as parameter (because `'email'` is not a top-level property) but it'd be returned by the `LIKE` implementation, which is not really expected.
   </details>

Thanks again!

classes/data-stores/ActionScheduler_DBStore.php
```diff
@@ -268,78 +268,113 @@ protected function get_query_actions_sql( array $query, $select_or_count = 'sele
...
+		}
+
+		if ( ! is_null( $query['args'] ) ) {
+			if ( isset( $query['args_partial_match'] ) && true === $query['args_partial_match'] && $mysql_version >= '5.7' ) {
+				foreach ( $query['args'] as $key => $value ) {
+					$supported_types = array( 'integer', 'boolean', 'double', 'string' );
+					if ( ! in_array( gettype( $value ), $supported_types ) ) {
+						continue;
+					}
+					switch ( gettype( $value ) ) {
```
jorgeatorres: For simplicity, instead of this `switch` with 4 almost identical cases, we could add the placeholders directly to the `$supported_types` array (such as `'double' => '%f'`) and use that in the `$sql` var.

classes/data-stores/ActionScheduler_DBStore.php
```diff
@@ -268,78 +268,113 @@ protected function get_query_actions_sql( array $query, $select_or_count = 'sele
...
-			'claimed'          => null,
-			'per_page'         => 5,
-			'offset'           => 0,
-			'orderby'          => 'date',
-			'order'            => 'ASC',
-		] );
+		$query = wp_parse_args( $query, array(
+			'hook'             		=> '',
+			'args'             		=> null,
+			'args_partial_match'	=> false,
```
jorgeatorres: Indentation seems funky here.

jorgeatorres (changes requested):

alexmigf:
@jorgeatorres @Konamiman sure, I just need some time to get back to this.

alexmigf:
@jorgeatorres 

> Finally, we're concerned about the LIKE implementation working differently from the JSON-based one (see example below) and wonder whether we should drop the LIKE implementation altogether and throw an exception or directly require MySQL 5.7 in AS. Alternatively, we could support different types of matching, for example, by allowing different values for args_partial_match such as 'json' and 'like' instead of it being a boolean. This would at least make the behavior you get explicit (and not depend on which MySQL version you have installed). What are your thoughts on this?

To make the implementation more clear to everyone dropping `LIKE` would be ideal, but let me know if you are able to change the MySQL version requirement or if you prefer both ways.

joppuyo:
I would be interested in seeing the LIKE-based implementation since in my case I don't have nested keys and I would like the ensure the broadest possible compatibility. But maybe it should be an opt-in feature along with a note in the readme about the limitations of this method to prevent any surprises for the users of the library.

According to the [wordpress.org statistics](https://wordpress.org/about/stats/), about half the sites are still running a version under 5.7.

jorgeatorres:
Hi @alexmigf! 👋

Thanks for getting back to me. We discussed this a bit more on the team, and considering the `LIKE` implementation works everywhere we feel it shouldn't be dropped, but we should instead support both approaches.

We propose this moves forward in this way: we rename the query arg as `partial_args_matching` to better convey what it does, but instead of it being either `true` or `false`, it'd support 3 possible values:

- `'off'` which would mean no partial matching occurs and would result in the exact same behavior we currently have. This value will be **the default** (i.e. the same as not setting `partial_args_matching` at all).
- `'like'` will engage the `LIKE` implementation.
- `'json'` will engage the `JSON_*()` methods implementation and will check the db server version, throwing an exception if < 5.7.

For any other value outside of those 3 we will either throw an exception or default to `'off'`, which is something we can decide later during PR review.
With those 3 options (and a default value) we should be able to handle all the 3 cases in a `switch` or similar, maybe simplifying the code a bit.

We feel this would allow us to reach a wider audience and for use cases that need the specialized JSON version (such as yours), that will be available too.

Some examples:

```php
as_get_scheduled_actions(
	[
		'per_page'              => -1,
		'args'                  => [
			'email_address' => 'peter.sellers@strangelove.test',
		],
		'partial_args_matching' => 'like',
	]
);

as_get_scheduled_actions(
	[
		'per_page'              => -1,
		'args'                  => [
			'email_address' => 'peter.sellers@strangelove.test',
		],
		'partial_args_matching' => 'json',
	]
);

// This would be the same as passing `'partial_args_matching' => 'off'` as arg.
as_get_scheduled_actions(
	[
		'per_page'              => -1,
		'args'                  => [
			'email_address' => 'peter.sellers@strangelove.test',
		],
	]
);
```

Please let me know what your thoughts are on this suggested approach. Also, let me reiterate our appreciation for the work you're doing and for picking this up again after quite a while. 💯

alexmigf:
@jorgeatorres review requested and thank you for the valuable inputs.

classes/data-stores/ActionScheduler_DBStore.php
```diff
@@ -375,56 +375,103 @@ protected function get_query_actions_sql( array $query, $select_or_count = 'sele
...
-			$sql         .= ' AND a.hook=%s';
+		if ( ! empty( $query['hook'] ) ) {
+			$sql          .= " AND a.hook=%s";
 			$sql_params[] = $query['hook'];
 		}
-		if ( ! is_null( $query['args'] ) ) {
-			$sql         .= ' AND a.args=%s';
-			$sql_params[] = $this->get_args_for_query( $query['args'] );
+
+		if ( ! is_null( $query['args'] ) && isset( $query['partial_args_matching'] ) ) {
```
jorgeatorres: `$query['partial_args_matching']` should always be set thanks to the `wp_parse_args()` call above. It should be enough to check that `$query['args']` is not null here.

classes/data-stores/ActionScheduler_DBStore.php
```diff
@@ -375,56 +375,103 @@ protected function get_query_actions_sql( array $query, $select_or_count = 'sele
...
+			'orderby'               => 'date',
+			'order'                 => 'ASC',
+		 ) );
 
 		/** @var \wpdb $wpdb */
 		global $wpdb;
-		$sql        = ( 'count' === $select_or_count ) ? 'SELECT count(a.action_id)' : 'SELECT a.action_id';
-		$sql       .= " FROM {$wpdb->actionscheduler_actions} a";
-		$sql_params = array();
+		$mysql_version = $wpdb->db_version();
```
jorgeatorres: `$wpdb->db_version()` doesn't report versions for MariaDB correctly: the method will return "5.5.5" under certain PHP versions even if you're on MariaDB 10.6.x or similar. And even when it'll correctly report the version (say, 10.6.8), it will be the MariaDB version, not the MySQL version, so comparing to 5.7 might give us incorrect values, as MariaDB versions have been > 5.7 for quite a while now, even though support for `JSON_*` functions was introduced [in 10.2](https://mariadb.com/kb/en/changes-improvements-in-mariadb-102/).

In those cases, the correct information can be read in the "server info" data (`$wpdb->db_server_info()`), which could look like `5.5.5-10.6.8-MariaDB`.

Ideally proper handling of the MariaDB/MySQL version would be included in `$wpdb` itself, but in the mean time, we might want to be more clever about it.

A better way to detect `JSON_*` support accounting for MariaDB, would be something like the following:

```php
if ( str_contains( $wpdb->db_server_info(), 'MariaDB' ) ) {
	$supports_json = version_compare(
		PHP_VERSION_ID >= 80016 ? $wpdb->db_version() : preg_replace( '/[^0-9.].*/', '', str_replace( '5.5.5-', '', $wpdb->db_server_info() ) ),
		'10.2',
		'>='
	);
} else {
	$supports_json = version_compare( $wpdb->db_version(), '5.7', '>=' );
}
```

Even if it's a bit cumbersome, that covers most cases. `$supports_json` will be `true` or `false` depending on whether the `JSON_EXTRACT` function can be used.

classes/data-stores/ActionScheduler_DBStore.php
```diff
@@ -375,56 +375,103 @@ protected function get_query_actions_sql( array $query, $select_or_count = 'sele
...
+					if ( $mysql_version < '5.7' ) {
+						throw new \RuntimeException( __( 'JSON partial matching requires MySQL 5.7+.', 'action-scheduler' ) );
+					}
+					foreach ( $query['args'] as $key => $value ) {
+						$supported_types = array(
+							'integer' => '%d',
+							'boolean' => '%s',
+							'double'  => '%f',
+							'string'  => '%s',
+						);
```
jorgeatorres: This could be outside of the `foreach()` as it applies to all `$query['args']` so not necessary to assign it on every iteration.

classes/data-stores/ActionScheduler_DBStore.php
```diff
@@ -375,56 +375,103 @@ protected function get_query_actions_sql( array $query, $select_or_count = 'sele
...
+						foreach ( $supported_types as $type => $placeholder ) {
+							if ( gettype( $value ) === $type ) {
+								if ( 'boolean' === $type ) {
+									$value = $value ? 'true' : 'false';
+								}
+								$sql .= ' AND JSON_EXTRACT(a.args, %s)='.$placeholder;
+							}
+						}
```
jorgeatorres: Instead of iterating over all supported types, and given we already know the type, we could do something like the following, which might be a bit simpler:

```php
$value_type  = gettype( $value );
$placeholder = isset( $supported_types[ $value_type ] ) ? $supported_types[ $value_type ] : false;
...
```

That way, we can also react to types that aren't supported (say, "object" or "array"). Not sure what we want to do in those cases, though. Do we want to silently ignore the arg or throw an exception? What are your thoughts @alexmigf?

classes/data-stores/ActionScheduler_DBStore.php
```diff
@@ -375,56 +375,103 @@ protected function get_query_actions_sql( array $query, $select_or_count = 'sele
...
+				$sql  .= " AND a.status IN (";
+				$last = end( $query[ 'status' ] );
+				foreach( $query[ 'status' ] as $status ) {
+					$sql .= "%s";
+					if( $status != $last ) {
+						$sql .= ",";
+					}
+					$sql_params[] = $status;
+				}
+				$sql .= ")";
```
jorgeatorres: I might be missing something, but why do we need to overwrite this part? Even if it's eventually necessary, it might be best to leave this as-is as it's a bit out of scope for this particular PR. What do you think?

jorgeatorres (changes requested):
Hi @alexmigf!

Thanks for making the requested changes. I really like how this is turning out 💯.
I left a few comments, which should be relatively easy to address. After that, and some more testing, I believe we'll be ready to merge.

Please let me know what you think. And once again, thanks!

classes/data-stores/ActionScheduler_DBStore.php
alexmigf: I think we should throw an exception, at least we give some feedback if something is not supported, but let me know your opinion.

classes/data-stores/ActionScheduler_DBStore.php
alexmigf: Sorry, must have been reverted during the trunk merge.

jorgeatorres (approved):
Hi @alexmigf!

Thanks again. I've had to push a couple of commits to address a PHP 7.x/WP 5.2.x compatibility issue as Action Scheduler needs to work with those versions.

I've also added a tiny change to make sure we properly escape things when using the `like` algorithm, so that queries with `args` such as `['key' => 'value_with%']` are interpreted properly (i.e. find actions with `value_with%` as value and not with _any_ value that starts with `value_with`).

Other than that, I really like how the implementation turned out here. It's pure gold 🥇.

Thanks for all your hard work and patience.

alewolf:
This has not been added to the documentation yet. Is there a particular reason? Would you accept a PR that adds it to the docs?

jorgeatorres:
> This has not been added to the documentation yet. Is there a particular reason? Would you accept a PR that adds it to the docs?

Hey @alewolf! There isn't any particular reason. We will be happy to review a PR that adds some documentation on this feature.

mellinger:
Is there any reason this shouldn't be modified to also work with extended args?

PR #567, opened 2020-05-29
Improve "Scheduled Date" for async actions

### LINKED ISSUE #564: Improve "Scheduled Date" for async actions
Yesterday I was providing support to a client, and took a look at the Scheduled Actions page, and saw several actions with `0000-00-00 00:00:00` in the _Scheduled Date_ column. Initially I thought something was wrong, but then recalled that particular hook was running asynchronously.

Is it objectionable to change the value of the _Scheduled Date_ column for async actions to something that is not in date format, to make it clear that the action was created as an async action?

Something like "Asynchronous" would be much clearer. Or perhaps just the time that the async action was registered would be clear enough (because actions don't always run at the scheduled time anyway).

crstauf:
Closes #564.

### DIFF
--- a/classes/ActionScheduler_ListTable.php
+++ b/classes/ActionScheduler_ListTable.php
@@ -445,7 +445,7 @@ class ActionScheduler_ListTable extends ActionScheduler_Abstract_ListTable {
 		$schedule_display_string = '';
 
 		if ( is_a( $schedule, 'ActionScheduler_NullSchedule' ) ) {
-			return 'async';
+			return __( 'async', 'action-scheduler' );
 		}
 
 		if ( ! $schedule->get_date() ) {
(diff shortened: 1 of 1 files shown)


classes/ActionScheduler_ListTable.php
```diff
@@ -444,6 +444,10 @@ protected function get_schedule_display_string( ActionScheduler_Schedule $schedu
 
 		$schedule_display_string = '';
 
+		if ( is_a( $schedule, 'ActionScheduler_NullSchedule' ) ) {
+			return 'async';
```
rrennick, line 448: @crstauf This needs to be translatable. Also, I've created a `version/3.2.0` branch that we will be merging to. If you could base future PRs against it that would be great :)

crstauf: > needs to be translatable

Made translatable in 590a793.

crstauf: > This needs to be translatable.

@rrennick Made translatable in 590a793.

rrennick (approved):
@crstauf Thanks for the PR.

PR #572, opened 2020-06-03
Fatal Error when WooCommerce is not installed

### LINKED ISSUE #549: Fatal Error when WooCommerce is not installed
Uncaught Error: Call to undefined function wc_help_tip() in /wp-content/plugins/my-plugin/packages/action-scheduler/classes/ActionScheduler_wcSystemStatus.php:101

crstauf:
Closes #549.

### DIFF
--- a/classes/ActionScheduler_wcSystemStatus.php
+++ b/classes/ActionScheduler_wcSystemStatus.php
@@ -96,8 +96,9 @@ class ActionScheduler_wcSystemStatus {
 		$as_datastore = get_class( ActionScheduler_Store::instance() );
 
 		$tip = '';
-		if ( function_exists( 'wc_help_tip' ) )
+		if ( function_exists( 'wc_help_tip' ) ) {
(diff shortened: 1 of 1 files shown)


classes/ActionScheduler_wcSystemStatus.php
```diff
@@ -94,12 +94,16 @@ protected function get_action_status_date( $status, $date_type = 'oldest' ) {
 	protected function get_template( $status_labels, $action_counts, $oldest_and_newest ) {
 		$as_version = ActionScheduler_Versions::instance()->latest_version();
 		$as_datastore = get_class( ActionScheduler_Store::instance() );
+
+		$tip = '';
+		if ( function_exists( 'wc_help_tip' ) )
```
rrennick, line 99: @crstauf Could you add braces to the `if` block?

rrennick (approved):
@crstauf Thanks for the fix.

PR #575, opened 2020-06-10
Persist WP CLI `migrate` command with friendly message

### LINKED ISSUE #574: wpcli `migrate` subcommand disappears after one use, help disappears, and not like `wp wc update`
`migrate` subcommand disappears after one use, its help disappears, and it behaves differently than commands like `wp wc update`.

https://github.com/woocommerce/action-scheduler/blob/f3a13eb9b5c9b1f06db1144caa219202c27deef6/classes/abstracts/ActionScheduler.php#L169-L172

## Setup

* Woo 4.2.0 which includes Ac 3.1.6
* Also reproduces on current Ac master as linked above

## Repo

1. Install a complete Woo setup by command-line (not UI or webpages)
2. `wpcli action-scheduler migrate`
3. `wpcli action-scheduler migrate`

## Result

Error and a non-zero exit code

```
Error: 'migrate' is not a registered subcommand of 'action-scheduler'. See 'wp help action-scheduler' for available subcommands.`
```

Further...when you follow those instructions to get help with `wp help action-scheduler` all help on it has vanished.

```
...
SUBCOMMANDS

  run      Run the Action Scheduler

GLOBAL PARAMETERS
...
```

## Expected

No error and action scheduler should provide a friendly notice that no migrate/update is needed and note the current version. For example, this is the friendly notice (with exit code=0) that Woo provides when `wp wc update` is run:

```
Success: No updates required. Database version is 4.2.0
```

Also, complete help should always be available. When `wp help action-scheduler` is run, it should provide a list of all subcommands (including `migrate`) and help for them all.

crstauf:
Closes #574.

### DIFF
--- a/classes/WP_CLI/Migration_Command.php
+++ b/classes/WP_CLI/Migration_Command.php
@@ -75,6 +75,16 @@ class Migration_Command extends WP_CLI_Command {
 	 * @return void
 	 */
 	public function migrate( $positional_args, $assoc_args ) {
+		if ( ! Controller::instance()->allow_migration() ) {
+			WP_CLI::error( 'Migration prevented due to missing depdendencies or by current data store.' );
+			return;
+		}
+
+		if ( \ActionScheduler_DataController::is_migration_complete() ) {
+			WP_CLI::success( sprintf( 'No migration required. Data store is %s.', get_class( \ActionScheduler_Store::instance() ) ) );
+			return;
+		}
+
 		$this->init_logging();
 
 		$config = $this->get_migration_config( $assoc_args );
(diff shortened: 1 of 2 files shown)


classes/WP_CLI/Migration_Command.php
```diff
@@ -75,6 +75,16 @@ public function register() {
 	 * @return void
 	 */
 	public function migrate( $positional_args, $assoc_args ) {
+		if ( ! Controller::instance()->allow_migration() ) {
+			WP_CLI::error( 'Migration prevented due to missing depdendencies or by current data store.' );
+			return;
+		}
+
+		if ( \ActionScheduler_DataController::is_migration_complete() ) {
```
rrennick, line 83: @crstauf Could you move this to be the first condition check? When writing early exit checks it is preferable that the most common conditions be tested first.

crstauf: @rrennick Done.

rrennick (approved):
@crstauf Nice work.

PR #596, opened 2020-08-02
Fix #589: Add check that function add_action() exists

### LINKED ISSUE #589: Add function_exists( 'add_action' ) check to initialization file
This is the same issue as #303, which was resolved with #304 however while the PR was merged into `master` it appears it was never merged into the [version_3_0_0](https://github.com/woocommerce/action-scheduler/tree/version_3_0_0) branch and eventually the `version_3_0_0` branch was merged into `master` overwriting the fix (I believe this happened in 52bd355 as the fix exists in the parent commit (9275e2b)).

To summarize #303 

When we load Action Scheduler via Composer, we still need to execute the file `action-scheduler.php` to initialize everything.

Composer allows us to autoload a file, which handles this beautifully

```
"autoload": {
    "files": [
        "vendor/prospress/action-scheduler/action-scheduler.php"
    ]
},
```

**except** when we run a PHP tool added via Composer (e.g. `phpcs`) that runs _without_ WordPress

When `phpcs` runs, WordPress is not loaded and therefore the function `add_action()` does not exist and a Fatal Error is thrown.

By adding the check `function_exists( 'add_action' )`,

- when WordPress is loaded, everything works as it always has because `add_action()` exists
- when WordPress is NOT loaded (e.g. when `phpcs` is run), all of the Action Scheduler initialization is skipped (which isn't a problem because WordPress is not present) and we don't throw a Fatal Error

asharirfan:
By checking that the function add_action() exists, before we call it, we
can avoid throwing a fatal error when a PHP tool is run (e.g. phpcs)
even if we are loading `action-scheduler.php` via Composer's autoload.

See #303
See #589

### DIFF
--- a/action-scheduler.php
+++ b/action-scheduler.php
@@ -27,6 +27,11 @@
 
 if ( ! function_exists( 'action_scheduler_register_3_dot_1_dot_6' ) ) {
 
+	if ( ! function_exists( 'add_action' ) ) {
+		// We are running outside of the context of WordPress.
+		return;
+	}
+
 	if ( ! class_exists( 'ActionScheduler_Versions' ) ) {
 		require_once( 'classes/ActionScheduler_Versions.php' );
 		add_action( 'plugins_loaded', array( 'ActionScheduler_Versions', 'initialize_latest_version' ), 1, 0 );
--- a/action-scheduler.php
+++ b/action-scheduler.php
@@ -25,12 +25,7 @@
  *
  */
 
-if ( ! function_exists( 'action_scheduler_register_3_dot_1_dot_6' ) ) {
-
-	if ( ! function_exists( 'add_action' ) ) {
-		// We are running outside of the context of WordPress.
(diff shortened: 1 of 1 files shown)


rrennick:
@asharirfan Thanks for the PR. Since this PR aims to restore a regression it would be preferred to restore the original code. The original code added the condition to the main `if`. Can you update the PR accordingly?

asharirfan:
@rrennick No problem! I have restored the original code. 

It is worth noting that this condition might have been overlooked during a merge operation before. That's why I added the code in a separate condition.

Cheers!

rrennick:
> It is worth noting that this condition might have been overlooked during a merge operation before.

That is what happened. It wasn't merged into the 3.0.0 branch before the 3.0.0 branch was merged back into master.

rrennick (approved):
Looks good. Thanks for the help.

PR #604, opened 2020-08-20
Change relative include paths to absolute include paths

romarioraffington:
Relates to woocommerce/woocommerce#27269 and woocommerce/woocommerce#27433

## Overview
Relative include paths in PHP can break, when server is running opcache. See here about vulnerability details.

WordPress.com deploy system refuses to include WooCommerce, because of that issue - for more context, see here - internal a8c discussion This PR changes the relative include paths to absolute include pathss.

### DIFF
--- a/action-scheduler.php
+++ b/action-scheduler.php
@@ -26,9 +26,10 @@
  */
 
 if ( ! function_exists( 'action_scheduler_register_3_dot_1_dot_6' ) ) {
+	$dir = __DIR__ . DIRECTORY_SEPARATOR;
 
 	if ( ! class_exists( 'ActionScheduler_Versions' ) ) {
-		require_once( dirname( __FILE__ ) . 'classes/ActionScheduler_Versions.php' );
+		require_once( $dir . 'classes/ActionScheduler_Versions.php' );
 		add_action( 'plugins_loaded', array( 'ActionScheduler_Versions', 'initialize_latest_version' ), 1, 0 );
 	}
 
@@ -40,7 +41,7 @@ if ( ! function_exists( 'action_scheduler_register_3_dot_1_dot_6' ) ) {
 	}
 
 	function action_scheduler_initialize_3_dot_1_dot_6() {
-		require_once( dirname( __FILE__ ) . 'classes/abstracts/ActionScheduler.php' );
+		require_once( $dir . 'classes/abstracts/ActionScheduler.php' );
 		ActionScheduler::init( __FILE__ );
 	}
 
--- a/action-scheduler.php
+++ b/action-scheduler.php
@@ -26,7 +26,7 @@
  */
 
 if ( ! function_exists( 'action_scheduler_register_3_dot_1_dot_6' ) ) {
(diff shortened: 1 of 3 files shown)


rrennick:
> Not sure why the CI is throwing this error:

@romarioraffington Sometimes CI has hiccups that cause tests to fail. I triggered a rebuild of the failing run and it passed.

artpi:
Romario, there's one more file to change:

```
Error in: trunk/wp-content/plugins/woocommerce/packages/action-scheduler/classes/abstracts/ActionScheduler.php:
Relative path includes are not allowed. Please fix this relative include before committing the file:
			include( "{$dir}{$class}.php" );

```
https://github.com/woocommerce/action-scheduler/blob/master/classes/abstracts/ActionScheduler.php#L122

I think we can pull `$dir` ahead of `"` and it'll be ok

romarioraffington:
> Romario, there's one more file to change:
> 
> ```
> Error in: trunk/wp-content/plugins/woocommerce/packages/action-scheduler/classes/abstracts/ActionScheduler.php:
> Relative path includes are not allowed. Please fix this relative include before committing the file:
> 			include( "{$dir}{$class}.php" );
> ```
> 
> https://github.com/woocommerce/action-scheduler/blob/master/classes/abstracts/ActionScheduler.php#L122
> 
> I think we can pull `$dir` ahead of `"` and it'll be ok

Fixed in  9924e3b

rrennick:
@romarioraffington I updated this PR to be consistent with https://github.com/woocommerce/woocommerce/pull/27433.

action-scheduler.php
```diff
@@ -28,7 +28,7 @@
 if ( ! function_exists( 'action_scheduler_register_3_dot_1_dot_6' ) && function_exists( 'add_action' ) ) {
 
 	if ( ! class_exists( 'ActionScheduler_Versions' ) ) {
-		require_once( 'classes/ActionScheduler_Versions.php' );
+		require_once( $dir . 'classes/ActionScheduler_Versions.php' );
```
claudiosanches, line 31:
```suggestion
		require_once( __DIR__ . '/classes/ActionScheduler_Versions.php' );
```

action-scheduler.php
```diff
@@ -40,7 +40,7 @@ function action_scheduler_register_3_dot_1_dot_6() {
 	}
 
 	function action_scheduler_initialize_3_dot_1_dot_6() {
-		require_once( 'classes/abstracts/ActionScheduler.php' );
+		require_once( $dir . 'classes/abstracts/ActionScheduler.php' );
```
claudiosanches, line 43:
```suggestion
		require_once( __DIR__ . '/classes/abstracts/ActionScheduler.php' );
```

claudiosanches (changes requested):
There's no more a `$dir` variable, and also should be easier to just use `__DIR__` in those.

claudiosanches (approved):
All fixed now, but maybe we could revert the changes from `classes/abstracts/ActionScheduler.php`, since those doesn't change too much.

Thanks for your help!

PR #612, opened 2020-09-09
Add retries to Actions

n3f:
Adds an optional field to the ActionScheduler API for a retry parameter.  (For non-recurring actions).

```php
// For example, this tries `do_work` 1 time (if it fails the first time).
as_enqueue_async_action( 'do_work', [], '', 1 );
```

Exceptions thrown by an Action schedule a new Action (with the same hook and args) on an exponential backoff algorithm.

Tests included.

- [x] Update the API
- [x] Update store
   - wpPostStore
      - [x] `save_action()`
      - [x] `fetch_action()`
   - DBStore
      - [x] `save_action()`
      - [x] `fetch_action()`
- [x] Implement retries in QueueRunner
- [x] Update UI
- [x] Ensure backwards compatibility
- [x] Add tests

### DIFF
--- a/classes/ActionScheduler_InvalidActionException.php
+++ b/classes/ActionScheduler_InvalidActionException.php
@@ -44,4 +44,22 @@ class ActionScheduler_InvalidActionException extends \InvalidArgumentException i
 
 		return new static( $message );
 	}
+
+	/**
+	 * Create a new exception when the action's retry isn't parsed.
+	 *
+	 * @param string $action_id The action ID with bad args.
+	 * @param array  $retry     Retry params.
+	 * @return static
+	 */
+	public static function from_retry( $action_id, $retry ) {
+		$message = sprintf(
+			/* translators: 1: action ID 2: retry */
+			__( 'Action [%1$s] has an invalid retry: %2$s', 'action-scheduler' ),
+			$action_id,
+			var_export( $retry, true )
+		);
+
+		return new static( $message );
+	}
 }
--- a/classes/ActionScheduler_Retry.php
+++ b/classes/ActionScheduler_Retry.php
@@ -1,0 +1,65 @@
+<?php
+
+/**
+ * Class ActionScheduler_Retry
+ */
+class ActionScheduler_Retry {
+	/** @var int */
+	protected $limit;
+	/** @var int */
+	protected $fails;
+
+	/**
+	 * ActionScheduler_Retry constructor.
+	 *
+	 * @param $limit int Number of retries to attempt.
+	 * @param $fails int Track failures.
+	 */
+	public function __construct( $limit = 0, $fails = 0 ) {
+		$this->set_limit( $limit );
+		$this->set_fails( $fails );
+	}
+
+	/**
+	 * @return int
+	 */
+	public function get_limit() {
+		return $this->limit;
+	}
+
+	/**
+	 * @param int $limit
+	 */
+	public function set_limit( $limit ) {
+		$this->limit = abs( intval( $limit ) );
+	}
+
+	/**
+	 * @return int
+	 */
+	public function get_fails() {
+		return $this->fails;
+	}
+
+	/**
+	 * @param int $fails
+	 */
+	public function set_fails( $fails ) {
+		$this->fails = abs( intval( $fails ) );
+	}
+
+	/**
+	 * Increment the failures and check that the failures <= limit.
+	 * @return bool
+	 */
+	public function is_valid_after_fail() {
+		// Increment the failures.
+		$this->fails ++;
+
+		// Evaluate attempts vs. failures.
+		if ( $this->fails > $this->limit ) {
+			return false;
+		}
+		return true;
+	}
+}
--- a/classes/abstracts/ActionScheduler_Abstract_QueueRunner.php
+++ b/classes/abstracts/ActionScheduler_Abstract_QueueRunner.php
@@ -114,7 +114,7 @@ abstract class ActionScheduler_Abstract_QueueRunner extends ActionScheduler_Abst
 		// schedule when the webhook should be retried (in seconds).  Grabbing the filter return first and using that
 		// unless it is not a number then defaulting to algorithm.
 		$backoff = apply_filters( 'action_scheduler_retry_backoff', null, $failures, $action );
-		$backoff = is_numeric( $backoff ) ? intval( $backoff ) : ( 2 ** $failures ) * 100;
+		$backoff = is_numeric( $backoff ) ? intval( $backoff ) : pow( 2, $failures ) * 100;
 		return ActionScheduler::factory()->single(
 			$action->get_hook(),
 			$action->get_args(),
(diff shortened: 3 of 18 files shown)


n3f:
Closing/opening to re-trigger travis.

rrennick:
> Update version / hooks

@n3f We have other PRs in progress that will require a version bump so we will be bumping the version in a separate PR.

classes/data-stores/ActionScheduler_DBStore.php
```diff
@@ -156,7 +141,18 @@ public function fetch_action( $action_id ) {
...
 		if ( ! empty( $data->extended_args ) ) {
-			$data->args = $data->extended_args;
+			$extended_args = json_decode( $data->extended_args, true );
+			// Stored data from before the extended usage
+			if ( ! is_array( $extended_args ) ) {
+				$data->args = $data->extended_args;
+				$extended_args = array();
+			}
+			if ( array_key_exists( 'args', $extended_args ) ) {
+				$data->args = $extended_args['args'];
```
rrennick, line 151: @n3f I know you haven't marked this PR for review yet. This implementation of retry storage looks like it might be problematic in a couple ways:

- Prior to the change in this block of code `$data->args` is a `json_encoded` array. With this block it's an array. In `make_action_from_db_record()` it is passed to `json_decode()` which expects a string
- The action search includes the extended args field https://github.com/woocommerce/action-scheduler/blob/version_3_2_0/classes/data-stores/ActionScheduler_DBStore.php#L346 and adding non-args data to that field could result in false positives

In addition, there are other use cases where retries would be useful (eg. Subscriptions). Adding a new field to the schema would simplify this PR and leave room for other implementations.

vdwijngaert:
The ability to retry certain actions is a must-have for any queue runner, nice work!

What's the current status on this PR? Anything I can do to help?

n3f:
It works as is, but @rrennick has a valid point about the storage.  I was hoping to not require schema changes, but that's probably what's needed at this point.

n3f:
@ObliviousHarmony, I wanted to get your thoughts on this and how it could potentially be integrated with WooCommerce Webhooks.
* Would we need some mechanism for the webhook to interrogate the process to see if there would be a retry (before logging a failed delivery)?
* Is it OK to throw an Exception to trigger a retry?

classes/ActionScheduler_InvalidActionException.php
```diff
@@ -44,4 +44,22 @@ public static function from_decoding_args( $action_id, $args = array() ) {
...
 	}
+
+	/**
+	 * Create a new exception when the action's retry isn't parsed.
+	 *
+	 * @param string $action_id The action ID with bad args.
+	 * @param array  $retry     Retry params.
+	 * @return static
+	 */
+	public static function from_retry( $action_id, $retry ) {
```
vedanshujain, line 55: this method does not seem to be used.

classes/abstracts/ActionScheduler_Abstract_QueueRunner.php
```diff
@@ -68,6 +68,9 @@ public function process_action( $action_id, $context = '' ) {
 		} catch ( Exception $e ) {
 			if ( $valid_action ) {
 				$this->store->mark_failure( $action_id );
+				if ( isset( $action ) ) {
+					$this->retry_action( $action );
```
vedanshujain, line 72: what do you think about setting `store->mark_failure` after the retry action. so in case, we are able to schedule a retry then the action isn't failed yet. should we have another tab for action in retry? (or some other way to indicate that while the action was failed, we are retrying)

classes/abstracts/ActionScheduler_Abstract_QueueRunner.php
```diff
@@ -79,6 +82,43 @@ public function process_action( $action_id, $context = '' ) {
...
+		// schedule when the webhook should be retried (in seconds).  If an invalid value (not numeric) is returned
+		// DON'T schedule the retry.
+		$backoff = pow( 2, $retry->get_fails() ) * 100;
+		$backoff = apply_filters( 'action_scheduler_retry_backoff', $backoff, $action );
+		if ( empty( $backoff ) || ! is_numeric( $backoff ) ) {
+			return;
+		}
+
+		$date = as_get_datetime_object( time() + abs( (int) $backoff ) );
+		$schedule = new ActionScheduler_SimpleSchedule( $date );
```
vedanshujain, line 117: this is duplicate implementation of creating an action. lets just use `ActionScheduler::factory()->single( ` here instead

classes/ActionScheduler_Retry.php
```diff
@@ -0,0 +1,65 @@
+<?php
+
+/**
+ * Class ActionScheduler_Retry
+ */
+class ActionScheduler_Retry {
```
vedanshujain, line 6: should this class be integrated into `ActionScheduler_Action`. `retry` looks like a property of an action instead of a separate entity. wdyt?

vedanshujain (commented):
I am still reviewing and testing but had some preliminary comments that wanted to share. BTW, looks great, and thanks so much for the work here.

> Is it OK to throw an Exception to trigger a retry?

IMO that's fine. Since retry is not enabled by default (if I understand correctly), someone configuring retry params are giving consent to have job retried.

> Would we need some mechanism for the webhook to interrogate the process to see if there would be a retry (before logging a failed delivery)?

Not sure what means, can you explain a bit more here. Webhook, or any consumer should n't care about underlying retry mechanism right?

I have some other questions:
Is it worth it to have a separate queue or at least a tab in UI for retry jobs. As part of reviewing this PR I also looked at similar systems in other frameworks, and for example, [sidekiq has a separate queue for retry jobs](https://www.rubydoc.info/github/mperham/sidekiq/Sidekiq/JobRetry) before they are moved to morgue.

michaelrimbach:
Any news here? Would love to use this kind of functionality!

barryhughes:
Noodling on this a little, I'm curious why we don't extend support for retries to recurring actions.

On one hand I understand they _may_ not be useful in some cases (perhaps for actions that recur with high frequency) but, hypothetically, what if the service I have sold to my customers is a weekly email: I really want that email to go out, every week, because that's my business offering: and if it doesn't work I'd like it to automatically retry _n_ times. Why wouldn't we allow for that?

Just curious if there are other technical reasons that make it unattractive.

zewelor:
Is this still planned ?

TIBI4:
Webhooks automatically retrying themselves is a must have. Thanks for your effort in this. @n3f 

A reviewer (@vedanshujain) commented, but I thnk that **n3f** didn't read it.

**If someone know a bit about it, and acknowledge the comments, this may get resolved. Or is @rrennick the only one that can merge the pull request?**

Anyways, thanks!

JasonTheAdams:
Would be useful to see this come to fruition. Even simply having the ability to manually re-run/retry an action would be useful. We have an action that gets scheduled as follow up to receiving a webhook. If the action fails we don't have a good way of setting the same action with the same parameters back up. Ideally, we'd see the failed action, look into the problem, fix it the issue, and run the action again.

barryhughes:
Closing, as it doesn't seem like there is an active need for this (and, it would be possible to implement from the parent application via a self-cancelling recurring action, etc).

PR #617, opened 2020-09-16
change all _ in FQ file names to -

### LINKED ISSUE #614: Remove underscores from PHP filenames
The underscores are interfering with some code loaders: pMz3w-bWp-p2

For classes named `ActionScheduler_ClassName` changing the name to `ActionSchedulerClassName` will still be quite readable.

The files in the `/lib/`folder should have an appropriate note at the top documenting the file name change.

rrennick:
Closes #614 

The file name changes in this PR are to address [this vulnerability](https://stackoverflow.com/a/30913177).

### Testing Instructions

1- Go to Tools -> Scheduled Actions -> Pending
2- Add this test snippet
```
    for ( $i = 1; $i < 200; $i++ )
        as_enqueue_async_action( 'as-async-filename' . $i );
} );
```
3- Refresh the screen
  - You should have 200 pending actions

4- Remove the snippet
5- Wait up to a minute to allow the async runner to run
6- Check that the 200 actions created by the snippet were processed
7- Run `wp help action-scheduler` to verify WP CLI code is loading correctly


classes/abstracts/ActionScheduler.php
```diff
@@ -95,7 +95,7 @@ public static function autoload( $class ) {
 
 			switch ( $type ) {
 				case 'WPCLI':
-					$dir = $classes_dir . 'WP_CLI' . $d;
+					$dir = $classes_dir . 'WPCLI' . $d;
```
vedanshujain, line 98: any specific reason as to why this folder name is not changed to `WP-CLI` instead like in other cases?

vedanshujain (commented):
Is it feasible to add files for old classes with a `require_once realpath(..)` statement and a deprecation notice? This is a breaking change in its current form no?

classes/abstracts/ActionScheduler.php
rrennick: > Is it feasible to add files for old classes with a require_once realpath(..) statement and a deprecation notice?

Any plugin requiring an AS class file directly risks installs where AS is broken, logging errors, and/or fatal errors on the site. There were several plugins that had to address this for AS 3.0. The only file that should be accessed outside of AS is `action-scheduler/action-scheduler.php`.

> any specific reason as to why this folder name is not changed to WP-CLI instead like in other cases?

The `_` -> `-` string replacement is done on the filename not the folder. I would have removed the `_` from the filenames instead of replacing except it made some of the class file names difficult to visually parse. Contributors with easily recognize `WPCLI` as it is often referenced in online content without the `_`.

vedanshujain: I see, but we need to mark it as such when we update the version in the WC core.

vedanshujain: I am going to test a bit more before approving, but looks good so far.

vedanshujain (approved):
This is testing well so I am approving, however, I am still not sure that this is not a breaking change. I think adding old file-names with deprecation notice and including correct is a better way (with intention of dropping them later on in major version upgrade) is a better path. Unless, we also mark this as a major version upgrade and then do it directly.

rrennick:
> Unless, we also mark this as a major version upgrade and then do it directly.

3.2.0 will have DB schema changes and was planned as a major version. This now has conflicts that I will address tomorrow.

rrennick:
> I am still not sure that this is not a breaking change. I think adding old file-names with deprecation notice and including correct is a better way

@vedanshujain In https://github.com/woocommerce/woocommerce/pull/25404 Action Scheduler was moved from `includes/libraries` to `packages/action-scheduler` and in https://github.com/woocommerce/woocommerce/pull/27100 the REST API was moved from `packages/woocommerce-rest-api` to `includes/rest-api`. Neither PR included stubs for the old file names. Unless there is evidence that a plugin has started referencing AS files directly since WC 4.0, we should not include  stubs for the changes in this PR.

vedanshujain:
> Action Scheduler was moved from includes/libraries to packages/action-scheduler and in woocommerce/woocommerce#27100 the REST API was moved from packages/woocommerce-rest-api to includes/rest-api. Neither PR included stubs for the old file names.

I agree here, and given that those changes did not cause problems for us, it does give more confidence to go ahead directly. Note that API was getting loaded automatically via PSR4, so requiring individual files was not required anyway and we still monitored it as a high impact change, I think this is the case for ActionScheduler as well.

rrennick:
> Note that API was getting loaded automatically via PSR4 so requiring individual files was not required

AS wasn't/isn't loading using PSR4 but does have its own autoloader. AS is a stand alone plugin and implementation only requires `action-scheduler/action-scheduler.php` which is effectively the same as `woocommerce/woocommerce.php`.

vedanshujain:
@rrennick sounds good, lets 🚢

PR #620, opened 2020-09-18
Update the Travis test matrix

rrennick:
This PR updates the WP test versions in the test matrix. PHP 5.3 isn't supported in any of the tested WP versions so it was removed.

The unit test mock functions were throwing deprecated notices under PHP 7.4. That will need to be investigated and resolved before PHP 7.4 can be added to the matrix.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,43 +3,42 @@
 # Tell Travis CI we're using PHP
 language: php
 
-# We nee to use Precise, not Trusty, to test against PHP 5.3, see https://github.com/travis-ci/travis-ci/issues/8219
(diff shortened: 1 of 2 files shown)


vedanshujain (approved):
LGTM! I assume we will add PHP 7.4 along with php8?

rrennick:
> I assume we will add PHP 7.4 along with php8?

@vedanshujain The mocks in the unit tests cause the tests to fail with 7.4 due to deprecated warnings in the newer version of PHPUnit. We should create an issue for that. PHP 7.4 and possibly 8 should be added with that fix.

PR #626, opened 2020-10-13
Fix unit tests infrastructure and adapt tests to PHP 8

Konamiman:
This pull requests applies the necessary fixes so that unit tests can be run:

- Modify `bootstrap.php` to use the WordPress tests lib from the temporary directory, instead of assuming WordPress is up in the directory hierarchy.

- Modify `bootstrap.php` to require `action-scheduler.php`, necessary to run the tests.

- Modify `phpunit.xml.dist` to remove obsolete attributes that were giving warning when running the tests.

- Add the `tests/bin/install.sh` file to install the WordPress tests library.

- Add a minimal `tests/README.md` file.

- Modify `composer.json` to use PHPUnit 7.5 instead of 5.6.

- Fix the "test" command defined in `composer.json`.

- Fix small portions of tests code for PHP 8 compatibility.

**Note!**

There are still two problems related to PHPUnit:

1. It won't install with PHP 8 active, since the package has a PHP version requirement of `^7.1`.
2. Some tests will throw an error due to incompatibilities of PHPUnit 7 with PHP 8.

These two can be temporarily fixed using [the custom fork of PHPUnit 7 that was created for WooCommerce core](https://github.com/woocommerce/phpunit/pull/1). Apply the changes in `composer.json` from [this pull request](https://github.com/woocommerce/woocommerce/pull/27844), do `composer update` and you're all set.

### DIFF
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -11,7 +11,12 @@ foreach ( array('WP_CONTENT_DIR', 'WP_CONTENT_URL', 'WP_PLUGIN_DIR', 'WP_PLUGIN_
 	}
 }
 
+if( ! defined( 'WP_PLUGIN_DIR' ) ) {
+	define( 'WP_PLUGIN_DIR', dirname( dirname( dirname( __FILE__ ) ) ) );
+}
+
 $wordpress_tests_dir = getenv( 'WP_TESTS_DIR' ) ? getenv( 'WP_TESTS_DIR' ) : sys_get_temp_dir() . '/wordpress-tests-lib';
+require_once $wordpress_tests_dir . '/includes/functions.php';
 require $wordpress_tests_dir . '/includes/bootstrap.php';
 
 require_once dirname(dirname( __FILE__ ) ) .'/action-scheduler.php';
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -11,18 +11,10 @@ foreach ( array('WP_CONTENT_DIR', 'WP_CONTENT_URL', 'WP_PLUGIN_DIR', 'WP_PLUGIN_
 	}
 }
 
-// If the wordpress-tests repo location has been customized (and specified
-// with WP_TESTS_DIR), use that location. This will most commonly be the case
-// when configured for use with Travis CI.
+$wordpress_tests_dir = getenv( 'WP_TESTS_DIR' ) ? getenv( 'WP_TESTS_DIR' ) : sys_get_temp_dir() . '/wordpress-tests-lib';
+require $wordpress_tests_dir . '/includes/bootstrap.php';
 
-// Otherwise, we'll just assume that this plugin is installed in the WordPress
-// SVN external checkout configured in the wordpress-tests repo.
-
-if( false !== getenv( 'WP_TESTS_DIR' ) ) {
-	require getenv( 'WP_TESTS_DIR' ) . '/includes/bootstrap.php';
-} else {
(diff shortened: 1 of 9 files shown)


tests/bootstrap.php
```diff
@@ -11,18 +11,15 @@
 	}
 }
 
-// If the wordpress-tests repo location has been customized (and specified
-// with WP_TESTS_DIR), use that location. This will most commonly be the case
-// when configured for use with Travis CI.
+if( ! defined( 'WP_PLUGIN_DIR' ) ) {
```
roykho, line 14: Space is needed after the `if`

tests/phpunit/runner/ActionScheduler_QueueCleaner_Test.php
```diff
@@ -56,7 +56,11 @@ public function test_delete_canceled_actions() {
 		remove_filter( 'action_scheduler_retention_period', '__return_zero' );
 		remove_action( 'action_scheduler_deleted_action', array( $mock_action, 'action' ), 10 );
 
-		$deleted_actions = array_map( 'reset', $mock_action->get_args() );
+		$deleted_actions = array();
+		foreach( $mock_action->get_args() as $action ) {
```
roykho, line 60: Space is needed after the `foreach`

roykho (commented):
So some small coding standards issue and I see that Travis is passing here but when I run it locally, I am getting two fails.

```
There were 2 failures:

1) Scheduler_Test::test_migration_is_scheduled
Failed asserting that false is true.

/srv/www/wordpress-trunk/public_html/build/wp-content/plugins/action-scheduler/tests/phpunit/migration/Scheduler_Test.php:33
/srv/www/wordpress-trunk/public_html/build/wp-content/plugins/action-scheduler/tests/ActionScheduler_UnitTestCase.php:35

2) Scheduler_Test::test_migration_is_scheduled
Failed asserting that false is true.

/srv/www/wordpress-trunk/public_html/build/wp-content/plugins/action-scheduler/tests/phpunit/migration/Scheduler_Test.php:33
/srv/www/wordpress-trunk/public_html/build/wp-content/plugins/action-scheduler/tests/ActionScheduler_UnitTestCase.php:38
```

Konamiman:
I pushed a commit fixing the code standard issues, and now I'm see the two test failures locally too. I wonder what's the issue? I think I'm using the same setup as Travis.

rrennick:
@Konamiman Are they timestamp related failures?

Konamiman:
@rrennick I think it's related to hooks. The test that fails is creating a `new Scheduler()` and then going straight to assert its `is_migration_scheduled()`, which is expected to be `true` but isn't. That method is verifying if `as_next_scheduled_action( self::HOOK )` is empty.

rrennick:
> I think it's related to hooks. 

@Konamiman That sounds like the bootstrap may need to be modified for PHP 8.

Konamiman:
> @Konamiman That sounds like the bootstrap may need to be modified for PHP 8.

But I'm getting these failures in PHP 7 too, not sure fhat's the case for @roykho too.

peterfabian:
@Konamiman can you please check this one so we can close it out if it's fine?

Konamiman:
@peterfabian It's "as fine as it can be": tests are passing in Travis but a couple of them are failing locally for @roykho and me. Perhaps we can merge this and revisit this issue later.

peterfabian:
@roykho if you agree, would you give this a 👍 please?

roykho (approved):
Yeah so as stated, code looks good with exception of failing tests which we can revisit at a later time if it continues to crop up.

PR #632, opened 2020-10-23
[PHP 8] Set the PHP version to 7.1 in composer.json for PHP 8 compatibility

Konamiman:
When the PHP version in the development system is 8.0 the `composer install` command will fail saying that some packages
require PHP `^7.0` or `^7.1` but version 8.0 doesn't match.

To fix, this commit adds a `platform` setting in composer.json telling composer to assume version 7.1 instead of the actual version in the development system.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,6 +12,11 @@
     "wp-cli/wp-cli": "~1.5.1",
     "woocommerce/woocommerce-sniffs": "0.0.8"
   },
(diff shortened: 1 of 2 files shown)


peterfabian (approved):
Thanks, nice trick. I don't like this solution much, but until the packages will get updated, I think it should be fine.

PR #668, opened 2021-02-27
Add secondary scheduled date checks when claiming actions (DBStore) | #634

### LINKED ISSUE #634: Double check action scheduled date in case they're processing too early
We've had a few reports of database errors causing actions to be processed ahead of the schedule. This happens when the database server's disk is full:

```
[11-Oct-2020 21:15:48 UTC] WordPress database error Disk full (/tmp/#sql_6ff4_0.MAI); waiting for someone to free some space… (errno: 28 "No space left on device") for query SHOW FULL COLUMNS FROM wpstg3_options made by shutdown_action_hook, do_action('shutdown'), WP_Hook->do_action, WP_Hook->apply_filters, ActionScheduler_QueueRunner->maybe_dispatch_async_request, ActionScheduler_OptionLock->set, update_option
```

As far as I understand, if an error like this comes up when looking for scheduled actions that are due for processing, the server sometimes sends back intermediate results (i.e. all rows, or all rows that are pending regardless of their scheduled date, etc) instead of only the actions that were due to process now.

Given that scheduled actions often control mission-critical functions like subscription renewals, could we add an extra check to make sure the scheduled date is not in the future before running it? Ideally this would only happen on queue runners so that users are still able to run an action manually if they need to.

barryhughes:
:ticket: [#634 → Double check action scheduled date in case they're processing too early](https://github.com/woocommerce/action-scheduler/issues/634)

Closes #634

### :confounded: The Problem

We can broadly divide pending scheduled actions into two categories:

- Those that are due now.
- Those that are not due until some future point in time.

When the Queue Runner claims a set of actions it does so with a query that is designed to ignore those actions that are not yet due, which makes sense (as a commenter in the linked issue noted, we want to _make sure the scheduled date is not in the future before running it)._

However, there's a body of reported evidence suggesting that—under unusual conditions, such as if MySQL finds it is out of disk space—the `WHERE` conditions we use to enforce this are seemingly not respected. To combat this, we need an extra set of checks that run at PHP-level.

### :hammer_and_wrench: This Changeset

We were already specifying a `$before_date` when claiming actions (defaulting to 'now') ... with this change, we also test _at PHP-level_ after fetching the individual actions to make sure that the `$before_date` condition was respected (we don't just rely on MySQL).

This way, besides adding extra safety for common cases, if some custom code wishes to stake a claim that explicitly includes future pending actions (by specifying a `$before_date` in the future) then that custom code should still function as previously.

### :memo: Some Notes

* I have focused on the `DBStore` in this PR. Let's confirm we're happy with this approach/refine and adjust as needed, and I can follow-up with a corresponding change for the `wpPostStore`.
* Try as I might, I could not actually replicate the problem so I've done my best to simulate it in the test suite by intercepting the [claim staking query](https://github.com/woocommerce/action-scheduler/blob/3.1.6/classes/data-stores/ActionScheduler_DBStore.php#L637-L646) and modifying it so that it also claims future actions. My thinking is this replicates a situation where the DBStore builds a query that should ignore actions that are not yet due, but where MySQL returns them anyway).

### DIFF
--- a/classes/data-stores/ActionScheduler_DBStore.php
+++ b/classes/data-stores/ActionScheduler_DBStore.php
@@ -9,6 +9,13 @@
  */
 class ActionScheduler_DBStore extends ActionScheduler_Store {
 
+	/**
+	 * Used to share information about the before_date property of claims internally.
+	 *
+	 * @var array
+	 */
+	private $before_dates_by_claim_id = array();
+
 	/** @var int */
 	protected static $max_args_length = 8000;
 
@@ -596,8 +603,11 @@ class ActionScheduler_DBStore extends ActionScheduler_Store {
 	 */
 	public function stake_claim( $max_actions = 10, \DateTime $before_date = null, $hooks = array(), $group = '' ) {
 		$claim_id = $this->generate_claim_id();
+
+		$this->before_dates_by_claim_id[ $claim_id ] = $before_date;
 		$this->claim_actions( $claim_id, $max_actions, $before_date, $hooks, $group );
-		$action_ids = $this->find_actions_by_claim_id( $claim_id, $before_date );
+		$action_ids = $this->find_actions_by_claim_id( $claim_id );
+		unset( $this->before_dates_by_claim_id[ $claim_id ] );
 
 		return new ActionScheduler_ActionClaim( $claim_id, $action_ids );
 	}
@@ -711,19 +721,15 @@ class ActionScheduler_DBStore extends ActionScheduler_Store {
 	/**
 	 * Retrieve the action IDs of action in a claim.
 	 *
-	 * @param string $claim_id Claim ID.
-	 * @param DateTime $before_date Should use UTC timezone.
-	 *
 	 * @return int[]
 	 */
-	public function find_actions_by_claim_id( $claim_id, DateTime $before_date = null ) {
+	public function find_actions_by_claim_id( $claim_id ) {
 		/** @var \wpdb $wpdb */
 		global $wpdb;
 
-		$action_ids = array();
-		$now        = as_get_datetime_object();
-		$date       = is_null( $before_date ) ? $now : clone $before_date;
-		$cut_off    = $date->format( 'Y-m-d H:i:s' );
+		$action_ids  = array();
+		$before_date = isset( $this->before_dates_by_claim_id[ $claim_id ] ) ? $this->before_dates_by_claim_id[ $claim_id ] : as_get_datetime_object();
+		$cut_off     = $before_date->format( 'Y-m-d H:i:s' );
 
 		$sql = $wpdb->prepare(
 			"SELECT action_id, scheduled_date_gmt FROM {$wpdb->actionscheduler_actions} WHERE claim_id = %d",
--- a/classes/data-stores/ActionScheduler_DBStore.php
+++ b/classes/data-stores/ActionScheduler_DBStore.php
@@ -15,7 +15,7 @@ class ActionScheduler_DBStore extends ActionScheduler_Store {
 	 * This is used in preference to passing the same information as a method param
 	 * for backwards-compatibility reasons.
 	 *
-	 * @var ?DateTime
+	 * @var DateTime|null
 	 */
 	private $claim_before_date = null;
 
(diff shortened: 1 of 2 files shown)


rrennick:
> under unusual conditions, such as if MySQL finds it is out of disk space—the WHERE conditions we use to enforce this are seemingly not respected.

The specific situation where this was observed is that MySQL was out of temporary disk space (which is required when joining tables). When that occurred during a claim query, MySQL ignored the WHERE clause which requires a join and retrieved action IDs according to the order of the PRIMARY KEY and the LIMIT clause of the query.

classes/data-stores/ActionScheduler_DBStore.php
```diff
@@ -9,6 +9,13 @@
  */
 class ActionScheduler_DBStore extends ActionScheduler_Store {
 
+	/**
+	 * Used to share information about the before_date property of claims internally.
+	 *
+	 * @var array
+	 */
+	private $before_dates_by_claim_id = array();
```
barryhughes, line 17: :memo: I'm using a private method to share this information rather than modify any of the method signatures (this is for backwards compatibility reasons...if we altered some of the methods to add a new optional param, and passed the information that way, then under PHP 8.0+ we'd face errors if third-party sub-classes didn't update in time).

claudiosanches (approved):
@barryhughes Great work solving this issue, special kudos for the unit test introduced, nice way to try to reproduce it. I agree that's a hard one to reproduce.

@rrennick do you like to also review this change?

classes/data-stores/ActionScheduler_DBStore.php
```diff
@@ -596,8 +603,11 @@ protected function get_date_gmt( $action_id ) {
 	 */
 	public function stake_claim( $max_actions = 10, \DateTime $before_date = null, $hooks = array(), $group = '' ) {
 		$claim_id = $this->generate_claim_id();
+
+		$this->before_dates_by_claim_id[ $claim_id ] = $before_date;
 		$this->claim_actions( $claim_id, $max_actions, $before_date, $hooks, $group );
 		$action_ids = $this->find_actions_by_claim_id( $claim_id );
+		unset( $this->before_dates_by_claim_id[ $claim_id ] );
```
rrennick, line 610: @barryhughes This looks good to me. A small nitpick is that `before_dates_by_claim_id` will never have more than one element in the array.

barryhughes: You're right, I over-complicated things there. Will simplify down to a scalar :+1:

classes/data-stores/ActionScheduler_DBStore.php
```diff
@@ -596,8 +606,11 @@ protected function get_date_gmt( $action_id ) {
 	 */
 	public function stake_claim( $max_actions = 10, \DateTime $before_date = null, $hooks = array(), $group = '' ) {
 		$claim_id = $this->generate_claim_id();
+
+		$this->claim_before_date = $before_date;
 		$this->claim_actions( $claim_id, $max_actions, $before_date, $hooks, $group );
 		$action_ids = $this->find_actions_by_claim_id( $claim_id );
+		unset( $this->claim_before_date );
```
claudiosanches, line 613: We could just assign to `null` again.

classes/data-stores/ActionScheduler_DBStore.php
```diff
@@ -9,6 +9,16 @@
  */
 class ActionScheduler_DBStore extends ActionScheduler_Store {
 
+	/**
+	 * Used to share information about the before_date property of claims internally.
+	 *
+	 * This is used in preference to passing the same information as a method param
+	 * for backwards-compatibility reasons.
+	 *
+	 * @var ?DateTime
```
claudiosanches, line 18:
```suggestion
	 * @var DateTime|null
```

claudiosanches (changes requested):
I left some notes about small tweaks that we could do in this PR.

barryhughes:
Updated, thank you!

claudiosanches (approved):
Looks great!

rrennick:
Looks great for me too. Okay to merge once it clears Travis.

PR #671, opened 2021-03-03
Add secondary scheduled date checks when claiming actions (wpPostStore) | #634

### LINKED ISSUE #634: Double check action scheduled date in case they're processing too early

barryhughes:
:bulb: **This is essentially a replica of [a previous PR](https://github.com/woocommerce/action-scheduler/pull/668)** → please refer back to that for a description of the problem and how we're solving it. In this PR, I'm basically taking the same approach used for the `DBStore`, but have ported it to the `wpPostStore`. Closes #634.

#### Changelog

> Fix - Prevents pending actions that are not yet due from being claimed prematurely.

### DIFF
--- a/classes/data-stores/ActionScheduler_wpPostStore.php
+++ b/classes/data-stores/ActionScheduler_wpPostStore.php
@@ -9,6 +9,16 @@ class ActionScheduler_wpPostStore extends ActionScheduler_Store {
 	const SCHEDULE_META_KEY = '_action_manager_schedule';
 	const DEPENDENCIES_MET = 'as-post-store-dependencies-met';
 
+	/**
+	 * Used to share information about the before_date property of claims internally.
+	 *
+	 * This is used in preference to passing the same information as a method param
+	 * for backwards-compatibility reasons.
+	 *
+	 * @var DateTime|null
+	 */
+	private $claim_before_date = null;
+
 	/** @var DateTimeZone */
 	protected $local_timezone = NULL;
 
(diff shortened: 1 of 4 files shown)


claudiosanches (approved):
Looks great!

claudiosanches (approved):
Everything looks good, just waiting for Travis before we merge it.

barryhughes:
Yar, seems like 2 of the 31 jobs have failed.

|Job|Note|
|---|---|
|PHP 5.3 \| WP 5.1 \| Single | :x: [test_claim_filters_out_unexpected_future_actions](https://github.com/woocommerce/action-scheduler/blob/fix/634-date-checks--wp-post-store/tests/phpunit/runner/ActionScheduler_QueueRunner_Test.php#L173) |
|PHP 5.3 \| WP 5.1 \| Multisite | :x: [test_claim_filters_out_unexpected_future_actions](https://github.com/woocommerce/action-scheduler/blob/fix/634-date-checks--wp-post-store/tests/phpunit/runner/ActionScheduler_QueueRunner_Test.php#L173) |

Not immediately sure why only that test would flop for those specific environments...triggered a re-run of tests for this branch and will see if this is consistent.

barryhughes:
[Another failing build](https://travis-ci.com/github/woocommerce/action-scheduler/jobs/489482900) :cry: ... pretty confident it is unrelated to the change made here, though (just another case of the same problem Ron and I [touched on here](https://github.com/woocommerce/action-scheduler/pull/671#issuecomment-793231739)). Going to trigger a re-run.

claudiosanches:
Tests are passing now, I agree that we could improve it later, but I'll merge this one for now.

Issue #641, opened 2020-11-12
WP-CLI command support to create missing AS related DB tables


lukecav:
There should be a way to force missing Action Scheduler DB tables from WP-CLI.

https://actionscheduler.org/wp-cli/

roykho:
> force missing Action Scheduler DB tables

That or have logic in place to just create the tables before anything runs.

barryhughes:
We forgot to close this after merging #676. Closing now!

PR #676, opened 2021-03-17
adding CLI command to force table creation for Action Scheduler

ovidiul:
This PR addresses #641 by adding a CLI command named "fix-schema" that will run the `register_tables` forcefully for each schema class that extends `ActionScheduler_Abstract_Schema`

This will allow future schema addon classes to be included automatically in the fix.


classes/WP_CLI/ActionScheduler_WPCLI_Scheduler_command.php
```diff
@@ -5,6 +5,37 @@
...
+	 *
+	 * @param array $args Positional arguments.
+	 * @param array $assoc_args Keyed arguments.
+	 * @throws \WP_CLI\ExitException When an error occurs.
+	 *
+	 * @subcommand fix-schema
+	 */
+	public function fix_schema( $args, $assoc_args ) {
+		$classes                 = get_declared_classes();
+		$action_scheduler_schema = array();
```
barryhughes, line 21: I think we can remove this one (doesn't look like it's used).

classes/WP_CLI/ActionScheduler_WPCLI_Scheduler_command.php
```diff
@@ -5,6 +5,37 @@
...
+
+		foreach ( $classes as $classname ) {
+			if ( is_subclass_of( $classname, 'ActionScheduler_Abstract_Schema' ) ) {
+				$obj = new $classname();
+				$obj->register_tables( true );
+
+				WP_CLI::success(
+					sprintf(
+						/* translators: %d refers to the total number of taskes completed */
```
barryhughes, line 30: Fix the translator note here? (In the string on the next line we're using `%s` rather than `%d`, and it represents the schema name rather than total tasks completed.)

barryhughes (changes requested):
Thanks so much for submitting this!

I added some inline notes, but in addition I wanted to note that we already have a method which performs much the same task (it's tucked away in the [List Table class](https://github.com/woocommerce/action-scheduler/blob/3.1.6/classes/ActionScheduler_ListTable.php#L502-L519)). With that in mind, rather than reinvent things here, I'd much rather we have a single implementation (that is accessible and can be called from both places).

We could also make that new/relocated logic return an indication of success.

classes/WP_CLI/ActionScheduler_WPCLI_Scheduler_command.php
barryhughes: (Just to be super clear, I mean the `$action_scheduler_schema` definition.)

classes/WP_CLI/ActionScheduler_WPCLI_Scheduler_command.php
```diff
@@ -5,6 +5,37 @@
...
+	 * @subcommand fix-schema
+	 */
+	public function fix_schema( $args, $assoc_args ) {
+		$classes                 = get_declared_classes();
+		$action_scheduler_schema = array();
+
+		foreach ( $classes as $classname ) {
+			if ( is_subclass_of( $classname, 'ActionScheduler_Abstract_Schema' ) ) {
+				$obj = new $classname();
+				$obj->register_tables( true );
```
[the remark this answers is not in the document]
rrennick: @vedanshujain Do you recall the discussion we had on confirming `dbDelta()` success last year? That might be useful here.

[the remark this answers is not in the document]
vedanshujain: This function is eventually calling `dbDelta`. In WC Core, we added the workaround of calling `dbDelta()` twice, and if there are still tables to update on the second consecutive call, then it means that table registration failed, and then we show a notice to the user.

Seems like we can solve here using a similar approach?

peterfabian:
Is this one still relevant, given that @barryhughes noted the following
> I wanted to note that we already have a method which performs much the same task (it's tucked away in the List Table class).

?

Thx!

barryhughes:
I think this could still be a useful addition to Action Scheduler, just needs a little bit of polish (and perhaps a rebase, given it's been dormant for a while). To summarize what's needed:

- Biggest thing would be to find a way to [reuse/share this existing logic](https://github.com/woocommerce/action-scheduler/blob/3.1.6/classes/ActionScheduler_ListTable.php#L502-L519) rather than have essentially the same logic in two different spots.
    - If we do this, we don't need to worry about cleaning up [this unused var](https://github.com/woocommerce/action-scheduler/pull/676#discussion_r616826610).
    - We also wouldn't need to worry about this [efficiency note](https://github.com/woocommerce/action-scheduler/pull/676#discussion_r616799089).
- Second biggest thing is the concern that [we don't know if the operation was successful or not](https://github.com/woocommerce/action-scheduler/pull/676#discussion_r616806514), but we always [report success](https://github.com/woocommerce/action-scheduler/pull/676/files#diff-00ef471df7bb761dd546e454f33dc6c7c1cf515491d629897615d786358beb18R28).
    - There is of course a more complete approach we could take to actually verify if table creation was successful and we could then report back appropriately...
    - ...Or, I think in general it would be probably be fine (and still a net improvement on what we have today) to simply conclude with something like, _'Operation complete. Please verify by running "wp db tables".'_

On the other hand, probably fine to close if there's no further interest in it. @ovidiul what do you think?

ovidiul:
@barryhughes  Would it make sense to reuse the logic from the  ActionScheduler_ListTable class in [display_admin_notices](https://github.com/woocommerce/action-scheduler/blob/295d5776ce8cb34bb0f1d424c472086b7245ccb7/classes/ActionScheduler_ListTable.php#L352-L371), and maybe put that in a separate class/method for reuse? 

This will solve the code reuse issue and also allow us to check that the actual tables were created and notify users accordingly at CLI level?

classes/WP_CLI/ActionScheduler_WPCLI_Scheduler_command.php
```diff
@@ -5,6 +5,37 @@
...
 class ActionScheduler_WPCLI_Scheduler_command extends WP_CLI_Command {
 
+	/**
+	 * Force tables schema creation for Action Scheduler
+	 *
+	 * ## OPTIONS
+	 *
+	 * @param array $args Positional arguments.
+	 * @param array $assoc_args Keyed arguments.
+	 * @throws \WP_CLI\ExitException When an error occurs.
```
barryhughes, line 15:
```suggestion
```

I don't think this exception can be thrown by this method (the only call into WP CLI is to `WP_CLI::success()` which won't trigger this).

classes/WP_CLI/ActionScheduler_WPCLI_Scheduler_command.php
```diff
@@ -5,6 +5,37 @@
...
+		foreach ( $schema_classes as $classname ) {
+			if ( is_subclass_of( $classname, 'ActionScheduler_Abstract_Schema' ) ) {
+				$obj = new $classname();
+				$obj->init();
+				$obj->register_tables( true );
+
+				WP_CLI::success(
+					sprintf(
+						/* translators: %s refers to the schema name*/
+						__( 'Registered schema for %s', 'action-sheduler' ),
```
barryhughes, line 31:
```suggestion
						__( 'Registered schema for %s', 'action-sheduler' ),
```
```suggestion
						__( 'Registered schema for %s', 'action-scheduler' ),
```

Typo in the text domain.

classes/WP_CLI/ActionScheduler_WPCLI_Scheduler_command.php
```diff
@@ -5,6 +5,37 @@
...
+	 * @param array $assoc_args Keyed arguments.
+	 * @throws \WP_CLI\ExitException When an error occurs.
+	 *
+	 * @subcommand fix-schema
+	 */
+	public function fix_schema( $args, $assoc_args ) {
+		$schema_classes = array( 'ActionScheduler_LoggerSchema', 'ActionScheduler_StoreSchema' );
+
+		foreach ( $schema_classes as $classname ) {
+			if ( is_subclass_of( $classname, 'ActionScheduler_Abstract_Schema' ) ) {
```
barryhughes, line 23:
```suggestion
		$schema_classes = array( ActionScheduler_LoggerSchema::class, ActionScheduler_StoreSchema::class );

		foreach ( $schema_classes as $classname ) {
			if ( is_subclass_of( $classname, ActionScheduler_Abstract_Schema::class ) ) {
```

Optional—but this would provide a couple of benefits especially in IDEs (easy navigation to the actual classes, and during future refactoring, etc).

barryhughes (changes requested):
Nice, couple small tweaks and I expect this will be good to merge :+1:

ovidiul:
> Nice, couple small tweaks and I expect this will be good to merge 👍

Thanks for the suggestions @barryhughes , they are now merged.

barryhughes (approved):
LGTM, thanks!

barryhughes:
✏️ We should also update [this page](https://actionscheduler.org/wp-cli/) with notes about this new command.

PR #702, opened 2021-06-01
Update pending dependencies

vedanshujain:
Tested out by:
1. Running unit tests.
2. Running multiple batches using CLI.

These are all dev dependencies, so hopefully will not have an impact in prod anyway.


rrennick (approved):
@vedanshujain LGTM

PR #703, opened 2021-06-01
Version bump for 3.2.0.

vedanshujain:


claudiosanches (approved):
Works great, also everything looks good.

action-scheduler.php
```diff
@@ -25,28 +25,28 @@
...
 	}
 
-	add_action( 'plugins_loaded', 'action_scheduler_register_3_dot_1_dot_6', 0, 0 );
+	add_action( 'plugins_loaded', 'action_scheduler_register_3_dot_2_dot_0', 0, 0 );
 
-	function action_scheduler_register_3_dot_1_dot_6() {
+	function action_scheduler_register_3_dot_2_dot_0() {
 		$versions = ActionScheduler_Versions::instance();
-		$versions->register( '3.1.6', 'action_scheduler_initialize_3_dot_1_dot_6' );
+		$versions->register( '3.1.6', 'action_scheduler_initialize_3_dot_2_dot_0' );
```
rrennick, line 39: `$versions->register( '3.1.6'` 
should be 
`$versions->register( '3.2.0'`

The version that is loaded is determined by sorting registered version descending. The first callback for a registered version is the one that is used to load it if it is the highest verson.

rrennick (changes requested):
@vedanshujain There is also a [readme.txt](https://plugins.trac.wordpress.org/browser/action-scheduler/trunk/readme.txt) on WordPress.org that should be added to the repo. I apologize for not doing that when I released 3.1.6.

action-scheduler.php
vedanshujain: Nice catch, added in https://github.com/woocommerce/action-scheduler/pull/703/commits/2a43795a99bebff2cd5fcc82ad26d3bc0330b8f1.

vedanshujain:
@rrennick ready for another round, can you take a look when you get a chance? Also generated this zip using woorelease for smoke testing if needed. 

[action-scheduler.zip](https://github.com/woocommerce/action-scheduler/files/6584056/action-scheduler.zip)

rrennick:
@vedanshujain The changes look good and the zip worked fine in my testing. 

The `build` script creates a zip that doesn't include the `action-scheduler` folder. That's not critical for release but I think it will interfere with the `wr release` which will be expecting the folder.

vedanshujain:
@rrennick added a command to do wrap it in a folder, it looks complicated but seems like this is what we do for many extensions as well.

[action-scheduler.zip](https://github.com/woocommerce/action-scheduler/files/6584680/action-scheduler.zip)

rrennick (approved):

rrennick:
> added a command to do wrap it in a folder

@vedanshujain This worked great and is good to go for me 🚢

PR #705, opened 2021-06-03
Update stable version.

vedanshujain:
Last thing pending from releasing 3.2.0


barryhughes:
Looks good! While we're here, though, should we also bump the _"Tested up to"_ field (to 5.7.2)?

barryhughes (approved):
:+1:

PR #715, opened 2021-06-17
Add extra safety/account for different versions of AS and different loading patterns

### LINKED ISSUE #714: ActionScheduler can be declared twice (causing fatal)
## Problem

If two plugins both include Action Scheduler (but different versions) and if one uses the standard approach of including Action Scheduler from top-level code in the main plugin file, and the other uses the alternative approach of doing this during `plugins_loaded:-10` then we end up with a fatal error:

`Cannot declare class ActionScheduler, because the name is already in use`

## Steps to replicate

* Activate WooCommerce (current trunk code) with AS 3.2.0.
* Activate another plugin that includes AS 3.1.6. This plugin should load Action Scheduler during `plugins_loaded` (priority `-10`), which is a strategy we document [over here](https://actionscheduler.org/usage/#load-order).
    * This plugin should be named something like `z-plugin` to make it load _after_ WooCommerce.
    * Note also that in the PR where this problem was originally reported, the reporter suggests we revise those load order docs: that's worth consideration.
* Try to access your WordPress site, observe the fatal error.

## Why does it happen?

📝 In an attempt at clarity, I'm using 'bootloader' and 'bootloading' to refer to the logic [living in `action-scheduler/action-scheduler.php`](https://github.com/woocommerce/action-scheduler/blob/3.1.6/action-scheduler.php) and 'loading' to mean the moment when [`classes/abstracts/ActionScheduler.php` is actually loaded](https://github.com/woocommerce/action-scheduler/blob/275d0ba54b1c263dfc62688de2fa9a25a373edf8/classes/abstracts/ActionScheduler.php).

* In this scenario, 3.2.0 (included in WooCommerce) starts bootstrapping first of all.
    * It doesn't [register itself](https://github.com/woocommerce/action-scheduler/blob/3.1.6/action-scheduler.php#L42-L45) until `plugins_loaded:0`.
    * Therefore, the `ActionScheduler` class has not yet been defined.
* Our custom plugin loads next but waits until `plugins_loaded:-10` to start bootstrapping Action Scheduler (version 3.1.6). Once that takes place, the [3.1.6 bootloader](https://github.com/woocommerce/action-scheduler/blob/3.1.6/action-scheduler.php) is called.
    * We have a chunk of code that [handles loading AS in the context of a theme](https://github.com/woocommerce/action-scheduler/blob/3.1.6/action-scheduler.php#L47-L52), rather than a plugin.
    * As part of this, it checks if `did_action( 'plugins_loaded' )` and I assume the intent is to stop that block of code executing until some time _after_ `plugins_loaded` has finished...but all this really means is `do_action('plugins_loaded`) was called, not that we've moved past it. In this case, we're still 'inside' `plugins_loaded`.
    * This leads to [Action Scheduler 3.1.6 being loaded](https://github.com/woocommerce/action-scheduler/blob/3.1.6/action-scheduler.php#L42-L45)
* The problem occurs when 3.2.0's callback (noted above) fires, which happens next of all (`plugins_loaded:0`)
    * It registers version 3.2.0, and `ActionScheduler_Versions` correctly identifies that as being the most up-to-date version of AS available, and goes ahead and [tries to load that version](https://github.com/woocommerce/action-scheduler/blob/3.2.0/action-scheduler.php#L42-L45) which is where things come crashing down, because `ActionScheduler` is already defined.

## Props

Props for the initial report and some solid findings go to @capuderg, [(see this PR)](https://github.com/woocommerce/action-scheduler/pull/713).

barryhughes:
A more complete breakdown of the problem can be found in #714 but, in essence, if two plugins are present and both include different versions of Action Scheduler (3.2.0 and 3.1.6, for example) and each uses a different loading pattern, a fatal error can occur.

`Cannot declare class ActionScheduler, because the name is already in use`

Closes #714.

## Steps to replicate

* Activate WooCommerce (current trunk code) with AS 3.2.0.
* Activate another plugin that includes AS 3.1.6. This plugin should load Action Scheduler during `plugins_loaded` (priority `-10`), which is a strategy we document [over here](https://actionscheduler.org/usage/#load-order).
    * This plugin should be named something like `z-plugin` to make it load _after_ WooCommerce.
* Try to access your WordPress site, observe the fatal error.

## Testing

* Make sure you can replicate the original problem (see also [PR 713](https://github.com/woocommerce/action-scheduler/pull/713) for additional context).
* With respect to the above steps to replicate, replace the version of Action Scheduler loaded by WooCommerce with this updated version.
* Problem should be resolved.

## Notes

* An alternative to this solution could be modifying `ActionScheduler_Versions::initialize_latest_version()` and placing an extra safety check there.

## Changelog

> Fix - Resolves a potential error when multiple versions of Action Scheduler are loaded by different plugins.

### DIFF
--- a/action-scheduler.php
+++ b/action-scheduler.php
@@ -40,12 +40,17 @@ if ( ! function_exists( 'action_scheduler_register_3_dot_2_dot_0' ) ) {
 	}
 
 	function action_scheduler_initialize_3_dot_2_dot_0() {
-		require_once( 'classes/abstracts/ActionScheduler.php' );
-		ActionScheduler::init( __FILE__ );
+		// A final safety check is required even here, because historic versions of Action Scheduler
+		// followed a different pattern (in some unusual cases, we could reach this point and the
+		// ActionScheduler class is already defined—so we need to guard against that).
+		if ( class_exists( 'ActionScheduler' ) ) {
+			require_once( 'classes/abstracts/ActionScheduler.php' );
+			ActionScheduler::init( __FILE__ );
+		}
 	}
 
 	// Support usage in themes - load this version if no plugin has loaded a version yet.
-	if ( did_action( 'plugins_loaded' ) && ! class_exists( 'ActionScheduler' ) ) {
+	if ( did_action( 'plugins_loaded' ) && ! doing_action( 'plugins_loaded' ) && ! class_exists( 'ActionScheduler' ) ) {
 		action_scheduler_initialize_3_dot_2_dot_0();
 		do_action( 'action_scheduler_pre_theme_init' );
 		ActionScheduler_Versions::initialize_latest_version();
--- a/action-scheduler.php
+++ b/action-scheduler.php
@@ -43,7 +43,7 @@ if ( ! function_exists( 'action_scheduler_register_3_dot_2_dot_0' ) ) {
 		// A final safety check is required even here, because historic versions of Action Scheduler
 		// followed a different pattern (in some unusual cases, we could reach this point and the
 		// ActionScheduler class is already defined—so we need to guard against that).
-		if ( class_exists( 'ActionScheduler' ) ) {
+		if ( ! class_exists( 'ActionScheduler' ) ) {
 			require_once( 'classes/abstracts/ActionScheduler.php' );
 			ActionScheduler::init( __FILE__ );
 		}


action-scheduler.php
```diff
@@ -40,12 +40,17 @@ function action_scheduler_register_3_dot_2_dot_0() {
...
+		// ActionScheduler class is already defined—so we need to guard against that).
+		if ( class_exists( 'ActionScheduler' ) ) {
+			require_once( 'classes/abstracts/ActionScheduler.php' );
+			ActionScheduler::init( __FILE__ );
+		}
 	}
 
 	// Support usage in themes - load this version if no plugin has loaded a version yet.
-	if ( did_action( 'plugins_loaded' ) && ! class_exists( 'ActionScheduler' ) ) {
+	if ( did_action( 'plugins_loaded' ) && ! doing_action( 'plugins_loaded' ) && ! class_exists( 'ActionScheduler' ) ) {
```
barryhughes, line 53: I'm adding this because `did_action( 'plugins_loaded' )` really only tells us that `do_action( 'plugins_loaded' )` was called. In our case, we want to know not only that that took place, but that we've moved past it.

barryhughes:
Placing back in code review, even though there is a Travis CI failure (it's a timestamp comparison issue, unrelated to this change, and re-running the checks is just as likely to result in it showing up someplace else—see [this issue](https://github.com/woocommerce/action-scheduler/issues/717)).

claudiosanches (approved):
Great work!
I could reproduce and works as described.

PR #719, opened 2021-06-18
Remove zip file

claudiosanches:
A zip file got added by accident after https://github.com/woocommerce/action-scheduler/pull/703 got merged.
This zip file will also get deployed to WP.org if we don't exclude it.

This PR just removes that file and prevent it to happens again by updating our `.gitignore`.


roykho (approved):
LGTM!

PR #721, opened 2021-06-21
Release/3.2.1

barryhughes:
Prep for 3.2.1 release (changelog will be populated by WooRelease, and will merge that in after release).

### DIFF
--- a/action-scheduler.php
+++ b/action-scheduler.php
@@ -5,7 +5,7 @@
  * Description: A robust scheduling library for use in WordPress plugins.
  * Author: Automattic
  * Author URI: https://automattic.com/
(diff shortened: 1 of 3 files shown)


roykho (approved):
Curious as to why these functions needs to be named with the version suffixes? `package.json` version did not get bumped but I think if you're using Woorelease, it should handle that. But just pointing it out just in case. I will let you merge this since you still need to add the changelogs. Otherwise LGTM!

barryhughes:
Cool—thanks!

On the version suffixes, I believe because it's so that if 2 or more plugins are all present and all ship the same version of Action Scheduler, only the first of those tries to register itself.

PR #722, opened 2021-06-23
Suppress warnings re safe_mode. #656

### LINKED ISSUE #656: Php 7.4 compatibility issue
Hi,

I am using version 'dev-release-225-amends' of this plugin (prospress/action-scheduler) . Php 7.4 compatibility check reports an error
' INI directive 'safe_mode' is deprecated since PHP 5.3 and removed since PHP 5.4' in '/classes/ActionScheduler_Compatibility.php' line 95. Is it safe to upgrade to 7.4 with this error? or do you have any plans to update the plugin to be php 7.4 compatible?

Can you please help?

thanks,
Radha

barryhughes:
Minor, adds a `phpcs:ignore` rule to suppress warnings about the deprecation of `safe_mode`. Closes #656.

### Testing

* Execute `composer run phpcs classes/ActionScheduler_Compatibility.php`
* Look for `95 | WARNING | INI directive 'safe_mode' is deprecated since PHP 5.3 and removed since PHP 5.4...`
* The above should be gone as of this change.

### Notes

* We could also consider removing the relevant `ini_get()` statement _but_ ostensibly we still support as far back as PHP 5.3 😱 (in terms of plugin readme's PHP declaration and even in terms of CI coverage), for which `safe_mode` is potentially still relevant.
* Could reconsider this status, but that's a bigger/broader discussion for another time, probably.

### DIFF
--- a/classes/ActionScheduler_Compatibility.php
+++ b/classes/ActionScheduler_Compatibility.php
@@ -92,7 +92,7 @@ class ActionScheduler_Compatibility {
 
 		if ( function_exists( 'wc_set_time_limit' ) ) {
 			wc_set_time_limit( $limit );
-		} elseif ( function_exists( 'set_time_limit' ) && false === strpos( ini_get( 'disable_functions' ), 'set_time_limit' ) && ! ini_get( 'safe_mode' ) ) {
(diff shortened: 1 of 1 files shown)


vedanshujain (approved):
LGTM! Let's change the support to PHP 5.6 at least, which is what WP supports from the next version.

PR #622, opened 2020-09-21
Remove the NOT NULL from date fields

rrennick:
Closes #587 

This PR
- updates the table schemas to remove the `NOT NULL` on `datetime` fields
- adds an update routine to remove the `NOT NULL` from `datetime` fields in existing installs

The update routine is necessary because `dbDelta()` fails on the tables with `NOT NULL default '0000-00-00 00:00:00` column definitions in MySQL 5.7 configured with `sql_mode=STRICT_TRANS_TABLES,NO_ZERO_DATE`.

### Testing

- Load the Tools -> Scheduled Actions screen
- In MySQL
```
> SHOW CREATE TABLE wp_actionscheduler_actions;
wp_actionscheduler_actions | CREATE TABLE `wp_actionscheduler_actions` (
  `action_id` bigint(20) unsigned NOT NULL AUTO_INCREMENT,
  `hook` varchar(191) COLLATE utf8mb4_unicode_520_ci NOT NULL,
  `status` varchar(20) COLLATE utf8mb4_unicode_520_ci NOT NULL,
  `scheduled_date_gmt` datetime DEFAULT '0000-00-00 00:00:00',
  `scheduled_date_local` datetime DEFAULT '0000-00-00 00:00:00',
  `args` varchar(191) COLLATE utf8mb4_unicode_520_ci DEFAULT NULL,
  `schedule` longtext COLLATE utf8mb4_unicode_520_ci,
  `group_id` bigint(20) unsigned NOT NULL DEFAULT '0',
  `attempts` int(11) NOT NULL DEFAULT '0',
  `last_attempt_gmt` datetime DEFAULT '0000-00-00 00:00:00',
  `last_attempt_local` datetime DEFAULT '0000-00-00 00:00:00',
  `claim_id` bigint(20) unsigned NOT NULL DEFAULT '0',
  `extended_args` varchar(8000) COLLATE utf8mb4_unicode_520_ci DEFAULT NULL,
  PRIMARY KEY (`action_id`),
  KEY `hook` (`hook`),
  KEY `status` (`status`),
  KEY `scheduled_date_gmt` (`scheduled_date_gmt`),
  KEY `args` (`args`),
  KEY `group_id` (`group_id`),
  KEY `last_attempt_gmt` (`last_attempt_gmt`),
  KEY `claim_id` (`claim_id`)
) ENGINE=InnoDB AUTO_INCREMENT=18003 DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_520_ci

[...]

PR #725, opened 2021-06-29
Remove the NOT NULL from date fields

### LINKED ISSUE #519: Error adding action with NO_ZERO_DATE
This is a follow up issue to #503, #437. My dev install MySQL 5.7 is still configured with

```
sql_mode=STRICT_TRANS_TABLES,NO_ZERO_DATE
```

Using `as_schedule_single_action()` generates the following error

```
[09-Apr-2020 16:27:33 UTC] WordPress database error Incorrect datetime value: '0000-00-00 00:00:00' for column 'last_attempt_gmt' at row 1 for query INSERT INTO `wp_actionscheduler_actions` (`hook`, `status`, `scheduled_date_gmt`, `scheduled_date_local`, `schedule`, `group_id`, `last_attempt_gmt`, `last_attempt_local`, `args`) VALUES ('as-db-test', 'pending', '2020-04-09 17:27:33', '2020-04-09 14:27:33', 'O:30:\"ActionScheduler_SimpleSchedule\":2:{s:22:\"\0*\0scheduled_timestamp\";i:1586453253;s:41:\"\0ActionScheduler_SimpleSchedule\0timestamp\";i:1586453253;}', '0', '0000-00-00 00:00:00', '0000-00-00 00:00:00', '[]')
```

I tested using `1000-01-01 00:00:00` based on https://stackoverflow.com/questions/36374335/error-in-mysql-when-setting-default-value-for-date-or-datetime for both attempt date fields and that eliminated the error.

From https://github.com/woocommerce/action-scheduler/pull/503#issuecomment-606338844

> I'm pretty sure I've seen 0000-00-00 00:00:00 in the posts table before. Maybe for auto-draft posts?

In my install with the Gutenberg editor enabled, an `auto-draft` post isn't created and there are no errors logged when creating a new post.

barryhughes:
This is a re-play of [PR 622](https://github.com/woocommerce/action-scheduler/pull/622) (authored by Ron) and already tested/approved. 

However, it was merged into a release branch that was originally intended to be the basis of the `3.2.0` release (which actually shipped already, but with a different set of changes). That release branch contains other changes we don't want/need here and so this PR:

* Is a cherry-pick of Ron's work relating *specifically to the problem described in [#519](https://github.com/woocommerce/action-scheduler/issues/519)* (therefore, please see [the original PR for testing instructions](https://github.com/woocommerce/action-scheduler/pull/622))
* The only modifications to Ron's work is to the DBStore schema version (as of 3.2.1 we are already at [schema version `4`](https://github.com/woocommerce/action-scheduler/blob/3.2.1/classes/schema/ActionScheduler_StoreSchema.php#L18), so I bumped it up to `5`)—the LoggerSchema version is fine as is

---

### Other notes

* In addition to the instructions in [PR 622](https://github.com/woocommerce/action-scheduler/pull/622), remember that if you have to test more than once you will need to reset the following options:
    * `schema-ActionScheduler_LoggerSchema` (set to `2`)
    * `schema-ActionScheduler_StoreSchema` (set to `4`)
* Because it's a schema change, this will need to ship as a major release (so, probably `3.3.0`)

Closes #519.

### DIFF
--- a/classes/data-stores/ActionScheduler_DBLogger.php
+++ b/classes/data-stores/ActionScheduler_DBLogger.php
@@ -68,7 +68,11 @@ class ActionScheduler_DBLogger extends ActionScheduler_Logger {
 			return new ActionScheduler_NullLogEntry();
 		}
 
-		$date = as_get_datetime_object( $record->log_date_gmt );
+		if ( is_null( $record->log_date_gmt ) ) {
+			$date = as_get_datetime_object( ActionScheduler_StoreSchema::DEFAULT_DATE );
+		} else {
+			$date = as_get_datetime_object( $record->log_date_gmt );
+		}
 
 		return new ActionScheduler_LogEntry( $record->action_id, $record->message, $date );
 	}
@@ -96,6 +100,7 @@ class ActionScheduler_DBLogger extends ActionScheduler_Logger {
 	 */
 	public function init() {
 
+		add_action( 'action_scheduler_before_schema_update', array( $this, 'update_logs_schema_4_0' ), 10, 2 );
 		$table_maker = new ActionScheduler_LoggerSchema();
 		$table_maker->register_tables();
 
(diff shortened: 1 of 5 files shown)


classes/data-stores/ActionScheduler_DBLogger.php
```diff
@@ -143,4 +148,31 @@ public function bulk_log_cancel_actions( $action_ids ) {
...
 	}
+
+	/**
+	 * Update the logs table schema to allow the datetime fields to be NULL.
+	 * The NOT NULL causes a conflict with some versions of MySQL configured with sql_mode=NO_ZERO_DATE.
+	 *
+	 * @param string $table Name of table being updated.
+	 * @param string $db_version The existing schema version of the table.
+	 */
+	public function update_logs_schema_4_0( $table, $db_version ) {
```
claudiosanches, line 159: I think we could move this method to a new class or file, just to avoid having this kind of method inside this class, since it could get bigger in the future.
Also this method doesn't seems to require anything from this class and could be hooked `action_scheduler_before_schema_update` from anywhere.

classes/data-stores/ActionScheduler_DBStore.php
```diff
@@ -580,6 +594,35 @@ public function get_date( $action_id ) {
...
 	}
 
+	/**
+	 * Update the actions table schema to allow the datetime fields to be NULL.
+	 * The NOT NULL causes a conflict with some versions of MySQL configured with sql_mode=NO_ZERO_DATE.
+	 *
+	 * @param string $table Name of table being updated.
+	 * @param string $db_version The existing schema version of the table.
+	 */
+	public function update_actions_schema_5_0( $table, $db_version ) {
```
claudiosanches, line 604: I think we could move this method as well.

claudiosanches (changes requested):
Great work! 
Everything looks good and seems to be working as expected.
But I think we could move the database migration methods to a new class or a file.

classes/data-stores/ActionScheduler_DBLogger.php
barryhughes: @claudiosanches what are your thoughts on moving it to the existing `schema/ActionScheduler_LoggerSchema` class (and similarly so for the corresponding `DBStore` method, except in that case we'd use `schema/ActionScheduler_StoreSchema`)?

barryhughes: ☝🏽 Updated per my last note, let me know your thoughts!

claudiosanches (approved):
Great work!

Issue #723, opened 2021-06-24
Update minimum requirements


barryhughes:
Currently we support PHP 5.3/WP 4.7, but on review we feel we can safely adopt PHP 5.6/WP 5.2 as our new minimums:

* This will provide a little more freedom to use slightly more modern functionality (and will reduce our backwards compatibility burden to some extent).
* If there are environments in which Action Scheduler is being used but where updating WP or PHP to meet these new minimums is not possible, then freezing the deployed version of the library at 3.2.0 would be a viable solution.

Additionally:

* As part of this task, we can also revise the Travis CI config accordingly.
* When this change is released, the version bump should reflect that we changed compatibility requirements.

barryhughes:
> As part of this task, we can also revise the Travis CI config accordingly.

...Or even replace it with an [action-based workflow](https://github.com/woocommerce/action-scheduler/pull/727), possibly.

Issue #717, opened 2021-06-17
Improve tests involving timestamp comparison (and reduce sporadic CI failures)


barryhughes:
We have various tests that check if two timestamps match. [Example here.](https://github.com/woocommerce/action-scheduler/blob/3.2.0/tests/phpunit/jobstore/ActionScheduler_DBStore_Test.php#L414) However, especially in Travis CI, there is potential for a discrepancy to arise (which usually doesn't exceed 1s).

This is a problem in practice because, it feels like, almost every 2nd PR has a bunch of CI failures which are sporadic in nature and clear up on re-running the checks (but may re-appear in different places).

```
...............................................................  63 / 366 ( 17%)
..........F.................................................... 126 / 366 ( 34%)
............................................................... 189 / 366 ( 51%)
............................................................... 252 / 366 ( 68%)
............................................................... 315 / 366 ( 86%)
...................................................             366 / 366 (100%)

Time: 10.16 seconds, Memory: 46.25MB

There was 1 failure:

1) ActionScheduler_DBStore_Test::test_get_run_date
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'1623948928'
+'1623948927'
```

:point_up: _The sort of output we see. This could show up for 'PHP 7.2 WP 5.1 Multisite' in one run and 'PHP 5.4 WP 5.3 Single Site' the next—it's surfaces unpredictably and so devalues our CI checks._

We should come up with a custom assertion (ie `$this->assertTimestampsAreApproximatelyEqual(...)`), or some similar solution, to better accommodate this and give us a ±1s tolerance that we can apply where relevant.

barryhughes:
:memo: [PR #727](https://github.com/woocommerce/action-scheduler/pull/727) doesn't actually solve this, but it certainly seems to reduce instances of the problem (and so it may be all we actually need).

barryhughes:
Setting to low priority...the potential for this still exists, but we encounter it far less now (it was especially prone to surfacing under Travis, because of that platform's performance characteristics...but of course we've now moved from that to GitHub Actions).

barryhughes:
See also #512.

PR #727, opened 2021-06-29
Add GitHub actions workflow for running tests for PRs.

barryhughes:
Moves us from Travis CI to GitHub Actions.

🚀 It's significantly faster than Travis (not just overall, but also per individual run). A nice side-effect of this is that it will reduce instances of the problem described #717—though it doesn't solve it altogether.

📜 Test matrix updated per #723. Instead of testing _all_ PHP versions from the minimum onwards we pick the first and last minor versions (ie, 7.0 and 7.4) of each major release, within the range we support. That means we will now test against (of course, we'll revise as needed as new versions of PHP are released):

- PHP 5.6 (first and last supported version from PHP 5.x)
- PHP 7.0 (first supported version from PHP 7.x)
- PHP 7.4 (last supported version from PHP 7.x)
- PHP 8.0 (first and last supported version from PHP 8.x)


barryhughes:
⏱️ All of the checks implemented via GitHub Actions completed **within 3 minutes of PR creation.** Over 20 minutes later, still waiting on the result to come back from Travis (recent experience tells me I can expect a delay of 15-30mins with our current Travis setup for this project).

barryhughes:
Some further tweaks/experimentation:

- Dropped Travis
- Reduced number of PHP versions in the test matrix, so we have:
    - 5.6 (which is the only 5.x version we still support)
    - 7.0 and 7.4 (first and last versions in the 7.x range)
    - 8.0 (but eventually we'd do the same as with 7.x, ie support the first and last in the 8.x range)

claudiosanches (commented):
Looks great!

claudiosanches (approved):

PR #731, opened 2021-07-14
fixes #730 issue: Call to undefined function as_next_scheduled_action()

### LINKED ISSUE #730: Call to undefined function as_next_scheduled_action()
If action-scheduler is used in a [plugin](https://wordpress.org/plugins/ethereum-wallet/) on a site running WooCommerce 5.5.0, site crashed with error `Call to undefined function as_next_scheduled_action()`

I've got 10+ reports from plugin users today :-)

olegabr:
The idea is to not load classes when checking for them. In other case class is loaded but `ActionScheduler::init` method is not called, and the next plugin or theme checking for it skips the `init` call, since class i already loaded.

### DIFF
--- a/action-scheduler.php
+++ b/action-scheduler.php
@@ -27,7 +27,7 @@
 
 if ( ! function_exists( 'action_scheduler_register_3_dot_2_dot_1' ) ) {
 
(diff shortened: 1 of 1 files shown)


barryhughes:
Merged via #776.

PR #743, opened 2021-08-11
Update stated PHP and WP minimums.

### LINKED ISSUE #723: Update minimum requirements

barryhughes:
Updates our stated minimum requirements, from WP 4.7 to WP 5.2 and from PHP 5.3 to PHP 5.6. 

This also means we can stop running Travis against those versions that are no longer supported—that's taken care of in [PR 727](https://github.com/woocommerce/action-scheduler/pull/727) (though, it's not essential that be merged first).

Closes #723.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -1,11 +1,11 @@
 === Action Scheduler ===
 Contributors: Automattic, wpmuguru, claudiosanches, peterfabian1000, vedjain, jamosova, obliviousharmony, konamiman, sadowski, royho, barryhughes-1
 Tags: scheduler, cron
-Requires at least: 4.7
+Requires at least: 5.2
 Tested up to: 5.7
 Stable tag: 3.2.1
 License: GPLv3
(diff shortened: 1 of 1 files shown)


vedanshujain (approved):
LGTM! This would be worth posting about on our blog as well!

barryhughes:
Good point!

PR #748, opened 2021-08-24
Drop index claim_id and remove Order BY to reduce/resolve deadlocks

### LINKED ISSUE #530: Diagnose intermittent deadlocks
Continuing discussion from #438 on intermittent deadlocks:

```
WordPress database error Deadlock found when trying to get lock; try restarting transaction for query UPDATE wp_actionscheduler_actions SET attempts = attempts+1, status='in-progress', last_attempt_gmt = '2020-04-20 17:00:02', last_attempt_local = '2020-04-20 18:00:02' WHERE action_id = 1877421 made by do_action('wp_ajax_as_async_request_queue_runner'), WP_Hook->do_action, WP_Hook->apply_filters, WP_Async_Request->maybe_handle, ActionScheduler_AsyncRequest_QueueRunner->handle, do_action('action_scheduler_run_queue'), WP_Hook->do_action, WP_Hook->apply_filters, ActionScheduler_QueueRunner->run, ActionScheduler_QueueRunner->do_batch, ActionScheduler_Abstract_QueueRunner->process_action, ActionScheduler_DBStore->log_execution, QM_DB->query
```
and
```
WordPress database error Deadlock found when trying to get lock; try restarting transaction for query UPDATE wp_actionscheduler_actions SET attempts = attempts+1, status='in-progress', last_attempt_gmt = '2020-04-20 17:02:08', last_attempt_local = '2020-04-20 18:02:08' WHERE action_id = 1877473 made by do_action_ref_array('action_scheduler_run_queue'), WP_Hook->do_action, WP_Hook->apply_filters, ActionScheduler_QueueRunner->run, ActionScheduler_QueueRunner->do_batch, ActionScheduler_Abstract_QueueRunner->process_action, ActionScheduler_DBStore->log_execution
```

glagonikas:
Fixes #530 

Even though the new index (`claim_id_status_scheduled_date_gmt`) was added in #696, MySQL wouldn't pick it up in our case. Instead, it would use a composite index made up of the two separate indexes `claim_id` and `status`. This leads to slower query times and therefore, deadlocks.

Any query that takes advantage of `claim_id` would take equal advantage from `claim_id_status_scheduled_date_gmt` (as long as it's doesn't include `FORCE INDEX`), so dropping the `claim_id` index should have no negative impact in the performance.

Additionally, having `ORDER BY` in the query created in `claim_actions` makes the query excessively slow and leads to deadlocks. Without `ORDER BY`, the query will still pick the oldest actions (in most cases) because it'll order them based on the `action_id` which is sequential. For most cases, dropping the `ORDER BY` will have no negative impact, but will greatly improve performance and reduce deadlocks.

### DIFF
--- a/classes/data-stores/ActionScheduler_DBStore.php
+++ b/classes/data-stores/ActionScheduler_DBStore.php
@@ -698,7 +698,7 @@ class ActionScheduler_DBStore extends ActionScheduler_Store {
 			$params[] = $group_id;
 		}
 
-		$order    = "ORDER BY attempts ASC, scheduled_date_gmt ASC, action_id ASC LIMIT %d";
+		$order    = "LIMIT %d";
 		$params[] = $limit;
 
 		$sql = $wpdb->prepare( "{$update} {$where} {$order}", $params );
--- a/classes/schema/ActionScheduler_StoreSchema.php
+++ b/classes/schema/ActionScheduler_StoreSchema.php
@@ -16,7 +16,7 @@ class ActionScheduler_StoreSchema extends ActionScheduler_Abstract_Schema {
 	/**
 	 * @var int Increment this value to trigger a schema update.
 	 */
-	protected $schema_version = 5;
+	protected $schema_version = 6;
 
 	public function __construct() {
 		$this->tables = [
@@ -64,7 +64,6 @@ class ActionScheduler_StoreSchema extends ActionScheduler_Abstract_Schema {
 				        KEY args (args($max_index_length)),
 				        KEY group_id (group_id),
 				        KEY last_attempt_gmt (last_attempt_gmt),
-				        KEY claim_id (claim_id),
 				        KEY `claim_id_status_scheduled_date_gmt` (`claim_id`, `status`, `scheduled_date_gmt`)
 				        ) $charset_collate";
 


barryhughes:
All tests were showing as failing, triggered a re-run (but it may also be worth rebasing against the latest master code).

glagonikas:
@barryhughes I'm not sure why this is? Don't see anything wrong with the proposed changes, so it might be something on the tests that i don't get?

barryhughes:
I'm not immediately sure either, but, worth pausing and figuring it out before we merge (I have limited time right now, but will definitely make a note to circle back—and others may be able to chime in before then).

classes/data-stores/ActionScheduler_DBStore.php
```diff
@@ -655,7 +655,7 @@ protected function claim_actions( $claim_id, $limit, \DateTime $before_date = nu
 			$params[] = $group_id;
 		}
 
-		$order    = "ORDER BY attempts ASC, scheduled_date_gmt ASC, action_id ASC LIMIT %d";
+		$order    = "LIMIT %d";
```
barryhughes, line 658: > Without ORDER BY, the query will still pick the oldest actions (in most cases) because it'll order them based on the action_id which is sequential.

Understood, though, creation date is not 1:1 with scheduled date. 

I'd be just a little worried about cases where honouring that commitment (to process the longest waiting actions relative to their scheduled date/time) is particularly important. There could definitely be lots of cases where it may make no difference, but on the other hand there may be projects and workloads where it is valuable.

Along with the change in table indicies, I wonder if this is something that should exist as an optional change—that is, let site operators opt-in or out of this (and there are various ways we might implement that, ranging from UI-level settings to feature flags or filters). 

Some other (but minor) issues:

- Unless we make a corresponding change to the WP Post Store, this will make the behavior of the two store implementations inconsistent with one another.
- I feel less concerned about dropping ordering on the `attempts` column but it would be interesting to hear from others on this.

Balancing all that, the potential performance improvement is clear. Will flag for further discussion.

glagonikas: I see what you mean about ordering, but i wonder if there are many real-world scenarios where this would be an issue? It would only affect stores that rely on actions running in a specific order, but their code creates them in the wrong order. It seems quite unlikely to me.


as a reminder, in our case, the deadlocks was the main issue and it seems the only way to overcome them is by altering this query. With WC/AS being unable to handle those deadlocks, this is the only viable solution and certainly beats any scheduling issues.

Maybe a middle-ground would be to add a filter and a section on docs advising people who get deadlocks to use it? At the moment, we have to hardcode this change to WC and any plugins that ship with AS which is a pain, but having a filter would permanently resolve our deadlocks without affecting other stores who are unaware of them or desperately need them firing in this order?

barryhughes: > I see what you mean about ordering, but i wonder if there are many real-world scenarios where this would be an issue? It would only affect stores that rely on actions running in a specific order, but their code creates them in the wrong order. It seems quite unlikely to me.

Yep, my instincts match your own, _but,_ the fact is, we don't know for sure (we don't have perfect visibility into all use cases). Some notes:

- Action Scheduler is not only used alongside WooCommerce—it's also included and used by plugins (and perhaps themes) with no relationship to WooCommerce or even e-commerce in general.
- Re the problem of code creating actions in the wrong order, I'm mostly thinking of a situation where multiple components—be they plugins or themes—which may or may not be related, are independently scheduling actions.

And, again, I'm not presenting this as a blocker—I just want to pause and get thoughts/think of other use cases before moving ahead.

> Maybe a middle-ground would be to add a filter and a section on docs advising people who get deadlocks to use it?

I definitely find that to be an attractive idea.

PR #752, opened 2021-08-27
Replace uses of `\ActionScheduler_Store::find_actions` with `\ActionScheduler_Store::query_actions`

danielbitzer:
The data store methods `\ActionScheduler_Store::find_actions` and `\ActionScheduler_Store::query_actions` essentially perform the same function and contain a lot of duplicate code. `find_actions` is less capable and makes a bunch of assumptions which are not implied in the name which makes `query_actions` a much better single choice to use moving forward.

This PR updates the `as_next_scheduled_action()` and `as_unschedule_action()` functions to use the preferred `query_actions` method.

Initially commented here: https://github.com/woocommerce/action-scheduler/pull/688#discussion_r696936795

The only risk I see here is that `find_action` has greater test coverage than `query_actions`. 

Note: `find_action` uses an INNER JOIN while `query_actions` uses as LEFT JOIN. This doesn't affect the end result AFAICS.

### DIFF
--- a/classes/abstracts/ActionScheduler_Store.php
+++ b/classes/abstracts/ActionScheduler_Store.php
@@ -100,6 +100,28 @@ abstract class ActionScheduler_Store extends ActionScheduler_Store_Deprecated {
 	abstract public function query_actions( $query = array(), $query_type = 'select' );
 
 	/**
+	 * Run query to get a single action ID.
+	 *
+	 * @since x.x.x
+	 *
+	 * @see ActionScheduler_Store::query_actions for $query arg usage but 'per_page' is always set to 1.
+	 *
+	 * @param array $query
+	 *
+	 * @return int|null
+	 */
+	public function query_action( $query ) {
+		$query['per_page'] = 1;
+		$results = $this->query_actions( $query );
+
+		if ( empty( $results ) ) {
+			return null;
+		} else {
+			return (int) $results[0];
+		}
+	}
+
+	/**
 	 * Get a count of all actions in the store, grouped by status
 	 *
 	 * @return array
--- a/classes/abstracts/ActionScheduler_Store.php
+++ b/classes/abstracts/ActionScheduler_Store.php
@@ -104,7 +104,7 @@ abstract class ActionScheduler_Store extends ActionScheduler_Store_Deprecated {
 	 *
 	 * @since x.x.x
 	 *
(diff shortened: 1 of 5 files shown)


danielbitzer:
I definitely see where you're coming from. I agree that `find_actions` is a somewhat helpful shortcut method. Retrieving a single action definitely seems like a common use case in the plugin.

The main thing I don't like is that it does too much for a "typical data retrieval function" i.e. the extra logic to change the query ordering based on the `$param[status]` value. This should at least be stated in the doc block. Setting the default `status` to `pending` also seems a bit presumptuous for a query-type function but I guess it's a nice shortcut. I personally prefer functions that are more explicitly named like:
```php
store()->get_next_pending_action( $hook, $args, $group );

// and additionally
store()->get_last_complete_action( $hook, $args, $group );
store()->get_last_failed_action( $hook, $args, $group );
```

But agree that it's probably worth maintaining this method for library consumers even though they should be using the public function like `as_get_scheduled_actions()`.

danielbitzer:
I've updated this PR and I think I've addressed all your feedback. More feedback is welcome of course the discussion is still ongoing about the `find_action` method.

I ended up adding `query_action` as a really simple shortcut to retrieve a single action from the upgraded `::query_actions()` method. I believe this will be useful. Part of the reasoning is that I don't think we can safely upgrade `find_actions` to support multiple statuses because of its additional status handling logic.

barryhughes:
> The main thing I don't like is that it does too much for a "typical data retrieval function"

:100: 

> Part of the reasoning is that I don't think we can safely upgrade find_actions to support multiple statuses because of its additional status handling logic.

Good call.

> But agree that it's probably worth maintaining this method for library consumers even though they should be using the public function like as_get_scheduled_actions().

Yes, I'm just trying to be extra cautious here. Plus, depending on how you define library consumer, some may have custom store implementations of their own (and custom helper functions), rather than being 'simple end-consumers'. Certainly in general, though, the ideal is probably that they use the procedural API...but I don't know that we've actually publicized that recommendation any place.

barryhughes (changes requested):
Nice—left a few more notes, but they are all very minor (linting) issues. Beyond that, LGTM!

danielbitzer:
Thanks @barryhughes! I fixed up those last few phpcs issues.

barryhughes (approved):
💥

Issue #745, opened 2021-08-18
Max execution time of 0 is longer than 30


EdwardBock:
Hi there,

we use the plugin Mailchimp for WooCommerce which uses your action-scheduler. We run wordpress schedules with cli and a max_execution_time of 0 which means infinit. Problem is that your ActionScheduler_Compatibility.php class raises the execution time to 30 which is not more that 0. You should check for zero or larger than $limit as return conditions. Could that be?

https://github.com/woocommerce/action-scheduler/blob/master/classes/ActionScheduler_Compatibility.php#L89

barryhughes:
Thanks for the report, @EdwardBock. If you are happy to submit a PR for this that would be great—though if not we'll do our best to address this from our side.

ovidiul:
The 30 seconds time limit is probably taken from [here](https://github.com/woocommerce/action-scheduler/blob/master/classes/abstracts/ActionScheduler_Abstract_QueueRunner.php#L130), so you might actually need to modify that filter at the Action Scheduler level as:

```
// define the action_scheduler_maximum_execution_time callback 
function custom_action_scheduler_maximum_execution_time( $time_limit ){ 
   //custom code here
    $time_limit = 0;

    return $time_limit
} 

//add the action 
add_filter('action_scheduler_maximum_execution_time', 'custom_action_scheduler_maximum_execution_time', 10, 1);
```

so on that indicated line, the condition will check if 30 < 0, which will return false, and continue with the setting the upper time limit to 30.

barryhughes:
📕  Closing as solved: https://github.com/woocommerce/action-scheduler/pull/755.

:bulb: Further related issue here: https://github.com/woocommerce/action-scheduler/issues/767

PR #755, opened 2021-08-31
Do not lower the max_execution_time if it is already set to 0 (unlimited)

### LINKED ISSUE #754: ActionScheduler_Compatibility::raise_time_limit imposes a limit
if max_execution_time is unlimited i.e. zero 
this function actually creates a limit

Can it please check for limit equals zero
```php
	/**
	 * Attempts to raise the PHP timeout for time intensive processes.
	 *
	 * Only allows raising the existing limit and prevents lowering it. Wrapper for wc_set_time_limit(), when available.
	 *
	 * @param int The time limit in seconds.
	 */
	public static function raise_time_limit( $limit = 0 ) {
		if ( (int)ini_get( 'max_execution_time' ) === 0 || ($limit !== 0 && $limit < ini_get( 'max_execution_time' ) )) {
			return;
		}

```

barryhughes:
Our [`ActionScheduler_Compatibility::raise_time_limit()`](https://github.com/woocommerce/action-scheduler/blob/3.2.1/classes/ActionScheduler_Compatibility.php#L81-L98) is designed so that it (theoretically), _"Only allows raising the existing limit and prevents lowering it."_

However, in cases where the max time is already 0 [(zero means unlimited)](https://www.php.net/manual/en/info.configuration.php#ini.max-execution-time) and a value such as 30 is passed to this method, the `max_execution_time` will indeed be changed to 30, effectively lowering it. This change addresses that.

### Testing

Testing this is a little awkward, but I'd suggest using the WP CLI shell (`wp shell`) and running through permutations similar to those you see in the related tests, for example:

```
% wp shell
>>> ini_set( 'max_execution_time', '0' )
=> "0"
>>> ActionScheduler_Compatibility::raise_time_limit( 30 )
=> null
>>> ini_get( 'max_execution_time' )
=> "0"
>>> # ↑ Should be '0' (with this branch)
>>> #   May be '30' (with current tagged release)
```

Things you can try with this approach:

- Ensure if the `max_execution_time` is zero (unlimited), subsequent calls to this method **do not** change it.
- Ensure if `max_execution_time` is a non-zero positive int (such as 30), calls to this method where a lower non-zero value (such as 20) are passed **do not** change it.
- Ensure if `max_execution_time` is a non-zero positive int (such as 30), calls to this method where zero _or_ a higher value (such as 60) are passed **do** change it appropriately.

If you decide to test this way, remember you will need to restart `wp shell` when you change branch.

Closes #754 and #745.

### DIFF
--- a/classes/ActionScheduler_Compatibility.php
+++ b/classes/ActionScheduler_Compatibility.php
@@ -83,17 +83,27 @@ class ActionScheduler_Compatibility {
 	 *
 	 * Only allows raising the existing limit and prevents lowering it. Wrapper for wc_set_time_limit(), when available.
 	 *
-	 * @param int The time limit in seconds.
+	 * @param int $limit The time limit in seconds.
 	 */
 	public static function raise_time_limit( $limit = 0 ) {
-		if ( $limit < ini_get( 'max_execution_time' ) ) {
+		$limit = (int) $limit;
+		$max_execution_time = (int) ini_get( 'max_execution_time' );
+
+		/*
+		 * If the max execution time is already unlimited (zero), or if it exceeds or is equal to the proposed
+		 * limit, there is no reason for us to make further changes (we never want to lower it).
+		 */
+		if (
+			0 === $max_execution_time
+			|| ( $max_execution_time >= $limit && $limit !== 0 )
+		) {
 			return;
 		}
 
 		if ( function_exists( 'wc_set_time_limit' ) ) {
 			wc_set_time_limit( $limit );
 		} elseif ( function_exists( 'set_time_limit' ) && false === strpos( ini_get( 'disable_functions' ), 'set_time_limit' ) && ! ini_get( 'safe_mode' ) ) { // phpcs:ignore PHPCompatibility.IniDirectives.RemovedIniDirectives.safe_modeDeprecatedRemoved
-			@set_time_limit( $limit );
+			@set_time_limit( $limit ); // phpcs:ignore WordPress.PHP.NoSilencedErrors.Discouraged
 		}
 	}
 }
--- a/tests/phpunit/helpers/ActionScheduler_Compatibility_Test.php
+++ b/tests/phpunit/helpers/ActionScheduler_Compatibility_Test.php
@@ -1,0 +1,46 @@
+<?php
+/**
+ * @group helpers
+ */
+class ActionScheduler_Compatibility_Test extends ActionScheduler_UnitTestCase {
+	/**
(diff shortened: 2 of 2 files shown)


dgoring:
@vedanshujain Can you please review and get this patch merged

vedanshujain (approved):
(Sorry for delayed review).

LGTM! Given that we are touching this method, might as well add also add return statements to both this and in WC's method so that callee can check whether the op was successful or not.

barryhughes:
Hi @vedanshujain!

> Given that we are touching this method, might as well add also add return statements

The more I think about it, and experiment with a few possibilities locally, the less sure I am how valuable a return value would be.

- If the requested `$limit` is less than or equal to `max_execution_time` then we [don't try to change anything](https://github.com/woocommerce/action-scheduler/blob/85aed5f9e12ed4a28cd3510e99a6a4e7dc764e6d/classes/ActionScheduler_Compatibility.php#L96-L101) (notably, this is true regardless of how much execution time has already elapsed).
- I also don't think there is a reliable way to determine how much execution time has elapsed at the point the call is made.
- This leads me to ask, if `$limit` <= `max_execution_time` should we return `false` or `true` (assuming the return type is boolean) and would either of those things actually be useful to the caller?
- In my local testing, rather confusingly, it is possible for a call to `set_time_limit()` to result in `(bool) false` (indicating failure) _yet_ the value of the `max_execution_time` ini setting may still have been modified.
- Rather than return a bool we could return an integer. This might reflect the value of `max_execution_time` at the end of the call, or the remaining execution time (if that can be reliably calculated) ... but I'm not sure that would be particularly useful either.

Last but not least, even if a call to `::raise_time_limit()` is made and that in turn triggers a successful call to `set_time_limit()` ... it doesn't necessarily mean much: execution could still stop prematurely (whether because of userland logic calling `exit` or `wp_die()` or because the web server _or some other supervisor_ decides to end execution). 

Given these things ... wdyt?

dgoring:
@barryhughes 
Given that `set_time_limit()` resets the clock for timeout and that php has no way to know reliably how long is left for execution.

I would suggest you always call `set_time_limit()` (unless already infinite) using either the current max_execution_time or the `$limit` which ever is bigger that way you are always "raising the time limit".

in terms of the return value there really isn't anything other than the result of the `set_time_limit()` until php add a way to retrieve how long left for execution, if it's not reliable then I suggest you leave as is.

vedanshujain:
Hmm, ok, let' leave as it is then for the time being.

PR #757, opened 2021-09-10
Update docs in preparation for 3.3.0 release.

barryhughes:
Changes to [our project docs](https://actionscheduler.org) for the upcoming 3.3.0 release, principally to reflect the addition of the new `as_has_scheduled_action()` function introduced in #688.

💡 Please review but do not merge (merge will happen once the 3.3.0 release happens).

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -108,7 +108,7 @@ require_once( plugin_dir_path( __FILE__ ) . '/libraries/action-scheduler/action-
  * so that our callback is run then.
  */
 function eg_schedule_midnight_log() {
-	if ( false === as_next_scheduled_action( 'eg_midnight_log' ) ) {
+	if ( false === as_has_scheduled_action( 'eg_midnight_log' ) ) {
 		as_schedule_recurring_action( strtotime( 'tomorrow' ), DAY_IN_SECONDS, 'eg_midnight_log' );
 	}
 }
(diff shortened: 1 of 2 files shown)


docs/usage.md
```diff
@@ -123,4 +123,4 @@ function eg_log_action_data() {
 add_action( 'eg_midnight_log', 'eg_log_action_data' );
 ```
 
-For more details on all available API functions, and the data they accept, refer to the [API Reference](/api/).
+Note that the `as_has_scheduled_action()` function was added in 3.3.0: if you are using an earlier version, you should substitute `as_next_scheduled_action()` instead. For more details on all available API functions, and the data they accept, refer to the [API Reference](/api/).
```
roykho, line 126: This might be a personal preference but I find the use of `substitute` here confusing. I am not sure if that means I indeed should substitute `as_next_scheduled_action()` with `as_has_scheduled_action()` or the other way around.

So I would prefer we change it to `use` to be more clear? Again it could be just me that is confused about it.

barryhughes: Good call—I agree that reads better. Updated :+1:

roykho (approved):
LGTM!

barryhughes:
[3.3.0 released](https://github.com/woocommerce/action-scheduler/releases/tag/3.3.0)—merging.

Issue #758, opened 2021-09-13
Release ActionScheduler 3.3.0


vedanshujain:
@barryhughes mentioned that we are ready to release a new minor version 3.3.0, this issue tracks releasing it.

barryhughes:
PRs merged since 3.2.1:

* https://github.com/woocommerce/action-scheduler/pull/688
* https://github.com/woocommerce/action-scheduler/pull/722
* https://github.com/woocommerce/action-scheduler/pull/725
* https://github.com/woocommerce/action-scheduler/pull/727
* https://github.com/woocommerce/action-scheduler/pull/741
* https://github.com/woocommerce/action-scheduler/pull/743
* https://github.com/woocommerce/action-scheduler/pull/744
* https://github.com/woocommerce/action-scheduler/pull/747
* https://github.com/woocommerce/action-scheduler/pull/751
* https://github.com/woocommerce/action-scheduler/pull/752

barryhughes:
🔔 Once 3.3.0 is released, we should go ahead and merge https://github.com/woocommerce/action-scheduler/pull/757 (docs change—already approved).

barryhughes:
[Follow up issue](https://github.com/woocommerce/woocommerce/pull/30719) for WooCommerce core.

PR #621, opened 2020-09-21
Version 3.2.0 schema changes

rrennick:
This is a parent PR for schema changes in version 3.2.0

[ ] - MySQL 5.7 and `NO_ZERO_DATE` conflicts #587, #577, #519 
[ ] - Index to alleviate deadlocks in larger actions tables PR #535 
[ ] - Retries on single actions PR #612

PR #759, opened 2021-09-13
Prepare for 3.3.0

barryhughes:
Updates the initialization functions, ready for the 3.3.0 release.

Also adds:

- Some changes to satisfy the project's PHPCS ruleset (adds docblocks, removes superfluous parens from the require statements, etc).
- Some safety changes that were originally intended to land in the [3.2.0 release](https://github.com/woocommerce/action-scheduler/pull/621/files#diff-3656dd24b02354fa4258e4ed49dafa655b6ca34dc8b436752e6c991f2ca8418a) (including the check to see if `add_action()` is defined, and prefixing of the require paths with `__DIR__`) that were missed previously.

Updates #758.

### Testing instructions

Simple:

- Deactivate all other plugins.
- Checkout this branch (into the plugin directory, if it isn't already there).
- Activate.
- Confirm the Tools → Scheduled Actions screen is available.

Further (optional) testing steps:

- Additionally clone a second copy of Action Scheduler.
- Clone into a plugin directory named something like `aaa_action_scheduler` (so that WP loads this new version first).
- Checkout `3.2.1` (or any earlier tag) and run `composer i`.
- Activate (you should now have two versions of Action Scheduler installed and active).
- Confirm the Tools → Scheduled Actions screen is available.
- Using the Help tab on that screen, confirm it reads **About Action Scheduler 3.3.0**.

help-tab

### DIFF
--- a/action-scheduler.php
+++ b/action-scheduler.php
@@ -1,5 +1,5 @@
 <?php
-/*
+/**
  * Plugin Name: Action Scheduler
  * Plugin URI: https://actionscheduler.org
  * Description: A robust scheduling library for use in WordPress plugins.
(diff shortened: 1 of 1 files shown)


action-scheduler.php
```diff
@@ -23,35 +23,42 @@
...
-			require_once( 'classes/abstracts/ActionScheduler.php' );
+			require_once __DIR__ . '/classes/abstracts/ActionScheduler.php';
 			ActionScheduler::init( __FILE__ );
 		}
 	}
 
 	// Support usage in themes - load this version if no plugin has loaded a version yet.
 	if ( did_action( 'plugins_loaded' ) && ! doing_action( 'plugins_loaded' ) && ! class_exists( 'ActionScheduler' ) ) {
-		action_scheduler_initialize_3_dot_2_dot_1();
+		action_scheduler_register_3_dot_3_dot_0();
```
jonathansadowski, line 61: Is this intentionally `register` instead of `initialize`?

jonathansadowski (commented):
Test-plan worked as described. I added a question about one of the changes... I also noticed that both plugins were listed as 3.2.1 from within the plugins screen, but I assume that version bump would be handled by the release script, so isn't of concern, but I just wanted to mention it, in case it isn't.

action-scheduler.php
barryhughes: Not intentional. Fixed, thank you.

(Unusual for the library to be included in this fashion, but we still want to support it, so good catch.)

barryhughes:
> I assume that version bump would be handled by the release script

Exactly: I should have mentioned that discrepancy but, yes—that's expected (at this point) and is not a concern.

jonathansadowski (approved):

Issue #660, opened 2021-02-20
Security Issues in ActionScheduler_Abstract_ListTable.php


amberhinds:
Hello!

We've included Action Scheduler in one of our plugins (thank you!) and while doing security audits on our plugin, identified problems related to Action Scheduler.

We ran the plugin through [WP Engine's linting test](https://github.com/wpengine/wpengine-coding-standards/) which helps identify best practices and potential problems. For this process, we are using [PHP Codesniffer](https://tracking.cirrusinsight.com/4bd98769-7476-4065-a754-2ca80b4094d2/github-com-squizlabs-php-codesniffer) with rules derived from both the [WordPress Coding Standards](https://tracking.cirrusinsight.com/4bd98769-7476-4065-a754-2ca80b4094d2/github-com-wordpress-coding-standards-wordpress-coding-standards) and [PHPCompatibility rulesets](https://tracking.cirrusinsight.com/4bd98769-7476-4065-a754-2ca80b4094d2/github-com-wimg-phpcompatibility). Below is a detailed line-by-line report of the sniff violation.  

Can you please let me know if these are actual errors that require fixes? If so, we may be able to submit a pull request with fixes.

FILE: /includes/action-scheduler/classes/abstracts/ActionScheduler_Abstract_ListTable.php
------------------------------------------------------------------------------------------------------------------------------------------
FOUND 48 ERRORS AND 30 WARNINGS AFFECTING 30 LINES
------------------------------------------------------------------------------------------------------------------------------------------
 151 | ERROR   | Missing wp_unslash() before sanitization. (WordPress.Security.ValidatedSanitizedInput.MissingUnslash)
 151 | ERROR   | Detected usage of a non-sanitized input variable: $_GET
     |         | (WordPress.Security.ValidatedSanitizedInput.InputNotSanitized)
 152 | ERROR   | Missing wp_unslash() before sanitization. (WordPress.Security.ValidatedSanitizedInput.MissingUnslash)
 152 | ERROR   | Detected usage of a non-sanitized input variable: $_GET
     |         | (WordPress.Security.ValidatedSanitizedInput.InputNotSanitized)
 152 | ERROR   | Use placeholders and $wpdb->prepare(); found $ids_sql (WordPress.DB.PreparedSQL.NotPrepared)
 152 | ERROR   | Missing wp_unslash() before sanitization. (WordPress.Security.ValidatedSanitizedInput.MissingUnslash)
[...]

PR #763, opened 2021-09-16
phpcs ActionScheduler_Abstract_ListTable fixes

ovidiul:
This PR will address the PHPCS issues mentioned here https://github.com/woocommerce/action-scheduler/issues/660

Current PHPCS scan shows:

```

FILE: .../action-scheduler/classes/abstracts/ActionScheduler_Abstract_ListTable.php
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
FOUND 7 ERRORS AND 2 WARNINGS AFFECTING 7 LINES
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
   1 | ERROR   | Missing file doc comment
 390 | WARNING | Comment refers to a TODO task
 406 | ERROR   | Use placeholders and $wpdb->prepare(); found $column
 406 | ERROR   | SQL wildcards for a LIKE query should be passed in through a replacement parameter and the variable part of the replacement should be escaped using "esc_like()". Found:  like "%%s%".
 406 | WARNING | Replacement variables found, but no valid placeholders found in the query.
 429 | ERROR   | Use placeholders and $wpdb->prepare(); found interpolated variable $column at "`$column` = %s"
 454 | ERROR   | Detected usage of a possibly undefined superglobal array index: $_SERVER['REQUEST_URI']. Use isset() or empty() to check the index exists before using it
 479 | ERROR   | Use placeholders and $wpdb->prepare(); found $sql
 482 | ERROR   | Use placeholders and $wpdb->prepare(); found $query_count
```

would it make sense to `phpcs:ignore` those?

### DIFF
--- a/classes/abstracts/ActionScheduler_Abstract_ListTable.php
+++ b/classes/abstracts/ActionScheduler_Abstract_ListTable.php
@@ -1,7 +1,7 @@
 <?php
 
 if ( ! class_exists( 'WP_List_Table' ) ) {
-	require_once( ABSPATH . 'wp-admin/includes/class-wp-list-table.php' );
+	require_once ABSPATH . 'wp-admin/includes/class-wp-list-table.php';
 }
 
 /**
@@ -13,8 +13,8 @@ if ( ! class_exists( 'WP_List_Table' ) ) {
  * which columns needs to be shown, filter, ordered by and more and forget about the details.
  *
  * This class supports:
- *	- Bulk actions
- *	- Search
+ *  - Bulk actions
+ *  - Search
  *  - Sortable columns
  *  - Automatic translations of the columns
  *
@@ -25,22 +25,30 @@ abstract class ActionScheduler_Abstract_ListTable extends WP_List_Table {
 
 	/**
 	 * The table name
+	 *
+	 * @var string
 	 */
 	protected $table_name;
 
 	/**
 	 * Package name, used to get options from WP_List_Table::get_items_per_page.
+	 *
+	 * @var string
 	 */
 	protected $package;
 
 	/**
 	 * How many items do we render per page?
+	 *
+	 * @var int
 	 */
 	protected $items_per_page = 10;
 
 	/**
 	 * Enables search in this table listing. If this array
 	 * is empty it means the listing is not searchable.
+	 *
+	 * @var array
 	 */
 	protected $search_by = array();
 
@@ -48,6 +56,8 @@ abstract class ActionScheduler_Abstract_ListTable extends WP_List_Table {
 	 * Columns to show in the table listing. It is a key => value pair. The
 	 * key must much the table column name and the value is the label, which is
 	 * automatically translated.
+	 *
+	 * @var array
 	 */
 	protected $columns = array();
 
@@ -58,34 +68,51 @@ abstract class ActionScheduler_Abstract_ListTable extends WP_List_Table {
 	 * The array of actions are key => value, where key is the method name
 	 * (with the prefix row_action_<key>) and the value is the label
 	 * and title.
+	 *
+	 * @var array
 	 */
 	protected $row_actions = array();
 
 	/**
 	 * The Primary key of our table
+	 *
+	 * @var string
 	 */
 	protected $ID = 'ID';
 
 	/**
 	 * Enables sorting, it expects an array
 	 * of columns (the column names are the values)
+	 *
+	 * @var array
 	 */
 	protected $sort_by = array();
 
+	/**
+	 * The default sort order
+	 *
+	 * @var string
+	 */
 	protected $filter_by = array();
 
 	/**
-	 * @var array The status name => count combinations for this table's items. Used to display status filters.
+	 * The status name => count combinations for this table's items. Used to display status filters.
+	 *
+	 * @var array
 	 */
 	protected $status_counts = array();
 
 	/**
-	 * @var array Notices to display when loading the table. Array of arrays of form array( 'class' => {updated|error}, 'message' => 'This is the notice text display.' ).
+	 * Notices to display when loading the table. Array of arrays of form array( 'class' => {updated|error}, 'message' => 'This is the notice text display.' ).
+	 *
+	 * @var array
 	 */
 	protected $admin_notices = array();
 
 	/**
-	 * @var string Localised string displayed in the <h1> element above the able.
+	 * Localised string displayed in the <h1> element above the able.
+	 *
+	 * @var string
 	 */
 	protected $table_header;
 
@@ -99,6 +126,8 @@ abstract class ActionScheduler_Abstract_ListTable extends WP_List_Table {
 	 * validations and afterwards will execute the bulk method, with two arguments. The first argument
 	 * is the array with primary keys, the second argument is a string with a list of the primary keys,
 	 * escaped and ready to use (with `IN`).
+	 *
+	 * @var array
 	 */
 	protected $bulk_actions = array();
 
@@ -106,7 +135,9 @@ abstract class ActionScheduler_Abstract_ListTable extends WP_List_Table {
 	 * Makes translation easier, it basically just wraps
 	 * `_x` with some default (the package name).
 	 *
-	 * @deprecated 3.0.0
+	 * @param string $text The new text to translate.
+	 * @param string $context The context of the text.
+	 * @return string|void The translated text.
 	 */
 	protected function translate( $text, $context = '' ) {
 		return $text;
@@ -116,6 +147,10 @@ abstract class ActionScheduler_Abstract_ListTable extends WP_List_Table {
 	 * Reads `$this->bulk_actions` and returns an array that WP_List_Table understands. It
 	 * also validates that the bulk method handler exists. It throws an exception because
 	 * this is a library meant for developers and missing a bulk method is a development-time error.
+	 *
+	 * @return array
+	 *
+	 * @throws RuntimeException Throws RuntimeException when the bulk action does not have a callback method.
 	 */
 	protected function get_bulk_actions() {
 		$actions = array();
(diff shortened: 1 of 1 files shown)


barryhughes:
> would it make sense to phpcs:ignore those?

I think in many cases that would be fine; for some like `Missing file doc comment` I wonder if instead we should modify `phpcs.xml` to ignore that rule everywhere.

> Comment refers to a TODO task

For this one, thinking it is probably better that we don't ignore it (so that we get a reminder to take action) _or_ remove it from the source and create an issue for it instead.

ovidiul:
I've added the `Squiz.Commenting.FileComment.Missing` rules to the phpcs.xml file, although maybe I should open a separate PR for that?

Also added the comments for the rest of the items, and also refactored this [query](https://github.com/woocommerce/action-scheduler/pull/763/files#diff-18ffca2c4125958b98cb71171cb6ddfc0ec7829f127c0094e24c6c2796cce326R403-R405) for better handling.

As for the TODO, I've opened a new issue here https://github.com/woocommerce/action-scheduler/issues/766 .

Latest PHPCS shows no Warnings/Errors.

classes/abstracts/ActionScheduler_Abstract_ListTable.php
```diff
@@ -99,14 +126,18 @@ abstract class ActionScheduler_Abstract_ListTable extends WP_List_Table {
...
+	 *
+	 * @var array
 	 */
 	protected $bulk_actions = array();
 
 	/**
 	 * Makes translation easier, it basically just wraps
 	 * `_x` with some default (the package name).
 	 *
-	 * @deprecated 3.0.0
```
barryhughes, line 109: Was this dropped accidentally? Seems like we'd want to preserve it and I don't think it breaks our ruleset.

classes/abstracts/ActionScheduler_Abstract_ListTable.php
```diff
@@ -343,21 +387,22 @@ protected function get_table_columns() {
...
 		}
 
-		$filter  = array();
+		$search_string = $wpdb->esc_like( sanitize_text_field( wp_unslash( $_GET['s'] ) ) ); //phpcs:ignore WordPress.Security.NonceVerification.Recommended
+
+		$filter = array();
 		foreach ( $this->search_by as $column ) {
-			$filter[] = $wpdb->prepare('`' . $column . '` like "%%s%"', $wpdb->esc_like( $_GET['s'] ));
+			$wild     = '%';
+			$sql_like = $wild . $wpdb->esc_like( $search_string ) . $wild;
```
barryhughes, line 404: Do we need to pass `$search_string` through `esc_like()` twice (looking up at like 399)?

classes/abstracts/ActionScheduler_Abstract_ListTable.php
```diff
@@ -574,7 +655,7 @@ protected function display_header() {
 	 */
 	protected function display_admin_notices() {
 		foreach ( $this->admin_notices as $notice ) {
-			echo '<div id="message" class="' . $notice['class'] . '">';
+			echo '<div id="message" class="' . esc_html( $notice['class'] ) . '">';
```
barryhughes, line 658: `esc_attr()` could be a better choice here, _or_ `sanitize_html_class()` (but the latter would only be safe if we are certain multiple classes are not being passed in here):

```suggestion
			echo '<div id="message" class="' . esc_attr( $notice['class'] ) . '">';
```

barryhughes (commented):
> I've added the Squiz.Commenting.FileComment.Missing rules to the phpcs.xml file, although maybe I should open a separate PR for that?

Perhaps, but since you've added it here let's just keep it in one place (thanks for adding it!).

barryhughes (changes requested):
Left a couple of minor observations: beyond that I just want to do a bit of testing before we merge (changes look great, but it's also a lot of change in a key class—so I just want to make sure we're covered).

classes/abstracts/ActionScheduler_Abstract_ListTable.php
ovidiul: Of course not, somehow I've overseen this. I've left the esc_like in the SQL statement as it makes more sense there and removed the duplicate above.

classes/abstracts/ActionScheduler_Abstract_ListTable.php
ovidiul: Yes, my bad. It's back now, I've added the text `Use `__()` instead.` along with it, does it make sense to be there?

classes/abstracts/ActionScheduler_Abstract_ListTable.php
ovidiul: Suggestion added :)

classes/abstracts/ActionScheduler_Abstract_ListTable.php
barryhughes: The method description suggests it is a wrapper around `_x()` (even though it isn't) and, like `_x()` it accepts a `$context`, so _probably_ this would be just a little better:

```
* @deprecated 3.0.0 Use `_x()` instead.
```

ovidiul: Makes sense, changed text added.

classes/abstracts/ActionScheduler_Abstract_ListTable.php
```diff
@@ -612,7 +695,7 @@ protected function display_filter_by_status() {
 
 		if ( $status_list_items ) {
 			echo '<ul class="subsubsub">';
-			echo implode( " | \n", $status_list_items );
+			echo implode( " | \n", esc_html( $status_list_items ) );
```
barryhughes, line 698: This triggers an error, breaking the Tools → Scheduled Actions screen entirely (because `implode` expects the second param to be an array—and passing the array through `esc_html()` gives us a string).

Each element is already [composed in a safe manner](https://github.com/woocommerce/action-scheduler/pull/763/files#diff-18ffca2c4125958b98cb71171cb6ddfc0ec7829f127c0094e24c6c2796cce326R693) and contains HTML, so `esc_html()` wouldn't be a good fit (since it would expose the HTML). 

Alternative:

```suggestion
			echo implode( " | \n", $status_list_items ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
```

classes/abstracts/ActionScheduler_Abstract_ListTable.php
```diff
@@ -146,22 +183,31 @@ protected function process_bulk_action() {
 
 		check_admin_referer( 'bulk-' . $this->_args['plural'] );
 
-		$method   = 'bulk_' . $action;
+		$method = 'bulk_' . $action;
 		if ( array_key_exists( $action, $this->bulk_actions ) && is_callable( array( $this, $method ) ) && ! empty( $_GET['ID'] ) && is_array( $_GET['ID'] ) ) {
 			$ids_sql = '(' . implode( ',', array_fill( 0, count( $_GET['ID'] ), '%s' ) ) . ')';
-			$this->$method( $_GET['ID'], $wpdb->prepare( $ids_sql, $_GET['ID'] ) );
+			$id      = sanitize_text_field( wp_unslash( $_GET['ID'] ) );
```
barryhughes, line 189: `$_GET['ID']` will be an array (we're dealing with bulk actions, so more than one can be passed), so we need to take a different approach as passing an array through `sanitize_text_field()` gives us an empty string and that means the bulk action will not be processed successfully.

Recommend:

```suggestion
			$id      = array_map( 'absint', $_GET['ID'] );
```

classes/abstracts/ActionScheduler_Abstract_ListTable.php
```diff
@@ -527,32 +593,49 @@ protected function maybe_render_actions( $row, $column_name ) {
...
+	/**
+	 * Process the bulk actions.
+	 *
+	 * @return void
+	 */
 	protected function process_row_actions() {
 		$parameters = array( 'row_action', 'row_id', 'nonce' );
 		foreach ( $parameters as $parameter ) {
-			if ( empty( $_REQUEST[ $parameter ] ) ) {
+			if ( ! isset( $_REQUEST[ $parameter ] ) || empty( $_REQUEST[ $parameter ] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```
barryhughes, line 604: Just curious, why the extra check here? I don't think it's needed to satisfy our linting rules (and I think `isset` and `empty` both have similar performance characteristics in this scenario—where the element is missing entirely, I mean)?

classes/abstracts/ActionScheduler_Abstract_ListTable.php
```diff
@@ -527,32 +593,49 @@ protected function maybe_render_actions( $row, $column_name ) {
...
 
-		$method = 'row_action_' . $_REQUEST['row_action'];
+		$action = sanitize_text_field( wp_unslash( $_REQUEST['row_action'] ) ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended, WordPress.Security.ValidatedSanitizedInput.InputNotValidated
+		$row_id = sanitize_text_field( wp_unslash( $_REQUEST['row_id'] ) ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended, WordPress.Security.ValidatedSanitizedInput.InputNotValidated
+		$nonce  = sanitize_text_field( wp_unslash( $_REQUEST['nonce'] ) ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended, WordPress.Security.ValidatedSanitizedInput.InputNotValidated
+		$method = 'row_action_' . $action; // phpcs:ignore WordPress.Security.NonceVerification.Recommended
 
-		if ( $_REQUEST['nonce'] === wp_create_nonce( $_REQUEST[ 'row_action' ] . '::' . $_REQUEST[ 'row_id' ] ) && method_exists( $this, $method ) ) {
-			$this->$method( $_REQUEST['row_id'] );
+		if ( wp_create_nonce( $action . '::' . $row_id ) === $nonce && method_exists( $this, $method ) ) {
```
barryhughes, line 614: I realize you didn't introduce this and are just making a change to satisfy the yoda rule, but thinking about it this isn't the right way to check a nonce.

*Optional* thing we can slip in:

```suggestion
		if ( wp_verify_nonce( $nonce, $action . '::' . $row_id ) && method_exists( $this, $method ) ) {
```

classes/abstracts/ActionScheduler_Abstract_ListTable.php
```diff
@@ -644,9 +727,9 @@ public function process_actions() {
 		$this->process_bulk_action();
 		$this->process_row_actions();
 
-		if ( ! empty( $_REQUEST['_wp_http_referer'] ) ) {
+		if ( ! empty( $_REQUEST['_wp_http_referer'] && ! empty( $_SERVER['REQUEST_URI'] ) ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```
barryhughes, line 730: Order of parentheses needs to be tweaked:

```suggestion
		if ( ! empty( $_REQUEST['_wp_http_referer'] ) && ! empty( $_SERVER['REQUEST_URI'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```

barryhughes (changes requested):
Sorry for the drip feed of notes here, but I think we'll be all set when the remaining points are addressed:

- [Change needed in the `get_bulk_actions()` method](https://github.com/woocommerce/action-scheduler/pull/763/files#r718788123)
- [Change needed in the `process_actions()` method](https://github.com/woocommerce/action-scheduler/pull/763/files#r718817923)
- [*Optional* change re nonce verification](https://github.com/woocommerce/action-scheduler/pull/763/files#r718811603)

classes/abstracts/ActionScheduler_Abstract_ListTable.php
ovidiul: definitely not needed, `isset` check removed now

ovidiul:
@barryhughes latest changes made, thank you for your guidance on this!

barryhughes (approved):
Great, thank you!

barryhughes:
Tests are failing, but very probably because of [this change](https://make.wordpress.org/core/2021/09/27/changes-to-the-wordpress-core-php-test-suite/) rather than anything in this PR.

barryhughes:
Thanks again!

Merged via https://github.com/woocommerce/action-scheduler/pull/779 (which was just to integrate changes from `master`, in order to have our CI checks run against this changeset).

PR #775, opened 2021-10-15
Test Suite Maintenance

### LINKED ISSUE #770: Updates needed for CI/Test Suite
For recent PRs, tests have started failing in various cases. We likely need to integrate [PHPUnit Polyfills](https://github.com/Yoast/PHPUnit-Polyfills) and make any other changes required per [this post](https://make.wordpress.org/core/2021/09/27/changes-to-the-wordpress-core-php-test-suite/).

barryhughes:
Some maintenance on our test suite (mostly needed because, as of some recent date, tests have been failing for all PRs).

🖖🏽 We previously used our custom PHPUnit 7.5 fork for PHP 7.4 upwards, but in the case of PHP 7.4 that is no longer necessary and even leads to breakages. Therefore we now only use the fork for PHP 8.0 upwards (I'd love to have switched to PHPUnit 9.5 for this, but that doesn't seem viable yet).

🤹🏽 We pull in the PHPUnit polyfills per recent changes in the [core WP test suite](https://make.wordpress.org/core/2021/09/27/changes-to-the-wordpress-core-php-test-suite/).

✨ Tests install script given a polish, in the sense that it has been updated to follow the current recommended template per WP CLI test scaffolding.

⚒️ Last but not least, fixed the way we use the mock's `with()` method (which expects an array, and is less tolerant of us breaking that expectation when run under more recent setups like PHP 8).

### DIFF
--- a/.github/workflows/pr-unit-tests.yml
+++ b/.github/workflows/pr-unit-tests.yml
@@ -59,22 +59,30 @@ jobs:
       - name: Setup and install composer
         run: composer install
 
-      - name: Setup PhpUnit
+      - name: Setup PHPUnit 
         run: |
-          # PHP 5.6 and 7.0: use PhpUnit 5.7
+          # PHPUnit 5.7 when using PHP 5.6 - 7.0.
           if [ "$(php -r "echo version_compare( PHP_VERSION, '7.1', '<' );")" ]; then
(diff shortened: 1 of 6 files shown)


jeffstieler (approved):
I initially had questions about the changes to `install-wp-tests.sh` but then I found https://github.com/wp-cli/scaffold-command/blob/master/templates/install-wp-tests.sh .

LGTM. :shipit:

PR #782, opened 2021-10-22
fix: parsing of day of week when `#` used

### LINKED ISSUE #781: Limitation of `as_schedule_cron_action()`
Currently, cron scheduler can not be used for scheduling tasks on a certain day (Saturday or Sunday) of the first (second, third, fourth) week every month.

**Used schedule string example:**
```
0 12 ? 1/1 SUN#2 *
```
**Thrown Exception**
```
InvalidArgumentException: Weekday must be a value between 1 and 5. 0 given
```

shohag121:
Currently, we only can use 1 to 5 as a **"day of week"** when we use `#` sign with it.

In this PR, **Day of week** range check is set to 0 to 7.
it fixes the parsing of string with `#` sign. eg. `0 12 ? 1/1 SUN#2 *`

fix: https://github.com/woocommerce/action-scheduler/issues/781

### DIFF
--- a/lib/cron-expression/CronExpression_DayOfWeekField.php
+++ b/lib/cron-expression/CronExpression_DayOfWeekField.php
@@ -53,8 +53,8 @@ class CronExpression_DayOfWeekField extends CronExpression_AbstractField
         if (strpos($value, '#')) {
             list($weekday, $nth) = explode('#', $value);
             // Validate the hash fields
-            if ($weekday < 1 || $weekday > 5) {
-                throw new InvalidArgumentException("Weekday must be a value between 1 and 5. {$weekday} given");
+            if ($weekday < 0 || $weekday > 7) {
+                throw new InvalidArgumentException("Weekday must be a value between 0 and 7. {$weekday} given");
             }
             if ($nth > 5) {
                 throw new InvalidArgumentException('There are never more than 5 of a given weekday in a month');


shohag121:
Hi @barryhughes,

Would you kindly review this PR and let me know if any additional change is needed?

Thanks.

barryhughes:
Sorry for the delay here, @shohag121. Testing this out:

| &hellip; | Expression | Before fix | After fix |
| --- | --- | --- | --- |
| ⛔️ | `0 12 ? 1/1 SUN#2 *` |  *Must be a value between 1 and 5* | *Impossible CRON expression* |
| ⛔️ | `0 12 * * SUN#2` |  *Must be a value between 1 and 5* | *Impossible CRON expression* |
| ✅ | `0 12 * * SAT#2`  |  *Must be a value between 1 and 5* | *(Works!)* |
| ✅ | `0 12 * * MON#2`  |  *(Works!)* | *(Works!)* |

It looks like:

- In the first two cases (using Sunday), the exception you noted is solved but we hit a different problem *(impossible CRON expression)*
- In the third case (using Saturday), the exception you noted is solved and there is no further issue
- In the final case (using Monday - Friday), there is no change

Do you find the same? It does look like this yields an incremental improvement—as demonstrated by the third scenario—but some further work is needed to fully smooth things out. Ideally, we would also add some further test cases to `ActionScheduler_CronSchedule_Test` (and we could help with that if needed).

barryhughes:
📝 Note to self: so far as I can tell, this lib is a custom fork that lives directly within our repo (ie, it is likely fine to adjust directly).

rrennick:
> so far as I can tell, this lib is a custom fork that lives directly within our repo (ie, it is likely fine to adjust directly).

It is to both.

barryhughes:
Closing/stale.

PR #783, opened 2021-10-27
fix: clearer logger timeout message

### LINKED ISSUE #646: Explain in logs that timeouts are for an unknown error and/or rename timeouts
Action Scheduler will mark an action as failed when it has been processing for 5 minutes. The reason behind this is that if the action has been _in-progress_ for 5 minutes, some kind of undetectable, unrecoverable error has likely occurred. It's rare an action needs that long, and even more a server would allow a PHP process to run for that long processing an action.

Action Scheduler calls this a "timeout", because basically the action has run out of the allowed processing time that Action Scheduler permits. The log entry will say:

```
action timed out after 300 seconds
```

This causes confusion. Folks often think the "timeout" being referred to there is a PHP or server timeout. This confusion can lead to situations where the first recommended or attempted fix is to increase `max_execution_time`, which will often be in vain, because the issue rarely relates to the action having enough time to be processed. 

Instead, to diagnose the cause of the issue usually requires reviewing server or MySQL logs to find server level or MySQL level errors that couldn't be caught and logged by PHP and Action Scheduler.

To help prevent this, instead of logging this:

```
action timed out after 300 seconds
```

Action Scheduler could log something like this:

```
Action timed out after 300 seconds. Unknown error occurred. Check server, PHP and database error logs to diagnose cause.
```

Or maybe remove the timeout language completely to something like:

```
Action marked as failed after 300 seconds. Unknown error occurred. Check server, PHP and database error logs to diagnose cause.
```

ovidiul:
This PR fixes https://github.com/woocommerce/action-scheduler/issues/646 by making the timeout message more clear and indicating this is a marked failure due to a time limit, so an intentional action that could help users locate the source of the issue faster.

There should be no performance issues around it.

### Changelog

> Tweak - Improve messaging when logging timeout errors.

### DIFF
--- a/classes/abstracts/ActionScheduler_Logger.php
+++ b/classes/abstracts/ActionScheduler_Logger.php
@@ -109,8 +109,7 @@ abstract class ActionScheduler_Logger {
 
 	public function log_timed_out_action( $action_id, $timeout ) {
 		/* translators: %s: amount of time */
-		$this->log( $action_id, sprintf( __( 'action marked as failed after %s seconds. Unknown error occurred. Check server, PHP and database error logs to diagnose cause.
-', 'action-scheduler' ), $timeout ) );
+		$this->log( $action_id, sprintf( __( 'action marked as failed after %s seconds. Unknown error occurred. Check server, PHP and database error logs to diagnose cause.', 'action-scheduler' ), $timeout ) );
 	}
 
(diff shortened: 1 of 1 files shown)


barryhughes (approved):
Nice, thank you.

PR #784, opened 2021-10-27
Feature: adding the --ignore-group CLI option

### LINKED ISSUE #736: Option to ignore some groups when using the wp-cli command
Hello,

It would be nice to have an option to ignore some groups when using the wp-cli command.

For example:
1. run actions from `cache` group: ```wp action-scheduler run --group=cache```
2. run actions from all groups, except `cache`: ```wp action-scheduler run --group=-cache``` or maybe ```wp action-scheduler run --ignore-group=cache```

What do you think?

ovidiul:
This PR fixes https://github.com/woocommerce/action-scheduler/issues/736.

As suggested in the issue, a `--ignore-group` option is the desired approach to name this.

The PR also adds support for multiple comma-separated groups, so `--ignore-group=group1,group2` could be used as well. This implementation would also allow for the `--group` option to take comma-separated values as `--group=group1,group2`.

## Why this approach?

Looking at the `$runner->setup( $batch, $hooks, $group, $force, $ignore_group );` alternative, where we would have to pass the `$ignore_group` additional parameter, the `$group` and `$ignore_group` are very similar, so maybe using a flag to specify which group to ignore as proposed in this PR might make the code a little bit simpler?

## Tests - TBD

Happy to add any related tests if this approach is considered valid.

### DIFF
--- a/classes/abstracts/ActionScheduler_Store.php
+++ b/classes/abstracts/ActionScheduler_Store.php
@@ -238,6 +238,8 @@ abstract class ActionScheduler_Store extends ActionScheduler_Store_Deprecated {
 	public static function sanitize_group_name( $slug ) {
 		// removing the ignored group prefix from slug if present.
 		$slug = preg_replace( '/(' . preg_quote( self::IGNORE_GROUP_FLAG, null ) . ')?(.*)/m', '$2', $slug );
+
+		// further sanitisation here with `sanitize_title` maybe?
 		return $slug;
 	}
 
--- a/classes/abstracts/ActionScheduler_Store.php
+++ b/classes/abstracts/ActionScheduler_Store.php
@@ -11,7 +11,7 @@ abstract class ActionScheduler_Store extends ActionScheduler_Store_Deprecated {
 	const STATUS_FAILED         = 'failed';
 	const STATUS_CANCELED       = 'canceled';
 	const DEFAULT_CLASS         = 'ActionScheduler_wpPostStore';
-	const IGNORE_GROUP_FLAG    = '--';
+	const IGNORE_GROUP_FLAG     = '--*IGNORE*--';
 
 	/** @var ActionScheduler_Store */
 	private static $store = NULL;
--- a/classes/abstracts/ActionScheduler_Store.php
+++ b/classes/abstracts/ActionScheduler_Store.php
@@ -226,7 +226,7 @@ abstract class ActionScheduler_Store extends ActionScheduler_Store_Deprecated {
 	 * @param string $slug Group slug name.
 	 * @return string
 	 */
-	public static function mark_group_for_exclussion( $slug ) {
+	public static function mark_group_for_exclusion( $slug ) {
 		return $slug ? self::IGNORE_GROUP_FLAG . $slug : $slug;
 	}
 
--- a/classes/abstracts/ActionScheduler_Store.php
+++ b/classes/abstracts/ActionScheduler_Store.php
@@ -238,7 +238,7 @@ abstract class ActionScheduler_Store extends ActionScheduler_Store_Deprecated {
 	 */
 	public static function sanitize_group_name( $slug ) {
 		// removing the ignored group prefix from slug if present.
-		$slug = preg_replace( '/(' . preg_quote( self::IGNORE_GROUP_FLAG, null ) . ')?(.*)/m', '$2', $slug );
+		$slug = preg_replace( '/^(' . preg_quote( self::IGNORE_GROUP_FLAG, null ) . ')?(.*)/m', '$2', esc_sql( $slug ) );
 
 		// further sanitisation here with `sanitize_title` maybe?
 		return $slug;
--- a/classes/abstracts/ActionScheduler_Store.php
+++ b/classes/abstracts/ActionScheduler_Store.php
@@ -220,7 +220,8 @@ abstract class ActionScheduler_Store extends ActionScheduler_Store_Deprecated {
 	}
 
 	/**
-	 * Mark group for exclussion by prefixing with '-'.
+	 * Mark group for exclusion by applying a special prefix (by default, the value of
+	 * the IGNORE_GROUP_FLAG property).
 	 *
 	 * @param string $slug Group slug name.
 	 * @return string
--- a/classes/abstracts/ActionScheduler_Store.php
+++ b/classes/abstracts/ActionScheduler_Store.php
@@ -226,7 +226,7 @@ abstract class ActionScheduler_Store extends ActionScheduler_Store_Deprecated {
 	 * @return string
 	 */
 	public static function mark_group_for_exclussion( $slug ) {
-		return self::IGNORE_GROUP_FLAG . $slug;
+		return $slug ? self::IGNORE_GROUP_FLAG . $slug : $slug;
 	}
 
 	/**
--- a/classes/abstracts/ActionScheduler_Store.php
+++ b/classes/abstracts/ActionScheduler_Store.php
@@ -5,12 +5,13 @@
  * @codeCoverageIgnore
  */
 abstract class ActionScheduler_Store extends ActionScheduler_Store_Deprecated {
-	const STATUS_COMPLETE = 'complete';
-	const STATUS_PENDING  = 'pending';
-	const STATUS_RUNNING  = 'in-progress';
-	const STATUS_FAILED   = 'failed';
-	const STATUS_CANCELED = 'canceled';
-	const DEFAULT_CLASS   = 'ActionScheduler_wpPostStore';
+	const STATUS_COMPLETE       = 'complete';
+	const STATUS_PENDING        = 'pending';
+	const STATUS_RUNNING        = 'in-progress';
+	const STATUS_FAILED         = 'failed';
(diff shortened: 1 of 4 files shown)


barryhughes:
Sorry for the delay, here.

- This looks like a valuable and useful change.
- Some very minor tweaks (formatting etc) to make.
- It seems making the ignore prefix even more 'obscure' would be an easy way to add just a little more safety.
- Let's also document the change (ie, in the `docs/` folder).

We are a little pressed for bandwidth right now, so can't act on the above immediately, but this looks good in general. Thank you!

classes/abstracts/ActionScheduler_Store.php
```diff
@@ -208,6 +209,40 @@ abstract public function get_claim_id( $action_id );
...
+		return self::IGNORE_GROUP_FLAG === substr( $slug, 0, strlen( self::IGNORE_GROUP_FLAG ) );
+	}
+
+	/**
+	 * Mark group for exclussion by prefixing with '-'.
+	 *
+	 * @param string $slug Group slug name.
+	 * @return string
+	 */
+	public static function mark_group_for_exclussion( $slug ) {
```
barryhughes: We should change to:

```suggestion
	public static function mark_group_for_exclusion( $slug ) {
```

(And update call-site.)

classes/abstracts/ActionScheduler_Store.php
```diff
@@ -5,12 +5,13 @@
...
-	const STATUS_FAILED   = 'failed';
-	const STATUS_CANCELED = 'canceled';
-	const DEFAULT_CLASS   = 'ActionScheduler_wpPostStore';
+	const STATUS_COMPLETE       = 'complete';
+	const STATUS_PENDING        = 'pending';
+	const STATUS_RUNNING        = 'in-progress';
+	const STATUS_FAILED         = 'failed';
+	const STATUS_CANCELED       = 'canceled';
+	const DEFAULT_CLASS         = 'ActionScheduler_wpPostStore';
+	const IGNORE_GROUP_FLAG    = '--';
```
barryhughes: It's unlikely of course that there are many existing group names with this prefix, but not impossible. Since there is no particular cost (to us) in terms of making this lengthier/more unique, perhaps we could consider:

```suggestion
	const IGNORE_GROUP_FLAG     = '--*IGNORE*--';
```

Or some similar variation. It's a small and easy change on our side, but reduces the potential for collisions with existing group names considerably (since, if I am reading correctly, method `group_has_ignore_prefix` won't guard against existing group names).

classes/abstracts/ActionScheduler_Store.php
```diff
@@ -208,6 +209,40 @@ abstract public function get_claim_id( $action_id );
...
+	 *
+	 * @param string $slug Group slug name.
+	 * @return boolean
+	 */
+	protected static function group_has_ignore_prefix( $slug ) {
+		return self::IGNORE_GROUP_FLAG === substr( $slug, 0, strlen( self::IGNORE_GROUP_FLAG ) );
+	}
+
+	/**
+	 * Mark group for exclussion by prefixing with '-'.
```
barryhughes:
```suggestion
	 * Mark group for exclusion by applying a special prefix (by default, the value of
	 * the IGNORE_GROUP_FLAG property).
```

classes/abstracts/ActionScheduler_Store.php
```diff
@@ -208,6 +209,40 @@ abstract public function get_claim_id( $action_id );
...
+	 *
+	 * @param string $slug Group slug name.
+	 * @return string
+	 */
+	public static function sanitize_group_name( $slug ) {
+		// removing the ignored group prefix from slug if present.
+		$slug = preg_replace( '/(' . preg_quote( self::IGNORE_GROUP_FLAG, null ) . ')?(.*)/m', '$2', $slug );
+
+		// further sanitisation here with `sanitize_title` maybe?
+		return $slug;
```
barryhughes: Perhaps we should also anchor `^`the regex? Re sanitization; I wonder if `esc_sql` would (also) make sense?

classes/data-stores/ActionScheduler_DBStore.php
```diff
@@ -138,11 +138,18 @@ protected function get_group_id( $slug, $create_if_not_exists = true ) {
...
+	 * @throws \InvalidArgumentException Throws exception when the group slug starts with IGNORE_GROUP_FLAG.
 	 */
 	protected function create_group( $slug ) {
 		/** @var \wpdb $wpdb */
 		global $wpdb;
-		$wpdb->insert( $wpdb->actionscheduler_groups, array( 'slug' => $slug ) );
+
+		if ( self::group_has_ignore_prefix( $slug ) ) {
+			// translators: %s: ignored group name prefix.
+			throw new InvalidArgumentException( sprintf( __( 'Group names cannot start with %s.', 'action-scheduler' ), self::IGNORE_GROUP_FLAG ) );
```
```suggestion
			throw new InvalidArgumentException( sprintf( esc_html__( 'Group names cannot start with %s.', 'action-scheduler' ), self::IGNORE_GROUP_FLAG ) );
```

classes/data-stores/ActionScheduler_DBStore.php
```diff
@@ -622,7 +629,7 @@ protected function generate_claim_id() {
 	 * @param int       $limit Number of action to include in claim.
 	 * @param \DateTime $before_date Should use UTC timezone.
 	 * @param array     $hooks Hooks to filter for.
-	 * @param string    $group Group to filter for.
+	 * @param string    $group Group to filter for, adding IGNORE_GROUP_FLAG in front of the group will have it ignored.
```
barryhughes: Would this be more accurate?

```suggestion
	 * @param string    $group Comma-separated list of groups to filter by. Adding IGNORE_GROUP_FLAG in front of a group will cause it to be ignored.
```

classes/data-stores/ActionScheduler_DBStore.php
```diff
@@ -653,18 +660,38 @@ protected function claim_actions( $claim_id, $limit, \DateTime $before_date = nu
...
+					$include_group_ids[] = $group_id;
+				}
 			}
+		}
 
-			$where   .= ' AND group_id = %d';
-			$params[] = $group_id;
+		if ( count( $include_group_ids ) ) {
+			$where   .= ' AND group_id IN (%s)';
+			$params[] = implode( ',', $include_group_ids );
```
barryhughes, line 872: Asking this without actually testing 😅 ... but, won't the group ID arrays at this point be comma separated strings without quotation? That is, we will get:

```
( foo, bar, baz )
```

Instead of:

```
( 'foo', 'bar', 'baz' )
```

If so, we should add quoting at some suitable point.

barryhughes (commented):
Leaving some notes for when we pick this up at some future point ✍🏽

classes/data-stores/ActionScheduler_DBStore.php
ovidiul: The $include_group_ids will actually contain the numerical ID's of each group found in the DB, taken from above line

```
$group_id             = $this->get_group_id( $sanitized_group_name, false );
```

so I don't think we need to add any quotes here...

classes/abstracts/ActionScheduler_Store.php
ovidiul: Added the ^ regex change as well as the `esc_sql` filtering

ovidiul:
Thanks for the suggestions @barryhughes , I've addressed the mentioned issues. 👍

docs/wp-cli.md
```diff
@@ -29,7 +29,8 @@ These are the commands available to use with Action Scheduler:
     * `--batch-size` - This is the number of actions to run in a single batch. The default is `100`.
     * `--batches` - This is the number of batches to run. Using 0 means that batches will continue running until there are no more actions to run.
     * `--hooks` - Process only actions with specific hook or hooks, like `'woocommerce_scheduled_subscription_payment'`. By default, actions with any hook will be processed. Define multiple hooks as a comma separated string (without spaces), e.g. `--hooks=woocommerce_scheduled_subscription_trial_end,woocommerce_scheduled_subscription_payment,woocommerce_scheduled_subscription_expiration`
-    * `--group` - Process only actions in a specific group, like `'woocommerce-memberships'`. By default, actions in any group (or no group) will be processed.
+    * `--group` - Process only actions in a specific group, like `'woocommerce-memberships'`. By default, actions in any group (or no group) will be processed. Accepts comma-separated list of groups to filter by. Adding --*IGNORE*-- in front of a group will cause it to be ignored.
+    * `--ignore-group` - Ignore processing actions from groups found inside the comma separated list of values.
```
barryhughes, line 33: Nice, thank you :+1:

rrennick:
@ovidiul Thanks for writing this. It's a great feature to add. I wrote #905 as a base for adding more granular queue runner control starting with excluding a group. It would great if we had your feedback on the PR.

barryhughes:
Thanks for all the work here, and for pioneering a way forward! Closing, since most of the proposed functionality has been introduced via https://github.com/woocommerce/action-scheduler/pull/905 and https://github.com/woocommerce/action-scheduler/pull/918.

I want to also note that one thing those PRs do **not** add support for is processing multiple groups in a single queue runner, which there may still be value in.

PR #791, opened 2021-12-01
[PHP 8.1] Fix code that causes deprecation notices

Konamiman:
This pull request just fixes a few parts of the code that causes deprecation notices under PHP 8.1 (caused by passing `null` to the `DateTime` constructor and by [mismatching return types](https://php.watch/versions/8.1/ReturnTypeWillChange)) when running the WooCommerce core unit tests. See [the fix pull request in WooCommerce core](https://github.com/woocommerce/woocommerce/pull/31333).

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -313,7 +313,7 @@ function as_get_datetime_object( $date_string = null, $timezone = 'UTC' ) {
 	} elseif ( is_numeric( $date_string ) ) {
 		$date = new ActionScheduler_DateTime( '@' . $date_string, new DateTimeZone( $timezone ) );
 	} else {
-		$date = new ActionScheduler_DateTime( $date_string ?? 'now', new DateTimeZone( $timezone ) );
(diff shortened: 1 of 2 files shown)


functions.php
```diff
@@ -313,7 +313,7 @@ function as_get_datetime_object( $date_string = null, $timezone = 'UTC' ) {
 	} elseif ( is_numeric( $date_string ) ) {
 		$date = new ActionScheduler_DateTime( '@' . $date_string, new DateTimeZone( $timezone ) );
 	} else {
-		$date = new ActionScheduler_DateTime( $date_string, new DateTimeZone( $timezone ) );
+		$date = new ActionScheduler_DateTime( $date_string ?? 'now', new DateTimeZone( $timezone ) );
```
barryhughes, line 316: Note that this breaks under PHP 5.6, which [we still support](https://github.com/woocommerce/action-scheduler/blob/3.4.0/readme.txt#L8).

Konamiman: Ouch. Fixed.

barryhughes (approved):
LGTM!

Issue #787, opened 2021-11-17
Add custom User-Agent for Async Post requests


ovidiul:
When dispatching an [async](https://github.com/woocommerce/action-scheduler/blob/7f4aaf0e011bb0b41b266afd479f6935a495231b/lib/WP_Async_Request.php#L94) QueueRunner request, the User-Agent sent with that POST request usually defaults to the WordPress default value.

Right now, it's pretty hard to identify if the requests to the 'wp-admin/admin-ajax.php' endpoint came from Action Scheduler or another plugin that used similar request parameters.

Would it be worth adding a filter for the User-Agent as well as adjusting the default User-Agent sent by Action Scheduler async requests?

Happy to work on a PR for this.

peterfabian:
Sounds like a good idea, thanks @ovidiul .

barryhughes:
@peterfabian | @ovidiul 

Though I like this idea and *suspect* it is low risk, it does still feel like there is a chance of accidentally causing disruption in some cases (suppose there is a deployment out there where user agent strings for admin-ajax requests need to be on an approved list, or suppose internal admin-ajax requests are already being routed to a different box on the basis of their user agent string). 

Instead, we could flip this round and continue to implicitly use the default WP user agent string, while making it possible to customize things on a project-by-project or request-by-request basis. Two ways we might do this:

- **Low effort** → the existing PR [already takes care of making the user agent string filterable](https://github.com/woocommerce/action-scheduler/pull/789/commits/2c66e168d5e105c98f763926930e455ec242666f#diff-1bc85997145e00ddcc40255901b06d55830851841501e079682ab9f05d978eaeR144): we could tweak this and make it so this value is only added to the args array if it is set to something other than `null`.
- **Slightly more effort** → looking at the latest [upstream version of `WP_Async_Request`](https://github.com/deliciousbrains/wp-background-processing/blob/1.0.2/classes/wp-async-request.php#L154) the entire post args array (and many other things besides) are now filterable, whereas they weren't previously. We could take advantage of this.

If we took the second approach, we would likely also want to create a child class with its own custom identifier (and projects could then override the user agent string or other args by filtering on something like `as_async_request_post_args`).

Then again, perhaps I'm seeing risks that don't exist. What do you think?

ovidiul:
Thanks for sharing the details @barryhughes 

I do agree that this change might come in with potential backward compatibility issues, so probably changing the User Agent might be unsafe.

[...]

PR #794, opened 2022-01-21
WP_Async_Request upgrade source

ovidiul:
As detailed in issue #787, to address that issue and introduce the proposed User-Agent filters, it would be beneficial to upgrade the `WP_Async_Request` library from [upstream](https://github.com/deliciousbrains/wp-background-processing/blob/master/classes/wp-async-request.php) as that includes the needed filters already.

As noted in the commit code, there are no major changes to the source library other than expanding the original code with some `apply_filters`, so performance-wise, there should be no impact.

### Changelog

> Dev - Updated WP_Async_Request class from the WP Background Processing library, adding several new filters.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -105,3 +105,14 @@ It requires no setup, and won't override any WordPress APIs (unless you want it 
 Action Scheduler is designed to manage the scheduled actions on a single site. It has no special handling for running queues across multiple sites in a multisite network. That said, because its storage and Queue Runner are completely swappable, it would be possible to write multisite handling classes to use with it.
 
 If you'd like to create a multisite plugin to do this and release it publicly to help others, [open a new issue to let us know](https://github.com/woocommerce/action-scheduler/issues/new), we'd love to help you with it.
+
+
+### How can I change the Action Scheduler User-Agent to better identify its requests?
+
+function eg_define_custom_user_agent( $args ) {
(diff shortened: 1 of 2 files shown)


docs/faq.md
```diff
@@ -105,3 +105,19 @@ It requires no setup, and won't override any WordPress APIs (unless you want it
 Action Scheduler is designed to manage the scheduled actions on a single site. It has no special handling for running queues across multiple sites in a multisite network. That said, because its storage and Queue Runner are completely swappable, it would be possible to write multisite handling classes to use with it.
 
 If you'd like to create a multisite plugin to do this and release it publicly to help others, [open a new issue to let us know](https://github.com/woocommerce/action-scheduler/issues/new), we'd love to help you with it.
+
+### How can I change the Action Scheduler User-Agent to better identify its requests?
+
+Action Scheduler has a filter available named `as_async_request_queue_runner_post_args` which can be used to filter the arguments that are being sent to the `wp_remote_post` call.
+
+The User-Agent parameter is just one of them and can be adjusted as:
```
barryhughes, line 113: Minor, but maybe:

```suggestion
The User-Agent parameter is just one of them and can be adjusted as follows:
```

barryhughes: (Went ahead and committed this change, but let me know if any concerns.)

barryhughes (approved):
LGTM, thank you for this change :-)

PR #803, opened 2022-02-16
Fixes #802 cancel action fatals

### LINKED ISSUE #802: `ActionScheduler_DBStore::cancel_action()` fatals
Hopefully I can provide enough information here for fixing my issue, because it's unclear to me how it's actually being triggered...

Our WooCommerce started logging exceptions (fatals) when we upgraded WooCommerce Payments, and the problem comes from the `DBStore::cancel_action()` function when an `$action_id` has already been cancelled. (The update returns 0 rows updated correctly, but then it incorrectly throws an 'Unspecified action' exception.

Would it make more sense to do the following?

```php
# currently line 449
if ( false === $updated ) {
	/* translators: %s: action ID */
	throw new \InvalidArgumentException( sprintf( __( 'Unidentified action %s', 'woocommerce' ), $action_id ) );
}
```

n3f:
Compare database update boolean result when determining to throw an exception (on cancel action).

### Changelog

> Fix - When cancelling actions, only throw an exception in the event of a database error and improve handing of this from within as_unschedule_action().

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -115,15 +115,23 @@ function as_unschedule_action( $hook, $args = array(), $group = '' ) {
 		$params['args'] = $args;
 	}
 
-	try {
-		$action_id = ActionScheduler::store()->query_action( $params );
-		if ( $action_id ) {
+	$action_id = ActionScheduler::store()->query_action( $params );
+
+	if ( $action_id ) {
+		try {
 			ActionScheduler::store()->cancel_action( $action_id );
+		} catch ( Exception $exception ) {
+			ActionScheduler::logger()->log(
+				$action_id,
+				sprintf(
+					/* translators: %s is the name of the hook to be cancelled. */
+					__( 'Caught exception while cancelling action: %s', 'action-scheduler' ),
+					esc_attr( $hook )
+				)
+			);
+
+			$action_id = null;
 		}
-	} catch ( Exception $exception ) {
-		$message = sprintf( __( 'Caught exception while cancelling action: %s', 'action-scheduler' ), esc_attr( $hook ) );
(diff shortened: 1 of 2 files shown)


functions.php
```diff
@@ -115,9 +115,15 @@ function as_unschedule_action( $hook, $args = array(), $group = '' ) {
...
+	try {
+		$action_id = ActionScheduler::store()->query_action( $params );
+		if ( $action_id ) {
+			ActionScheduler::store()->cancel_action( $action_id );
+		}
+	} catch ( Exception $exception ) {
+		$message = sprintf( __( 'Caught exception while cancelling action: %s', 'action-scheduler' ), esc_attr( $hook ) );
+		error_log( $message, E_WARNING );
+		$action_id = null;
 	}
```
barryhughes, line 127: Thinking we should be consistent in where we log things, and also because I think it is safe to move the `query_action()` call out of the `try {}` structure (afaik, it does not itself throw any exceptions), what do you think of the following tweaks?

```suggestion
	$action_id = ActionScheduler::store()->query_action( $params );

	if ( $action_id ) {
		try {
			ActionScheduler::store()->cancel_action( $action_id );
		} catch ( Exception $exception ) {
			ActionScheduler::logger()->log(
				$action_id,
				sprintf(
					/* translators: %s is the name of the hook to be cancelled. */
					__( 'Caught exception while cancelling action: %s', 'action-scheduler' ),
					esc_attr( $hook )
				)
			);

			$action_id = null;
		}
	}
```

Almost the same, but uses Action Scheduler's own logger (and some minor stuff, like a translator note).

barryhughes (approved):
Looks good, thank you!

PR #806, opened 2022-03-11
Throw an exception (mark as failed) scheduled actions without any callbacks

### LINKED ISSUE #805: Detect and Handle "Abandoned Actions"
Scenario:

* Plugin Foo adds a recurring/cron action with the hook `bar`.
* Plugin Foo is removed (and either does not perform clean-up or does not have the opportunity to do so).
* Action Scheduler will continue to execute and complete action `bar` and, each time, will create a replacement action (according to the specified schedule), *even though Plugin Foo is no longer present and no callbacks for 'bar' are being called.*

✍🏽 [Relevant queue runner code](https://github.com/woocommerce/action-scheduler/blob/3.4.0/classes/abstracts/ActionScheduler_Abstract_QueueRunner.php#L50-L80) | Internal link for further context: p9o2xV-1UT-p2

---

### Possible solution

- WP functions `do_action()` and `do_action_ref_array()` return void: they don't let us know if any callbacks were actually called.
- We can however test before hand if any callbacks are registered, using `has_action()`.
- By using this to detect an absence of callbacks, we could skip triggering the action and instead mark it as failed.
- From there, some logic (which could be added to the [Queue Cleaner](https://github.com/woocommerce/action-scheduler/blob/3.4.0/classes/ActionScheduler_QueueCleaner.php)) could identify recurring/cron actions where the most recent _x_ attempts were marked as failures and then cancel the action completely.

I don't know exactly what threshold makes sense for _x_ ... but it can anyway be made configurable or filterable. 

This would potentially solve problems where recurring actions continue to be perpetuated even when the code that created them no longer exists (a problem that is exacerbated by the fact that race conditions, documented elsewhere, may in some cases lead to duplicate recurring actions building up over time).

barryhughes:
This change is geared toward handling of scheduled actions that do not have any callbacks and, more broadly, recurring actions that continually and consistently fail.

Example: imagine a plugin has scheduled an hourly recurring action, but that same plugin is subsequently deleted without any opportunity to perform clean-up (or, it is deactivated normally but does not contain uninstall logic). Currently, those recurring actions will continue to fire and:

- Each time the action status will be updated to `complete` (so it will seem like they are running successfully).
- A replacement action will be scheduled.

Therefore, we can end up with a category of repeating zombie actions, clogging up the queue unnecessarily. This change attempts to address that problem and closes #805.

### How it works

Just because an action fails once doesn't mean it will always fail: so, we need to bake in some amount of tolerance. The way I propose we handle this is two-fold:

- When an action does not have any callbacks, mark it as `failed`.
- If it is a recurring action, check to see if it is *consistently* failing and, if so, don't reschedule it.
    - By default, the criteria for *"consistent failure"* is when the most recent 5 actions with the same hook, argument set, and schedule have all failed.
    - This will be adjustable via a filter.

### Testing

To make testing easier, I suggest temporarily stopping Action Scheduler from being triggered automatically:

- Add `define( 'DISABLE_WP_CRON', true );` to your `wp-config.php` file.
- Add `add_filter( 'action_scheduler_allow_async_request_runner', '__return_false' );` to a suitable location, such as a mu-plugin file.
- If you have any crontab entries that provide alternative ways of running cron or the Action Scheduler queue, suspend them.
- Also, to give yourself a 'clean view' of things, you may also wish to delete all existing actions: `wp db query "DELETE FROM $(wp db prefix)actionscheduler_actions"`

Now, processing of the queue should be entirely manual. Next:

- Create a new recurring action without a callback: `wp eval "as_schedule_recurring_action( time() - 1, 2, 'no_callback' );"`
- Visit **Tools ▸ Scheduled Actions** and locate action `no_callback` (should be pending), then run it manually.
- It should fail, and a replacement should be scheduled.
- Rinse and repeat a few times.
- In the end, you should find you have 5 failed scheduled actions for the hook `no_callback` and no more pending actions of that name should have been scheduled.

### Notes

- This is one facet of an effort to address zombie actions; we need to follow-up with further work that prevents unintentional duplication of actions (via race conditions).
- You will notice a large number of updates to tests that seem to be unrelated: that's because those tests relied on the previous behavior of Action Scheduler, where actions would be marked complete even if they didn't have a callback. Most of these updates are isolated inside commits https://github.com/woocommerce/action-scheduler/pull/806/commits/8276036dd57d3ba8ec8249e55a386c5acb1bfc4c and https://github.com/woocommerce/action-scheduler/pull/806/commits/37b8a1efb8911efc657609c0aa11c3392e3a4b86. The assertions specific to the core change can be [found here](https://github.com/woocommerce/action-scheduler/pull/806/files#diff-2fd59945ebe22bbea56dd19195fec979d5ee027dc705ac6b6af4eb01defad288).

### Changelog

> Enhancement - If a recurring action is found to be consistently failing, it will stop being rescheduled.

### DIFF
--- a/classes/abstracts/ActionScheduler_Abstract_QueueRunner.php
+++ b/classes/abstracts/ActionScheduler_Abstract_QueueRunner.php
@@ -115,45 +115,40 @@ abstract class ActionScheduler_Abstract_QueueRunner extends ActionScheduler_Abst
 	 * @return bool
 	 */
 	private function recurring_action_is_consistently_failing( ActionScheduler_Action $action, $action_id ) {
-		// Fetch the most recent historic actions having the same hook and argument set.
-		$previous_action_ids = $this->store->query_actions(
-			array(
-				'hook'         => $action->get_hook(),
-				'args'         => $action->get_args(),
-				'date'         => date_create( 'now', timezone_open( 'UTC' ) )->format( 'Y-m-d H:i:s' ),
-				'date_compare' => '<',
-				'per_page'     => 10,
-			)
-		);
-
 		/**
-		 * The number of failed actions to count when determining if a recurring action has been consistently failing.
+		 * Controls the failure threshold for recurring actions.
 		 *
-		 * @param int $threshold
+		 * Before rescheduling a recurring action, we look at its status. If it failed, we then check if all of the most
+		 * recent instances (upto the threshold set by the filter) have also failed, in which case we will not log a
+		 * further action.
+		 *
+		 * @param int $failure_threshold Number of actions of the same hook to examine for failure. Defaults to 5.
 		 */
 		$consistent_failure_threshold = (int) apply_filters( 'action_scheduler_recurring_action_failure_threshold', 5 );
-		$failed_actions               = 0;
 
-		foreach ( $previous_action_ids as $previous_action_id ) {
-			$previous_action = $this->store->fetch_action( $previous_action_id );
+		// This query should find the earliest *failing* action (for the hook we are interested in) within our threshold.
+		$query_args = array(
+			'hook'         => $action->get_hook(),
+			'status'       => ActionScheduler_Store::STATUS_FAILED,
+			'date'         => date_create( 'now', timezone_open( 'UTC' ) )->format( 'Y-m-d H:i:s' ),
+			'date_compare' => '<',
+			'per_page'     => 1,
+			'offset'       => $consistent_failure_threshold
+		);
 
-			// Check if the schedules match one another.
-			if ( ! $this->schedules_match( $previous_action->get_schedule(), $action->get_schedule() ) ) {
-				continue;
-			}
+		$first_failing_action_id = $this->store->query_actions( $query_args );
 
-			// If we encounter a single non-failed action before our threshold is met, then we do not consider the
-			// action to be consistently failing.
-			if ( ActionScheduler_Store::STATUS_FAILED !== $this->store->get_status( $previous_action_id ) ) {
-				return false;
-			}
-
-			if ( ++$failed_actions === $consistent_failure_threshold ) {
-				return true;
-			}
+		// If we didn't retrieve an action ID, then there haven't been enough failures for us to worry about.
+		if ( empty( $first_failing_action_id ) ) {
+			return false;
 		}
 
-		return false;
+		// Now let's fetch the first action (having the same hook) of *any status*ithin the same window.
+		unset( $query_args['status'] );
+		$first_action_id_with_the_same_hook = $this->store->query_actions( $query_args );
+
+		// If the IDs match, then actions for this hook must be consistently failing.
+		return $first_action_id_with_the_same_hook === $first_failing_action_id;
 	}
 
 	/**
--- a/classes/abstracts/ActionScheduler_Abstract_QueueRunner.php
+++ b/classes/abstracts/ActionScheduler_Abstract_QueueRunner.php
@@ -152,24 +152,6 @@ abstract class ActionScheduler_Abstract_QueueRunner extends ActionScheduler_Abst
 	}
 
 	/**
-	 * Specifically intended for use with recurring actions, this method assesses if they are of the same type
-	 * (for instance, they are both CronSchedule objects) and if the recurrence patterns are the same.
-	 *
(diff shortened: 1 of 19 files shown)


barryhughes:
Started adding code to guard against rescheduling a consistently failing recurring action. Need to add additional test-cases.

classes/abstracts/ActionScheduler_Abstract_QueueRunner.php
```diff
@@ -86,13 +86,94 @@ public function process_action( $action_id, $context = '' ) {
...
+	/**
+	 * Determine if the specified recurring action has been consistently failing.
+	 *
+	 * @param ActionScheduler_Action $action    The recurring action to be rescheduled.
+	 * @param int                    $action_id The ID of the recurring action.
+	 *
+	 * @return bool
+	 */
+	private function recurring_action_is_consistently_failing( ActionScheduler_Action $action, $action_id ) {
+		// Fetch the most recent historic actions having the same hook and argument set.
```
vedanshujain: May I suggest a bit more performant algorithm to solve this method:
1. Get the last `n`th action ID that failed for the same hook and args, where n is `action_scheduler_recurring_action_failure_threshold` using `per_page = 1`, `status = failed` and `offset = n`.  If there isn't any then`return false`. Store this action ID somewhere.
2. Get the last `n`th action ID with the same hook and args, but this time don't pass the `status = failed` filter. If the last n actions have been consistently failing, then both these ID have to be the same. So return true if this ID is the same as the previous one otherwise return false.

This will be more performant than the current query because you can do the comparison in just 2 queries, instead of `n+1` queries like we do now (and hopefully bit less code). This won't take schedule into account, but I think that's fine as we use the method in deciding whether to schedule another recurring action or not (so we know that the current action is recurring for sure).

vedanshujain (commented):
Great work, I have added an optional suggestion but looks pretty good.

classes/abstracts/ActionScheduler_Abstract_QueueRunner.php
barryhughes: @vedanshujain—that's a great idea, thank you.

> This won't take schedule into account, but I think that's fine as we use the method in deciding whether to schedule another recurring action or not (so we know that the current action is recurring for sure).

My only nit would be that, with this strategy, we don't know if all of those failing actions were recurring—we will only know that they have the same hook (ie, this algorithm does not account for a mix of single and recurring actions using the same hook).

Even so, it is probably still a very reasonable way to go. Talking it out some more:

- The specifics of the schedule (beyond the fact we're primarily interested in recurring actions) probably don't matter all that much: after all, when an action is triggered, the callback doesn't know anything about the schedule—it just knows what its arguments are.
- Taking this a step further, I am now wondering if checking that the args match is actually the best policy, because it means we will fail to detect and halt a failing sequence like the following:

| Hook | Args | Status |
| --- | --- | --- |
| foo | [ 'page' => 1 ] | failed |
| foo | [ 'page' => 2 ] | failed |
| foo | [ 'page' => 3 ] | failed |
| ... | ... | ... |

To that end, I think I will:

- Adopt the strategy you proposed
- Perform the matching based only on the hook and not the args

Let me know if you have thoughts on this!

barryhughes: Updated in https://github.com/woocommerce/action-scheduler/pull/806/commits/75b2766085b614d79699acfaacd29a1617cc1943

barryhughes:
Thanks @vedanshujain. Ready for another look! Principal changes are [here (algorithm)](https://github.com/woocommerce/action-scheduler/pull/806/files#diff-63b59dbe3f432227cb577929975389b1b2b929d1a8adbbf2ac0e2517d7e10066R109-R152) and [here (assertions)](https://github.com/woocommerce/action-scheduler/pull/806/files#diff-2fd59945ebe22bbea56dd19195fec979d5ee027dc705ac6b6af4eb01defad288R192-R243):

- Hopefully a more efficient algorithm per your suggestion.
- No longer matches on arguments or on schedule: reasoning is that if a callback for `some_hook` consistently fails, we want to stop it from propagating. Precisely matching the args and schedule set the bar to a level that may make this preventative measure less effective.

vedanshujain (approved):
Looks neat!!

Issue #729, opened 2021-07-11
Action hook after marked complete


noplanman:
Is there a way to hook in after an action has been marked `complete`?

As it is now, there is only a `action_scheduler_after_execute` which gets fired just before marking the action complete.

Use case:
I have a bunch of actions with the same hook and I'd like to perform an action after all actions have completed.
I tried with `as_next_scheduled_action` but that always gives me `true` because the action that is being worked on still counts as `busy`, not `completed`, which makes sense.

How can I perform an action after all actions with the same hook have been completed?
(Apart from using the `action_scheduler_after_execute` hook and checking for 1 action in the queue)

noplanman:
I've found a solution that works, not sure if it's the prettiest or gentlest on the DB with lots of queued actions though.

I'm open for other suggestions! 

```php
add_action( 'action_scheduler_after_execute', function ( int $action_id, \ActionScheduler_Action $action, $context ) {
	if ( ! as_get_scheduled_actions( [
		'hook'   => $action->get_hook(),
		'group'  => $action->get_group(),
		'status' => \ActionScheduler_Store::STATUS_PENDING,
	] ) ) {
		// Ok, queue is empty!
	}
}, 10, 3 );
```

There's also a `action_scheduler_after_process_queue` action, but that's a general queue processing which doesn't pass any data that I can use.

barryhughes:
Hi there! 

Not sure there is *quite* enough context to fully understand your needs, but here are some general thoughts in case useful:

* The `'action_scheduler_after_process_queue'` action fires after each batch is processed (so you could possibly perform your check at that stage: if one or more actions with the hook you are interested in monitoring were processed, the status changes will actually have been made by this point—in contrast to `'action_scheduler_after_execute'`).
* I'm not sure if you are only interested in completed actions (and not failed ones) but you may also be interested in either or both of these actions:
    * `'action_scheduler_failed_execution'`
    * `'action_scheduler_failed_validation'`
* Last but not least, though I'm not sure what sort of work these actions are doing, perhaps there is another way to monitor progress? Or, you might try using a recurring action to periodically 'poll' and check on the overall job status.

We can leave this open for a while in case others have ideas they wish to share, but note that we'll probably circle black and close the issue in due course (in its current form it is essentially a custom dev support request, whereas our primary focus here is on bug requests followed by feature requests).

noplanman:
@barryhughes Thanks so much for your feedback!

To give some more context, and my specific use-case.

[...]

PR #813, opened 2022-04-27
update: Added Action scheduler completed action hook

ManiruzzamanAkash:
- [Enhancement] Added Action scheduler completed action hook.

I think added a simple hook when completed an action would solve many use cases pretty much simpler. And as there is no normal way to detect if actions is completed, this hook could be integrated.

Discussion in #729 

And for reference, same way, there is already a hook for `Canceled action` on `line-460`
```php
do_action( 'action_scheduler_canceled_action', $action_id );
```

So, it would be awesome if the new hook is integrated for completed action.

### Changelog 

> Enhancement - Add new action hook `action_scheduler_completed_action`.

### DIFF
--- a/classes/data-stores/ActionScheduler_DBStore.php
+++ b/classes/data-stores/ActionScheduler_DBStore.php
@@ -841,7 +841,7 @@ class ActionScheduler_DBStore extends ActionScheduler_Store {
 		}
 
 		/**
-		 * Fires after an action schedule is being completed.
+		 * Fires after a scheduled action has been completed.
 		 *
 		 * @since 3.4.2
 		 *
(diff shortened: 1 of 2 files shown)


barryhughes (changes requested):
Thanks!

Ideally, the WP Post store should also be updated with the same hook. It would also be worth [documenting the hook](https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#4-hooks-actions-and-filters).

ManiruzzamanAkash:
@barryhughes Thanks for your feedback.
I've not added the docblock to keep the consistency with other similar hooks(cancel). However, added the docblock.

barryhughes:
Thanks!

Some of the older code does not meet our current standards, but where possible for new code we do try to implement these things.

Note that we would still want the [WP Post Store](https://github.com/woocommerce/action-scheduler/blob/3.4.1/classes/data-stores/ActionScheduler_wpPostStore.php) to be updated with the same functionality, to avoid problems arising from a discrepancy between the two.

barryhughes (approved):
Thanks!

PR #815, opened 2022-05-02
Change the include for better linting.

artpi:
There is no functional change.

This PR changes include style so that static linters detect the $dir in order to determine that this include path is not relative.
Otherwise, on WPCOM deploy linter complaints.

## Testing instructions

Automated tests


millerf (approved):
LGTM

PR #822, opened 2022-06-03
Make Action status links clickable

### LINKED ISSUE #685: Make current subsubsub link clickable
ie. When on the Pending status-page - The "**Pending**" item is not a link. It should be a link.
That's how WordPress does it. That's how all plugins do it.

bahiirwa:
## Motivation
Adds a clickable link for all the statuses, pending, failed, in-progress etc.
Fixes https://github.com/woocommerce/action-scheduler/issues/685

### Pre-patch
Screenshot(10)

### Post Patch
Screenshot(11)

## Other issues for discussion
- Do we remove the strong tag that highlights the current link and use the grey pre-patch instead of the default blue?

## Changelog

> Add - The active view link within the "Tools > Scheduled Actions" screen is now clickable.

### DIFF
--- a/classes/abstracts/ActionScheduler_Abstract_ListTable.php
+++ b/classes/abstracts/ActionScheduler_Abstract_ListTable.php
@@ -683,7 +683,7 @@ abstract class ActionScheduler_Abstract_ListTable extends WP_List_Table {
 			}
 
 			if ( $status_name === $request_status || ( empty( $request_status ) && 'all' === $status_name ) ) {
-				$status_list_item = '<li class="%1$s"><strong>%3$s</strong> (%4$d)</li>';
+				$status_list_item = '<li class="%1$s"><a href="%2$s"><strong>%3$s</strong></a> (%4$d)</li>';
 			} else {
 				$status_list_item = '<li class="%1$s"><a href="%2$s">%3$s</a> (%4$d)</li>';
 			}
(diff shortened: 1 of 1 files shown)


classes/abstracts/ActionScheduler_Abstract_ListTable.php
```diff
@@ -683,7 +683,7 @@ protected function display_filter_by_status() {
 			}
 
 			if ( $status_name === $request_status || ( empty( $request_status ) && 'all' === $status_name ) ) {
-				$status_list_item = '<li class="%1$s"><strong>%3$s</strong> (%4$d)</li>';
+				$status_list_item = '<li class="%1$s"><a href="%2$s"><strong>%3$s</strong></a> (%4$d)</li>';
```
barryhughes: To more closely match other WP Admin List Tables, I thint it makes sense to:

- Add class `current` to the `<a>` tag
- Drop the use of `<strong>` (since the above class should take care of formatting)

So:

```suggestion
				$status_list_item = '<li class="%1$s"><a href="%2$s" class="current">%3$s</a> (%4$d)</li>';
```

This should give us output that will render something like the following:

subsubsub-links-current-screen-example

What do you think?

barryhughes, line 686:
```suggestion
				$status_list_item = '<li class="%1$s"><a href="%2$s" class="current">%3$s</a> (%4$d)</li>';
```

↑ I think this will give us better consistency with other list tables.

bahiirwa: Agreed. 

## Tests:
Manual tests done to see the effect. The other links adopt the class on being clicked like the all.

## Screenshot
Screenshot 2022-06-14 at 09 59 55

barryhughes (approved):
Looks great, thank you.

PR #827, opened 2022-06-15
Take advantage of `wp_cache_flush_runtime()` when available.

### LINKED ISSUE #790: Race conditions with internal and external object cache
Since a CLI worker can keep running indefinitely when there are enough jobs there is a high chance of running into race conditions with for example WooCommerce. This happens because other processes invalidating or changing the cache will do it on the external one and won't ever be picked up by the internal one usually stored in **$wp_object_cache->cache**. In the case of WooCommerce this could with some gateways result in missing order items or wrong calculations under high load.

AS by default clears the object cache when an external object cache isn't in use after **do_batch**, but this is mainly done to save memory. As these types of bugs are pretty hard for users to figure out how is happening I would suggest that AS implements some way to solve it for most of the popular object cache dropins.

There isn't an standardized way of clearing just the internal object cache when external ones is in use, but for some of the dropins this filter should get the work done without affecting how anything works. A similar logic can be used directly in action scheduler.

```
add_filter( 'action_scheduler_queue_runner_flush_cache', function( $flush ) {
	global $wp_object_cache;

	if ( ! wp_using_ext_object_cache() || ! ( isset( $wp_object_cache->cache ) && is_array( $wp_object_cache->cache ) ) ) {
		return $flush;
	}

	$wp_object_cache->cache = []; // reset the internal cache, without flushing the persistent object cache

	return false;
} );
```

barryhughes:
WordPress 6.0.0 introduced [`wp_cache_flush_runtime()`](https://core.trac.wordpress.org/changeset/52772) as a way to flush the runtime cache without impacting the external object cache.

This is especially useful for long-running CLI processes (example: Action Scheduler's WP CLI queue runner) though also for long-running processes in general. It provides a way to wipe the in-memory (non-persistent) runtime cache without touching the external (persistent) object cache:

- This can reduce memory usage.
- Perhaps more importantly, it can prevent invalid values from building up in the in-memory cache and being presented to application code as if they are still valid.
- It does it in a way that avoids a negative performance impact for other processes (because if the persistent external cache is invalidated, all kinds of values now need to be recalculated).

Notes:

- We will not try to use `wp_cache_flush_runtime()` unless it is available (which would be because we are running on WordPress 6.0, or because a caching plugin provides the function as a 'polyfill').

Closes #790.

---

### Changelog 

> Enhancement - Improvements to cache invalidation when processing batches (when running on WordPress 6.0+).

### DIFF
--- a/classes/ActionScheduler_QueueRunner.php
+++ b/classes/ActionScheduler_QueueRunner.php
@@ -173,15 +173,34 @@ class ActionScheduler_QueueRunner extends ActionScheduler_Abstract_QueueRunner {
 	}
 
 	/**
-	 * Running large batches can eat up memory, as WP adds data to its object cache.
+	 * Flush the cache if possible (intended for use after a batch of actions has been processed).
 	 *
-	 * If using a persistent object store, this has the side effect of flushing that
-	 * as well, so this is disabled by default. To enable:
-	 *
-	 * add_filter( 'action_scheduler_queue_runner_flush_cache', '__return_true' );
+	 * This is useful because running large batches can eat up memory and because invalid data can accrue in the
+	 * runtime cache, which may lead to unexpected results.
 	 */
 	protected function clear_caches() {
-		if ( ! wp_using_ext_object_cache() || apply_filters( 'action_scheduler_queue_runner_flush_cache', false ) ) {
+		/*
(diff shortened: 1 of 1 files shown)


jorgeatorres (approved):
LGTM. Thanks @barryhughes!

PR #830, opened 2022-06-20
Schedule next action before current one is mark completed/failed.

### LINKED ISSUE #698: Scheduling just one recurring action makes more than one
On the Usage page there is an example of how to "schedule a function to run at midnight if it is not already scheduled". See here
https://actionscheduler.org/usage/

I have used this example to create an action that runs every minute, and only one gets scheduled. However, every once in a while I check on the scheduled actions and there are two of those running. My guess is that there must be a race condition two requests are running at the same time and end up scheduling two upcoming actions. They then stay in perpetuity.  

What is the best way to keep this from happening?

vedanshujain:
This prevents race conditions that may happen if we are trying to schedule the same action, and the check `as_has_scheduled_action` runs after the action is marked complete but before the next one is scheduled.

Fixes #698

### DIFF
--- a/classes/abstracts/ActionScheduler_Abstract_QueueRunner.php
+++ b/classes/abstracts/ActionScheduler_Abstract_QueueRunner.php
@@ -64,16 +64,26 @@ abstract class ActionScheduler_Abstract_QueueRunner extends ActionScheduler_Abst
 			$this->store->log_execution( $action_id );
 			$action->execute();
 			do_action( 'action_scheduler_after_execute', $action_id, $action, $context );
+			$this->maybe_schedule_next_instance( $action, $action_id );
 			$this->store->mark_complete( $action_id );
 		} catch ( Exception $e ) {
 			if ( $valid_action ) {
+				$this->maybe_schedule_next_instance( $action, $action_id );
 				$this->store->mark_failure( $action_id );
 				do_action( 'action_scheduler_failed_execution', $action_id, $e, $context );
 			} else {
 				do_action( 'action_scheduler_failed_validation', $action_id, $e, $context );
 			}
 		}
+	}
 
+	/**
+	 * Helper method to schedule next instance of an instance if there is a schedule.
+	 *
+	 * @param ActionScheduler_Action $action    Action object.
+	 * @param int                    $action_id Action ID.
+	 */
+	private function maybe_schedule_next_instance( $action, $action_id ) {
 		if ( isset( $action ) && is_a( $action, 'ActionScheduler_Action' ) && $action->get_schedule()->is_recurring() ) {
 			$this->schedule_next_instance( $action, $action_id );
 		}


classes/abstracts/ActionScheduler_Abstract_QueueRunner.php
```diff
@@ -64,16 +64,26 @@ public function process_action( $action_id, $context = '' ) {
 			$this->store->log_execution( $action_id );
 			$action->execute();
 			do_action( 'action_scheduler_after_execute', $action_id, $action, $context );
+			$this->maybe_schedule_next_instance( $action, $action_id );
 			$this->store->mark_complete( $action_id );
 		} catch ( Exception $e ) {
 			if ( $valid_action ) {
+				$this->maybe_schedule_next_instance( $action, $action_id );
```
barryhughes, line 71: Annotated this block with comments at the end of each line:

```php
	$this->maybe_schedule_next_instance( $action, $action_id );          # 1. Next instance is scheduled
	$this->store->mark_complete( $action_id );                           # 2. Exception can be thrown by this method
} catch ( Exception $e ) {                                               # 3. Execution enters the catch block
	if ( $valid_action ) {                                               # 4. Action still 'valid'
		$this->maybe_schedule_next_instance( $action, $action_id );      # 5. We schedule another instance
```

↑ This chain of events seems unlikely, but still technically possible and could result in a duplicate.

classes/abstracts/ActionScheduler_Abstract_QueueRunner.php
```diff
@@ -64,16 +64,26 @@ public function process_action( $action_id, $context = '' ) {
...
 				$this->store->mark_failure( $action_id );
```
barryhughes, line 72: Thinking of [PR #806](https://github.com/woocommerce/action-scheduler/pull/806), it may be better to wait until the action has been marked as a failure **before** we call `::maybe_schedule_next_instance()` (though, if this is merged first we can also update that other PR accordingly).

classes/abstracts/ActionScheduler_Abstract_QueueRunner.php
```diff
@@ -64,16 +64,26 @@ public function process_action( $action_id, $context = '' ) {
...
+	}
 
+	/**
+	 * Helper method to schedule next instance of an instance if there is a schedule.
+	 *
+	 * @param ActionScheduler_Action $action    Action object.
+	 * @param int                    $action_id Action ID.
+	 */
+	private function maybe_schedule_next_instance( $action, $action_id ) {
 		if ( isset( $action ) && is_a( $action, 'ActionScheduler_Action' ) && $action->get_schedule()->is_recurring() ) {
```
barryhughes, line 87: Does testing if `$action` is set still make sense, now this is in its own method?

PR #831, opened 2022-06-20
 Add support for scheduling unique actions.

### LINKED ISSUE #828: Support creation of unique actions
### The problem

The following snippet is a fairly common example of creating a recurring action, where at any given time we only want there to be *a single pending (or in progress) action of this type.* The idea is that this snippet can run during `init` on every request and, in ideal circumstances, duplicate actions will be prevented by the `as_has_scheduled_action()` check inside the callback:

```php
function eg_schedule_midnight_log() {
	if ( false === as_has_scheduled_action( 'eg_midnight_log' ) ) {
		as_schedule_recurring_action( strtotime( 'tomorrow' ), DAY_IN_SECONDS, 'eg_midnight_log' );
	}
}
add_action( 'init', 'eg_schedule_midnight_log' );
```

The above is, in fact, taken from [our own docs](https://actionscheduler.org/usage/) *[[source]](https://github.com/woocommerce/action-scheduler/blob/17250b493dc14bece2ae5f11b663227bb09efcef/docs/usage.md?plain=1#L106-L115)*. In practice, it falls a little short—because it's a race condition: two concurrent processes could each test for an existing action at approximately the same time, and both could go on to schedule a future action (causing duplication). Because of this, some users of the library adapt the above into the following form:

```php
function eg_schedule_midnight_log() {
	if ( false === as_has_scheduled_action( 'eg_midnight_log' ) ) {
		as_unschedule_all_actions( 'eg_midnight_log' );
		as_schedule_recurring_action( strtotime( 'tomorrow' ), DAY_IN_SECONDS, 'eg_midnight_log' );
	}
}
add_action( 'init', 'eg_schedule_midnight_log' );
```

This is still suboptimal, at best it narrows the race condition a little and prevents the excessive build-up of duplicates.

### Proposal

Introduce new public API functions or update our existing ones (implementation TBD) to support the creation of unique pending actions (in the sense that, only one pending action for the specified hook is allowed to exist). 

If we introduce new functions for this purpose, they could be closely modelled on our existing ones:

- `as_enqueue_unique_async_action( $hook, $args = array(), $group = '' )`
- `as_schedule_unique_cron_action( $timestamp, $schedule, $hook, $args = array(), $group = '' )`
- `as_schedule_unique_recurring_action( $timestamp, $interval_in_seconds, $hook, $args = array(), $group = '' )`
- `as_schedule_unique_single_action( $timestamp, $hook, $args = array(), $group = '' )`

Or, we could revise our existing functions and add an optional `$unique` param:

- `as_enqueue_async_action( $hook, $args = array(), $group = '', $unique = false )`
- `as_schedule_cron_action( $timestamp, $schedule, $hook, $args = array(), $group = '', $unique = false )`
- `as_schedule_recurring_action( $timestamp, $interval_in_seconds, $hook, $args = array(), $group = '', $unique = false )`
- `as_schedule_single_action( $timestamp, $hook, $args = array(), $group = '', $unique = false )`

This would remove the burden of enforcing uniqueness from the library consumer to the library itself, which can then develop an optimized means of meeting this need.

vedanshujain:
As outlined in #828, we want to add support for scheduling unique actions in the action scheduler. This PR implements the second approach by adding another param in scheduling functions that specify whether the actions should be unique or not. 
For backward compatibility, the default value of that param is `false`.

Fixes #828.

### Changelog

> Enhancement - Added the ability to schedule unique actions via an atomic operation.

### DIFF
--- a/classes/ActionScheduler_ActionFactory.php
+++ b/classes/ActionScheduler_ActionFactory.php
@@ -70,18 +70,35 @@ class ActionScheduler_ActionFactory {
 	}
 
 	/**
-	 * @param string $hook The hook to trigger when this action runs
-	 * @param array $args Args to pass when the hook is triggered
-	 * @param int $when Unix timestamp when the action will run
-	 * @param string $group A group to put the action in
+	 * Create single action.
 	 *
-	 * @return int The ID of the stored action
+	 * @param string $hook  The hook to trigger when this action runs.
+	 * @param array  $args  Args to pass when the hook is triggered.
+	 * @param int    $when  Unix timestamp when the action will run.
+	 * @param string $group A group to put the action in.
+	 *
+	 * @return int The ID of the stored action.
 	 */
 	public function single( $hook, $args = array(), $when = null, $group = '' ) {
+		return $this->single_with_unique( $hook, false, $args, $when, $group );
+	}
+
+	/**
+	 * Create single action only if there is no pending or running action with same name and params.
+	 *
+	 * @param string $hook The hook to trigger when this action runs.
+	 * @param bool   $unique Whether action scheduled should be unique.
+	 * @param array  $args Args to pass when the hook is triggered.
+	 * @param int    $when Unix timestamp when the action will run.
+	 * @param string $group A group to put the action in.
+	 *
+	 * @return int The ID of the stored action.
+	 */
+	public function single_unique( $hook, $unique, $args = array(), $when = null, $group = '' ) {
 		$date = as_get_datetime_object( $when );
 		$schedule = new ActionScheduler_SimpleSchedule( $date );
 		$action = new ActionScheduler_Action( $hook, $args, $schedule, $group );
-		return $this->store( $action );
+		return $unique ? $this->store_unique_action( $action ) : $this->store( $action );
 	}
 
 	/**
@@ -175,5 +192,18 @@ class ActionScheduler_ActionFactory {
 	protected function store( ActionScheduler_Action $action ) {
 		$store = ActionScheduler_Store::instance();
 		return $store->save_action( $action );
+	}
+
+	/**
+	 * Store action if it's unique.
+	 *
+	 * @param ActionScheduler_Action $action Action object to store.
+	 *
+	 * @return int ID of the created action. Will be 0 if action was not created.
+	 */
+	protected function store_unique_action( ActionScheduler_Action $action ) {
+		$store = ActionScheduler_Store::instance();
+		return method_exists( $store, 'save_unique_action' ) ?
+			$store->save_unique_action( $action ) : $store->save_action( $action );
 	}
 }
--- a/classes/ActionScheduler_ActionFactory.php
+++ b/classes/ActionScheduler_ActionFactory.php
@@ -80,7 +80,7 @@ class ActionScheduler_ActionFactory {
 	 * @return int The ID of the stored action.
 	 */
 	public function single( $hook, $args = array(), $when = null, $group = '' ) {
-		return $this->single_with_unique( $hook, false, $args, $when, $group );
+		return $this->single_unique( $hook, false, $args, $when, $group );
 	}
 
 	/**
--- a/classes/ActionScheduler_ActionFactory.php
+++ b/classes/ActionScheduler_ActionFactory.php
@@ -66,20 +66,20 @@ class ActionScheduler_ActionFactory {
 	 * @return int The ID of the stored action.
 	 */
 	public function async( $hook, $args = array(), $group = '' ) {
-		return $this->async_unique( $hook, false, $args, $group );
+		return $this->async_unique( $hook, $args, $group, false );
 	}
 
 	/**
 	 * Same as async, but also supports $unique param.
 	 *
 	 * @param string $hook The hook to trigger when this action runs.
-	 * @param bool   $unique Whether to ensure the action is unique.
 	 * @param array  $args Args to pass when the hook is triggered.
 	 * @param string $group A group to put the action in.
+	 * @param bool   $unique Whether to ensure the action is unique.
 	 *
 	 * @return int The ID of the stored action.
 	 */
-	public function async_unique( $hook, $unique, $args = array(), $group = '' ) {
+	public function async_unique( $hook, $args = array(), $group = '', $unique = true ) {
 		$schedule = new ActionScheduler_NullSchedule();
 		$action   = new ActionScheduler_Action( $hook, $args, $schedule, $group );
 		return $unique ? $this->store_unique_action( $action, $unique ) : $this->store( $action );
@@ -96,21 +96,21 @@ class ActionScheduler_ActionFactory {
 	 * @return int The ID of the stored action.
 	 */
 	public function single( $hook, $args = array(), $when = null, $group = '' ) {
-		return $this->single_unique( $hook, false, $args, $when, $group );
+		return $this->single_unique( $hook, $args, $when, $group, false );
 	}
 
 	/**
 	 * Create single action only if there is no pending or running action with same name and params.
 	 *
 	 * @param string $hook The hook to trigger when this action runs.
-	 * @param bool   $unique Whether action scheduled should be unique.
 	 * @param array  $args Args to pass when the hook is triggered.
 	 * @param int    $when Unix timestamp when the action will run.
 	 * @param string $group A group to put the action in.
+	 * @param bool   $unique Whether action scheduled should be unique.
 	 *
 	 * @return int The ID of the stored action.
 	 */
-	public function single_unique( $hook, $unique, $args = array(), $when = null, $group = '' ) {
+	public function single_unique( $hook, $args = array(), $when = null, $group = '', $unique = true ) {
 		$date     = as_get_datetime_object( $when );
 		$schedule = new ActionScheduler_SimpleSchedule( $date );
 		$action   = new ActionScheduler_Action( $hook, $args, $schedule, $group );
@@ -129,22 +129,22 @@ class ActionScheduler_ActionFactory {
 	 * @return int The ID of the stored action.
 	 */
 	public function recurring( $hook, $args = array(), $first = null, $interval = null, $group = '' ) {
-		return $this->recurring_unique( $hook, false, $args, $first, $interval, $group );
+		return $this->recurring_unique( $hook, $args, $first, $interval, $group, false );
 	}
 
 	/**
 	 * Create the first instance of an action recurring on a given interval only if there is no pending or running action with same name and params.
 	 *
 	 * @param string $hook The hook to trigger when this action runs.
-	 * @param bool   $unique Whether action scheduled should be unique.
 	 * @param array  $args Args to pass when the hook is triggered.
 	 * @param int    $first Unix timestamp for the first run.
 	 * @param int    $interval Seconds between runs.
 	 * @param string $group A group to put the action in.
+	 * @param bool   $unique Whether action scheduled should be unique.
 	 *
 	 * @return int The ID of the stored action.
 	 */
-	public function recurring_unique( $hook, $unique, $args = array(), $first = null, $interval = null, $group = '' ) {
+	public function recurring_unique( $hook, $args = array(), $first = null, $interval = null, $group = '', $unique = true ) {
 		if ( empty( $interval ) ) {
 			return $this->single_unique( $hook, $unique, $args, $first, $group );
 		}
@@ -168,24 +168,25 @@ class ActionScheduler_ActionFactory {
 	 * @return int The ID of the stored action.
 	 */
 	public function cron( $hook, $args = array(), $base_timestamp = null, $schedule = null, $group = '' ) {
-		return $this->cron_unique( $hook, false, $args, $base_timestamp, $schedule, $group );
+		return $this->cron_unique( $hook, $args, $base_timestamp, $schedule, $group, false );
 	}
+
 
 	/**
 	 * Create the first instance of an action recurring on a Cron schedule only if there is no pending or running action with same name and params.
 	 *
 	 * @param string $hook The hook to trigger when this action runs.
-	 * @param bool   $unique Whether action scheduled should be unique.
 	 * @param array  $args Args to pass when the hook is triggered.
 	 * @param int    $base_timestamp The first instance of the action will be scheduled
 	 *        to run at a time calculated after this timestamp matching the cron
 	 *        expression. This can be used to delay the first instance of the action.
 	 * @param int    $schedule A cron definition string.
 	 * @param string $group A group to put the action in.
+	 * @param bool   $unique Whether action scheduled should be unique.
 	 *
 	 * @return int The ID of the stored action.
 	 **/
-	public function cron_unique( $hook, $unique, $args = array(), $base_timestamp = null, $schedule = null, $group = '' ) {
+	public function cron_unique( $hook, $args = array(), $base_timestamp = null, $schedule = null, $group = '', $unique = true ) {
 		if ( empty( $schedule ) ) {
 			return $this->single_unique( $hook, $unique, $args, $base_timestamp, $group );
 		}
--- a/classes/ActionScheduler_ActionFactory.php
+++ b/classes/ActionScheduler_ActionFactory.php
@@ -6,27 +6,29 @@
 class ActionScheduler_ActionFactory {
 
 	/**
-	 * @param string $status The action's status in the data store
-	 * @param string $hook The hook to trigger when this action runs
-	 * @param array $args Args to pass to callbacks when the hook is triggered
-	 * @param ActionScheduler_Schedule $schedule The action's schedule
-	 * @param string $group A group to put the action in
+	 * Return stored actions for given params.
 	 *
-	 * @return ActionScheduler_Action An instance of the stored action
+	 * @param string                   $status The action's status in the data store.
+	 * @param string                   $hook The hook to trigger when this action runs.
+	 * @param array                    $args Args to pass to callbacks when the hook is triggered.
+	 * @param ActionScheduler_Schedule $schedule The action's schedule.
+	 * @param string                   $group A group to put the action in.
+	 *
+	 * @return ActionScheduler_Action An instance of the stored action.
 	 */
 	public function get_stored_action( $status, $hook, array $args = array(), ActionScheduler_Schedule $schedule = null, $group = '' ) {
 
 		switch ( $status ) {
-			case ActionScheduler_Store::STATUS_PENDING :
+			case ActionScheduler_Store::STATUS_PENDING:
 				$action_class = 'ActionScheduler_Action';
 				break;
-			case ActionScheduler_Store::STATUS_CANCELED :
+			case ActionScheduler_Store::STATUS_CANCELED:
 				$action_class = 'ActionScheduler_CanceledAction';
 				if ( ! is_null( $schedule ) && ! is_a( $schedule, 'ActionScheduler_CanceledSchedule' ) && ! is_a( $schedule, 'ActionScheduler_NullSchedule' ) ) {
 					$schedule = new ActionScheduler_CanceledSchedule( $schedule->get_date() );
 				}
 				break;
-			default :
+			default:
 				$action_class = 'ActionScheduler_FinishedAction';
 				break;
 		}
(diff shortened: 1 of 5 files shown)


vedanshujain:
Note that only `as_schedule_single_action` is currently hooked with the unique query, but if approach is fine I can hook following methods as well:
- [x] as_enqueue_async_action( $hook, $args = array(), $group = '', $unique = false )
- [x] as_schedule_cron_action( $timestamp, $schedule, $hook, $args = array(), $group = '', $unique = false )
- [x] as_schedule_recurring_action( $timestamp, $interval_in_seconds, $hook, $args = array(), $group = '', $unique = false )
- [x] as_schedule_single_action( $timestamp, $hook, $args = array(), $group = '', $unique = false )

To test, open-shell and run following commands:
```
>>> as_schedule_single_action( time(), 'test_hook', array(), 'test_group', true); // scheduled since this is unique.
=> 121870 

>>> as_schedule_single_action( time(), 'test_hook', array(), 'test_group', true); // not scheduled since this is duplicate.
=> 0
```

classes/data-stores/ActionScheduler_DBStore.php
```diff
@@ -87,6 +119,107 @@ public function save_action( ActionScheduler_Action $action, \DateTime $date = n
...
+			ActionScheduler_Store::STATUS_RUNNING,
+		);
+		$pending_status_placeholders = implode( ', ', array_fill( 0, count( $pending_statuses ), '%s' ) );
+		// phpcs:disable WordPress.DB.PreparedSQL.NotPrepared, WordPress.DB.PreparedSQL.InterpolatedNotPrepared -- $pending_status_placeholders is hardcoded.
+		$where_clause = $wpdb->prepare(
+			"
+SELECT action_id FROM $table_name
+WHERE status IN ( $pending_status_placeholders )
+AND hook = %s
+AND `group_id` = %s
```
```suggestion
AND `group_id` = %d
```

AFAIK the `group_id` is always an integer.

tests/phpunit/jobstore/ActionScheduler_DBStore_Test.php
```diff
@@ -446,4 +446,47 @@ public function test_get_run_date() {
...
+		$action   = new ActionScheduler_Action( $hook, array(), $schedule );
+
+		$action_id = $store->save_action( $action );
+		$this->assertNotEquals( 0, $action_id );
+		$action_from_db = $store->fetch_action( $action_id );
+		$this->assertTrue( is_a( $action_from_db, ActionScheduler_Action::class ) );
+
+		$action_id_duplicate = $store->save_unique_action( $action );
+		$this->assertEquals( 0, $action_id_duplicate );
+	}
```
barryhughes, line 467: A thought: it might make the design clearer if we add another test that is almost the inverse of this one. That is:

- Save a unique action **first** (should be successful)
- Then save a non-unique action using the same hook second (should also successful—but this could surprise some users of the library)

This would illustrate that that action itself is not somehow unique—in its persisted form, it's really just another ordinary action—but that uniqueness is only achieved through consistent use of the new `$unique` property when setting up actions.

wdyt?

classes/data-stores/ActionScheduler_DBStore.php
```diff
@@ -87,6 +119,107 @@ public function save_action( ActionScheduler_Action $action, \DateTime $date = n
...
+		$table_name = ! empty( $wpdb->actionscheduler_actions ) ? $wpdb->actionscheduler_actions : $wpdb->prefix . 'actionscheduler_actions';
+
+		$column_sql                  = '`' . implode( '`, `', $columns ) . '`';
+		$placeholder_sql             = implode( ', ', $placeholders );
+		$where_clause = $this->build_where_clause_for_insert( $data, $table_name, $unique );
+		// phpcs:disable WordPress.DB.PreparedSQL.NotPrepared, WordPress.DB.PreparedSQL.InterpolatedNotPrepared -- $column_sql and $where_clause are already prepared. $placeholder_sql is hardcoded.
+		$insert_query                = $wpdb->prepare(
+			"
+INSERT INTO $table_name ( $column_sql )
+SELECT $placeholder_sql FROM DUAL
```
barryhughes, line 145: Mostly asking out of ignorance—what does `FROM DUAL` give us (I realize `DUAL` is a 'dummy table', but the query seems to work without it; does it aid performance)?

barryhughes: Further on we also do `SELECT NULL FROM DUAL` instead of just `SELECT NULL` ... I'm guessing the same answer applies in both places?

classes/data-stores/ActionScheduler_DBStore.php
```diff
@@ -87,6 +119,107 @@ public function save_action( ActionScheduler_Action $action, \DateTime $date = n
...
+	private function build_where_clause_for_insert( $data, $table_name, $unique ) {
+		global $wpdb;
+
+		if ( ! $unique ) {
+			return 'SELECT NULL FROM DUAL';
+		}
+
+		$pending_statuses            = array(
+			ActionScheduler_Store::STATUS_PENDING,
+			ActionScheduler_Store::STATUS_RUNNING,
```
barryhughes, line 172: There's a pattern we sometimes see in which each scheduled action callback will assess if more work is needed and, if so, schedules its own replacement (sometimes used for migration-type tasks instead of an actual recurring action):

```php
add_action( 'do_migration', function () {
	do_migration_step();

	if ( migration_work_still_needed() ) {
		as_schedule_single_action( time() + HOUR_IN_SECONDS, 'do_migration' );
	}
} );

as_schedule_single_action( time(), 'do_migration' );
```

If this was updated to use unique actions, I think it would fail (because the currently running action will have a status of `in-progress` and that's enough to block new unique actions from being scheduled). I don't know if it makes sense to change this, and it won't cause a problem for existing code, but I wanted to flag it—I'm a little unsure how obvious this will be for library users (though we can document it, both inline and [over here](https://actionscheduler.org)).

barryhughes (commented):
Do we need to implement this in the `wpPostStore`? 

So far we've maintained parity between both stores, but if we want to move away from that we should probably clarify in the inline docs that `$unique` only works with the `DBStore` (and/or consider a dev advisory)—wdyt?

classes/data-stores/ActionScheduler_DBStore.php
vedanshujain: So `DUAL` provides backward compatibility with MySQL 5.5. After that, it wasn't needed to specify `DUAL` (MySQL will assume it), but since this is a pretty low effort to get some back-compat I added it.

classes/data-stores/ActionScheduler_DBStore.php
vedanshujain: Hmm, I suppose unique should only be used when scheduling the first action, for example from the update hook. We can document this, but I wonder if we should have another API such as `repeat` as part of a separate PR which will make it much more obvious.

tests/phpunit/jobstore/ActionScheduler_DBStore_Test.php
vedanshujain: Agreed, added in https://github.com/woocommerce/action-scheduler/pull/831/commits/b8abba6ff11316a6ab8e622b33fcdc65171ae098

vedanshujain (commented):
> Do we need to implement this in the wpPostStore?

What are our plans for wpPostStore? Given that we migrate actions to DBStore I assume we would like to deprecate it? In this case, we case we can go ahead and ignore it (with a devAdvisory of course) and maybe a doing_it_wrong warning as well.

vedanshujain:
Note that tests are failing because unrelated network issue, i will try again after some time.

classes/data-stores/ActionScheduler_DBStore.php
barryhughes: TIL, thank you.

barryhughes:
> What are our plans for wpPostStore?

My assumptions match your own, but I don't think this was solidified (or communicated to the community) and there are still cases where the post store is used that we should be aware of:

- Indirectly, via the HybridStore. Example:
    - We monitor [plugin deactivations](https://github.com/woocommerce/action-scheduler/blob/3.4.2/classes/ActionScheduler_DataController.php#L169) and this resets the migration status.
    - This leads to the [HybridStore](https://github.com/woocommerce/action-scheduler/blob/3.4.2/classes/migration/Controller.php#L59-L61) temporarily becoming the default.
    - The effect is that the post store becomes [read only](https://github.com/woocommerce/action-scheduler/blob/3.4.2/classes/data-stores/ActionScheduler_HybridStore.php#L45-L46) and the DB store becomes [read/write](https://github.com/woocommerce/action-scheduler/blob/3.4.2/classes/data-stores/ActionScheduler_HybridStore.php#L45-L46)
- It's also possible for 3PDs to force the use of the post store:
    - This can be done explicitly via filter hook [`action_scheduler_store_class`](https://github.com/woocommerce/action-scheduler/blob/3.4.2/classes/abstracts/ActionScheduler_Store.php#L417).
    - It can be achieved by indicating the migraton to DB Store is not possible. [Here is a non-theoretical example](https://plugins.trac.wordpress.org/browser/ip-locator/tags/3.5.0/ip-locator.php#L41) of that happening via a third party plugin.

> We case we can go ahead and ignore it (with a devAdvisory of course) and maybe a doing_it_wrong warning as well.

I agree. Despite the above, by default the only condition that results in the post store coming back to life is a temporary one following a plugin deactivation and, when that happens, we don't actually write anything to the post store. So it seems safe and reasonable to me to ignore and/or emit a *doing-it-wrong* message (+thanks for talking this out!).

classes/data-stores/ActionScheduler_DBStore.php
barryhughes: > I suppose unique should only be used when scheduling the first action

I could go either way on this, but on reflection I think the above is what I tend to prefer (vs adding a repeat API) and documenting this as one of our preferred patterns is probably sufficient.

Though if you strongly prefer to add new functions (or make further changes to the signature of our existing ones), that's probably reasonable, too :+1:

barryhughes:
Just to avoid confusion or delay: my reading is you will update this PR by expanding the change to [cover these other functions](https://github.com/woocommerce/action-scheduler/pull/831#issuecomment-1160454275) (so I didn't mark this as approved). If I'm wrong, and you're handling those other functions via distinct PRs, that's fine too—and this can be merged.

vedanshujain:
> my reading is you will update this PR by expanding the change to https://github.com/woocommerce/action-scheduler/pull/831#issuecomment-1160454275

You are right, I am working on updating these existing functions as well, and incorporate additional feedback as well (basically to find ways to communicate to user that post store is not going to be unique)

classes/data-stores/ActionScheduler_DBStore.php
vedanshujain: I agree, let's update the usage of API after this lands.

vedanshujain:
Looks like tests are failing for PHP 5.6, taking a look.

Edit: Done now, ready for another review.

classes/ActionScheduler_ActionFactory.php
```diff
@@ -57,76 +59,141 @@ public function get_stored_action( $status, $hook, array $args = array(), Action
...
+	 * @param array  $args Args to pass when the hook is triggered.
+	 * @param int    $base_timestamp The first instance of the action will be scheduled
+	 *        to run at a time calculated after this timestamp matching the cron
+	 *        expression. This can be used to delay the first instance of the action.
+	 * @param int    $schedule A cron definition string.
+	 * @param string $group A group to put the action in.
+	 *
+	 * @return int The ID of the stored action.
+	 **/
+	public function cron_unique( $hook, $unique, $args = array(), $base_timestamp = null, $schedule = null, $group = '' ) {
```
barryhughes: Optional thing, just wanted to drop here for consideration (and applies to all of the new `*_unique()` methods):

Bearing in mind this method is public, if I was calling it directly I'd probably assume that (outside of special cases) I don't need to specify `$unique` is `true` (and the method description implies the same thing). With those things in mind, would:

```php
public function cron_unique( $hook, $args = array(), $base_timestamp = null, $schedule = null, $group = '', $unique = true )
```

Be a better signature (that is, make `$unique` optional and have it default to true)? Of course, changing this would also require updates to the matching non-unique methods.

barryhughes (approved):
This is looking good to me! Provisionally approving (I did leave a further suggestion, but it's optional). 

The only other thing I want to pause and ask is, are we happy with the default level of uniqueness? Right now, unique means same hook and same args; but is this going to be useful to all users of the library/could there be a valid case to supporting a lesser level of uniqueness that only matches the hook and not the args?

In asking this I'm just being cautious, because once we cement the new public API signatures with the new bool arg `$unique` it may be hard to gracefully address the above in the future.

If we're content with this approach (and though I'm asking this question I also think what we have is entirely reasonable), though, we can merge—nice work on this :+1:

barryhughes:
Relating to WooCommerce, at some point we should think about implications for `WC_Queue_Interface`. Though, that's a discussion we don't need to progress in this repo 🙃

classes/ActionScheduler_ActionFactory.php
vedanshujain: Hmm you are right, let me take care of this.

vedanshujain:
> re we happy with the default level of uniqueness? Right now, unique means same hook and same args; but is this going to be useful to all users of the library/could there be a valid case to supporting a lesser level of uniqueness that only matches the hook and not the args?

I think I can go either way, but I can remove the clause that compares for args if your intuition is going that way. I looked at usages for AS library in the wild and seems like both ways are quite popular. In https://github.com/woocommerce/action-scheduler/pull/831/commits/4ef26e1374be946b4f4e7e26ee47070913a79924 I modified the behavior a bit so that when scheduling action with empty args, it won't use args for comparing uniqueness. This method also has it's pros and cons, but now this is functionally same as `as_has_scheduled_action`, and given that we haven't faced problems with behavior of `as_has_scheduled_action`, this should now be fine too. wdyt?

barryhughes:
> I modified the behavior a bit so that when scheduling action with empty args, it won't use args for comparing uniqueness. This method also has it's pros and cons, but now this is functionally same as `as_has_scheduled_action`, and given that we haven't faced problems with behavior of `as_has_scheduled_action`, this should now be fine too. wdyt?

I think `as_has_scheduled_action()` is a good reference point, but a distinction is that it doesn't match arguments if `$args === null` (and *does* match if `$args === []`, if that is what the caller explicitly provides). That seems like a good strategy, because it supports both behaviors (zero matching of arguments, and exact matching of arguments including no arguments).

classes/data-stores/ActionScheduler_DBStore.php
```diff
@@ -37,21 +37,47 @@ public function init() {
...
 	 * @param ActionScheduler_Action $action Action object.
-	 * @param DateTime              $date Optional schedule date. Default null.
+	 * @param \DateTime              $scheduled_date Optional schedule date. Default null.
 	 *
 	 * @return int Action ID.
 	 * @throws RuntimeException     Throws exception when saving the action fails.
 	 */
-	public function save_action( ActionScheduler_Action $action, \DateTime $date = null ) {
-		try {
+	public function save_action( ActionScheduler_Action $action, \DateTime $scheduled_date = null ) {
```
barryhughes, line 61: I don't think any action is needed here—it makes sense because the second param is also defined as `DateTime $scheduled_date = NULL` in the base class, so this brings the DBStore into alignment with that—but I wonder if it could cause issues if a user was interacting with the concrete object using named params in their method calls?

Probably safe to ignore, though (because it should align with the interface provided by the abstract class, and because we encourage use of top-level functions and not methods like this one).

classes/data-stores/ActionScheduler_DBStore.php
```diff
@@ -37,21 +37,47 @@ public function init() {
...
-	 * Save an action.
+	 * Save an action, checks if this is a unique action before actually saving.
+	 *
+	 * @param ActionScheduler_Action $action         Action object.
+	 * @param \DateTime              $scheduled_date Optional schedule date. Default null.
+	 *
+	 * @return int                  Action ID.
+	 * @throws RuntimeException     Throws exception when saving the action fails.
+	 */
+	public function save_unique_action( ActionScheduler_Action $action, \DateTime $scheduled_date = null ) {
```
barryhughes, line 48: Sorry for not asking this earlier ... but should this be stubbed out in the [abstract class](https://github.com/woocommerce/action-scheduler/blob/3.4.2/classes/abstracts/ActionScheduler_Store.php) as a concrete method that is either no-op, or that throws a doing-it-wrong ('missing implementation error'), or that provides a universal but non-atomic approximation of this operation?

vedanshujain: Actually adding a stub in the base class will cause issues with the check defined here: https://github.com/woocommerce/action-scheduler/pull/831/files#diff-ba3b93f44933cf6c97c53e92cdc95e3b0e492de26ce0c87430f2d3e832bc95c1R259 which will then evaluate to true for PostStore (and any other custom store). Maybe we can an API which the custom store can use to declare unique support (although in a different PR) and keep this one as it is? wdyt?

classes/data-stores/ActionScheduler_DBStore.php
vedanshujain: Yeah, I think we should be able to ignore this since having consistency with the base class is probably better.

classes/ActionScheduler_ActionFactory.php
vedanshujain: Note that this was addressed in https://github.com/woocommerce/action-scheduler/pull/831/commits/c8652fe2344ebe2d264f64f7f8ad9874b33161a5

vedanshujain:
> a distinction is that it doesn't match arguments if $args === null (and does match if $args === [], if that is what the caller explicitly provides)

Just to confirm, this is what the current behavior should be doing, unless I am not reading this correctly. Note that the constrcutor in `ActionScheduler_Action` does not allow `null` in the `args` array (for egs `as_schedule_single_action( time(), 'action_hook_testing', null, 'group' );` will fail, so probably exact matching is not possible without doing more refactoring.

classes/data-stores/ActionScheduler_DBStore.php
barryhughes: I saw that check ... then forgot it was there. I think let's leave it as is;  a more formal system for declaring support seems like overkill (to me).

barryhughes:
I think we're talking about the same thing—but, to be extra clear, this is the handling I'm referring to:

```php
as_schedule_single_action( $time, $hook, $args );   // Create an action with non-empty args
var_dump( as_has_scheduled_action( $hook, null ) ); // TRUE  (matches *any* args)
var_dump( as_has_scheduled_action( $hook, [] ) );   // FALSE (matches *empty* args)
```

> so probably exact matching is not possible without doing more refactoring

Yes: to align with `as_has_scheduled_action()`, we'd need to make changes like the following (and propagate along the line as needed):

```php
as_schedule_single_action( $timestamp, $hook, $args = array(), $group = '', $unique = false )    // ← FROM THIS
as_schedule_single_action( $timestamp, $hook, array $args = null, $group = '', $unique = false ) // ← TO THIS
```

I think this can be done in a backwards compatible way: the above change (to top-level function signatures) feels safe, and the matching changes within `ActionScheduler_ActionFactory` could be limited to the newly introduced methods. 

... We don't **need** to do this, we can roll with the current implementation (but adding this later will be harder—the usual trade-off). In a way it depends how much we think the final type of uniqueness in the following table might matter to users of the library:

| Uniqueness | Current PR | Proposed |
| --- | --- | --- |
| Can match the same hook with any args | **YES** | **YES** |
| Can match the same hook with a specific set of non-empty args | **YES** | **YES** |
| Can match the same hook with an empty argument set | **NO** | **YES** |

(And though I'm talking this out ... I realize developer bandwidth is constrained.)

vedanshujain:
@barryhughes based on our discussion I have removed the args matching completely in https://github.com/woocommerce/action-scheduler/pull/831/commits/7b24ea6d4d20bfae09c523c784f15a6d7db73441. I do see that it's easier to use the API now, let's the add param matching later. Let me know what you think?

classes/data-stores/ActionScheduler_DBStore.php
```diff
@@ -68,13 +95,18 @@ public function save_action( ActionScheduler_Action $action, \DateTime $date = n
...
-			$table_name = ! empty( $wpdb->actionscheduler_actions ) ? $wpdb->actionscheduler_actions : $wpdb->prefix . 'actionscheduler_actions';
-			$wpdb->insert( $table_name, $data );
+			$insert_sql = $this->build_insert_sql( $data, $unique );
+
+			// phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared -- $insert_sql should be already prepared.
+			$wpdb->query( $insert_sql );
 			$action_id = $wpdb->insert_id;
 
 			if ( is_wp_error( $action_id ) ) {
 				throw new \RuntimeException( $action_id->get_error_message() );
```
barryhughes, line 105: Sidenote: it would seem `$action_id` will only ever be an int, or possibly a numerical string in some cases, and can never be an instance of `WP_Error`.

(Of course, though I'm noting it, it is also not related to your change.)

barryhughes (approved):
LGTM!

Tests are failing for an unrelated reason, which I think should be solved [via this PR](https://github.com/woocommerce/action-scheduler/pull/835). As soon as that's merged I'll try to circle back, rebase and merge :+1:

PR #846, opened 2022-09-04
fix: PHP 8.2 deprecated notice

JosephGabito:
Specific error details:

Deprecated: Using ${var} in strings is deprecated, use {$var} instead in /usr/local/var/www/wordpress/wp-content/plugins/action-scheduler/classes/schema/ActionScheduler_LoggerSchema.php on line 40

https://php.watch/versions/8.2/$%7Bvar%7D-string-interpolation-deprecated#dollar-outside

### Changelog

> Fix - Update variable interpolation syntax for improved compatibility with PHP 8.2 onwards.


barryhughes (approved):
LGTM, thanks for tidying that up :+1:

Issue #848, opened 2022-09-06
Find out if is last task in queue using `as_next_scheduled_action`


mischnu:
Hi there,

I have a bunch of one-time tasks tied to a callback. Within my callback function I want to know if this is the last run for this kind of task. How would I do this?

I tried using `as_next_scheduled_action( 'callback_name' )`, but during my testing this always returned `1` in any case if called within the in-progress callback. Is this meant to be like that? Couldn‘t find anything about this in the docs.

barryhughes:
Hi @mischnu,

> I tried using as_next_scheduled_action( 'callback_name' ), but during my testing this always returned 1 in any case if called within the in-progress callback. Is this meant to be like that?

Sort of ... it should behave like this:

- If the next instance is in progress, it returns `(bool) true`.
- If the next instance is pending, it returns the timestamp as an integer.
- If there is no next instance, it returns `(bool) false`.

Possibly you are getting `(string) '1'` because you are implicitly casting the result (which would be `(bool) true` in this scenario) to a string. In any case, I'm not sure this function is ideal for your use case.

> Within my callback function I want to know if this is the last run for this kind of task. How would I do this?

Would something like the following work for you?

```php
// Within the callback for YOUR_HOOK:
$future_actions = ActionScheduler::store()->query_action( [
    'hook'   => 'YOUR_HOOK',
    'status' => ActionScheduler_Store::STATUS_PENDING,
] );

if ( ! $future_actions ) {
    // This is the last currently scheduled instance of YOUR_HOOK.
}
```

mischnu:
Hi @barryhughes,

thank you so much, your code does exactly what I need!

I would suggest to add this case:

> If the next instance is in progress, it returns `(bool) true`.

… to the [API function reference](https://actionscheduler.org/api/#function-reference--as_next_scheduled_action), as this one is missing there somehow.

Once again, thanks!

barryhughes:
> to the [API function reference](https://actionscheduler.org/api/#function-reference--as_next_scheduled_action), as this one is missing there somehow.

Thanks for flagging that: you're right, and it also appears it's being rendered a little weirdly (seems the return type is being transformed into a table, for some reason). I'll make a note to fix that up :+1:

mischnu:
Just an idea: How about adding a fourth parameter (boolean) to exclude in-progress actions from the return value? Would be an easy way for the scenario described above. Default value could be `FALSE` to have backward compatibility.

[...]

PR #849, opened 2022-09-07
Maintenance on A/S docs

barryhughes:
Updates the library docs. Partly triggered by [this observation](https://github.com/woocommerce/action-scheduler/issues/848#issuecomment-1238987973), though I then noticed a few things needing adjustment.

### What I changed and why

- **as_next_scheduled_action() return type** → As highlighted in the issue linked above, the description missed a key detail about in-progress and async actions. I have updated it to match the text in the source docblock.
- **All return types** → These have been wrapped in backticks, so they render as inline code. That's because, without backticks, any union types written as `(integer|boolean)` are rendered by Jekyll as simple tables...which looks a little broken.
- **Added optional unique param** → An optional `$unique` param is available for various functions but we forgot to update these docs.
- **Improved usage example** → I updated a code snippet to take advantage of the `$unique` param, but in a backwards compatible way.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -194,7 +194,7 @@ as_next_scheduled_action( $hook, $args, $group );
 
 ### Return value
 
-`(integer|boolean)` The timestamp for the next occurrence, or false if nothing was found.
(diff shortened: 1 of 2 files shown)


vedanshujain (approved):
LGTM!

PR #855, opened 2022-09-15
Error saving action: `ActionScheduler_Action::$args` too long

### LINKED ISSUE #851: PHP Fatal error: Uncaught RuntimeException: Error saving action: `ActionScheduler_Action::$args` too long
Hey,

A few of our plugin users (Pro version of the [Automator plugin](https://github.com/UncannyOwl/Uncanny-Automator)) have reported this issue. We are using v3.4.0 of the Action Scheduler library in our plugin. Has anyone else reported the problem?
```
"require": {
    "woocommerce/action-scheduler": "^3.4.0"
  },
```

Can this hardcoded **extended_args** `varchar(8000)` limit be changed to `longext` instead?

https://github.com/woocommerce/action-scheduler/blob/3fe1e79f031c7300e5c60c61f732b358552ddc16/classes/schema/ActionScheduler_StoreSchema.php#L59

```
[02-Sep-2022 08:54:31 UTC] PHP Fatal error: Uncaught RuntimeException: Error saving action: ActionScheduler_Action::$args too long. To ensure the args column can be indexed, action args should not be more than 8000 characters when encoded as JSON. in /wp-content/plugins/xxxxxx/vendor/woocommerce/action-scheduler/classes/data-stores/ActionScheduler_DBStore.php:86

Stack trace:

#0 /wp-content/plugins/xxxxxx/vendor/woocommerce/action-scheduler/classes/ActionScheduler_ActionFactory.php(177): ActionScheduler_DBStore->save_action(Object(ActionScheduler_Action))

#1 /wp-content/plugins/xxxxxx/vendor/woocommerce/action-scheduler/classes/ActionScheduler_ActionFactory.php(84): ActionScheduler_ActionFactory->store(Object(ActionScheduler_Action))

#2 /wp-content/plugins/xxxxxx/vendor/woocommerce/action-scheduler/functions.php(36): ActionScheduler_ in /wp-content/plugins/xxxxxx/vendor/woocommerce/action-scheduler/classes/data-stores/ActionScheduler_DBStore.php on line 86
```

saad-siddique:
Closes #851

PHP Fatal error: Uncaught RuntimeException: Error saving action: ActionScheduler_Action::$args too long

### DIFF
--- a/classes/data-stores/ActionScheduler_DBStore.php
+++ b/classes/data-stores/ActionScheduler_DBStore.php
@@ -19,8 +19,13 @@ class ActionScheduler_DBStore extends ActionScheduler_Store {
 	 */
 	private $claim_before_date = null;
 
-	/** @var int */
-	protected static $max_args_length = 4294967295; // longtext length (4GB)
+	/**
+	 * The "longtext" field type can store up to 4,294,967,295 (4GB) characters.
+	 * For sanity, allow up to half the length of 4GB (2GB) in the field.
+	 *
+	 * @var int
+	 */
+	protected static $max_args_length = 2147483647; // longtext length (4GB)
 
 	/** @var int */
 	protected static $max_index_length = 191;
--- a/classes/data-stores/ActionScheduler_DBStore.php
+++ b/classes/data-stores/ActionScheduler_DBStore.php
@@ -20,7 +20,7 @@ class ActionScheduler_DBStore extends ActionScheduler_Store {
 	private $claim_before_date = null;
 
 	/** @var int */
-	protected static $max_args_length = 8000;
+	protected static $max_args_length = 4294967295; // longtext length (4GB)
 
 	/** @var int */
 	protected static $max_index_length = 191;
--- a/classes/schema/ActionScheduler_StoreSchema.php
+++ b/classes/schema/ActionScheduler_StoreSchema.php
@@ -56,7 +56,7 @@ class ActionScheduler_StoreSchema extends ActionScheduler_Abstract_Schema {
 				        last_attempt_gmt datetime NULL default '${default_date}',
 				        last_attempt_local datetime NULL default '${default_date}',
 				        claim_id bigint(20) unsigned NOT NULL default '0',
-				        extended_args varchar(8000) DEFAULT NULL,
+				        extended_args longtext,
 				        PRIMARY KEY  (action_id),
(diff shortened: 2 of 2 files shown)


barryhughes (changes requested):
Thanks for putting this together, @saad-siddique.

* In a related conversation, there was [discussion of using a hash](https://github.com/woocommerce/action-scheduler/issues/418#issuecomment-583176940) to keep performance snappy: I think there is a lot of merit in adding that (but let me know if you considered this and found it was unnecessary or unhelpful).
* Even if we move ahead without hashing, we would need various other changes. For example, this change would represent a new schema version and the change to existing method `update_schema_5_0` won't be "seen" by installations where the schema is already up-to-date.

saad-siddique:
Hi @barryhughes ,

Thank you for taking the time to look at the issue.

Thank you for suggesting the `hash` method. It's possible to use that method, but it seems like a band-aid solution. The solution does work, however. 

My main reason not to use the hash method is that instead of storing the Action Scheduler's related `$args` into its table, it's forcing us to store it in the postmeta or options table. It will pollute the tables meant for posts or options/settings. Another reason is the cleanup of past/failed/successful scheduled actions isn't going to drop related records from these tables in case the action failed to complete (like the action threw a fatal error during the process). At least deleting a record from AS tables will remove that data too.

Based on your comment above, I've updated the schema to `7` and added all the necessary code. The `$max_args_length` validation still exists but increased to store approx. half the length of the `longtext` field, which I believe should be sufficient enough for several cases, including at scale sites. 

Please let me know if anything else needs to be updated.

barryhughes:
> My main reason not to use the hash method is that instead of storing the Action Scheduler's related $args into its table, it's forcing us to store it in the postmeta or options table.

Yep, I agree that wouldn't be ideal. 

Though, my reading of the conversation *("you could keep args as json encoded/serialized in a new meta field")* was not that we would necessarily depend on core WordPress tables to store the full length args (which, of course, is currently possible even without this change).

In any case, I wanted to let you know we haven't lost sight of this change—but there are a few other items we're prioritizing right now, so there might be a further delay before we can really tackle this one.

xdividr:
Following up to add 2¢ that I for one would appreciate seeing a robust solution for longer stored values to improve performance and stability relative to the hash method mentioned above. Thanks to all.

Issue #857, opened 2022-09-16
Disable with a constant


Mte90:
So to achieve that you need to use a filter but in the automated tests world this means that you need a WP instance already available.

There is https://github.com/woocommerce/action-scheduler-disable-default-runner that use that filter and let to disable it but it isn't more simple to have a constant that you can disable on the wp-config.php?

barryhughes:
Hi @Mte90,

I could certainly see the value in that. I'll mark this as a low priority enhancement and, if you wish, you could submit a PR to add this and we'll be happy to review.

Mte90:
So is not a difficult PR https://github.com/woocommerce/action-scheduler/search?q=action_scheduler_run_queue but what is the name that you prefer for the constant? So I do everything once :-)

barryhughes:
I don't believe we currently define or look for any top-level constants (except for some that are defined by WordPress or WP CLI), so there isn't really any precedent to look at, but I would propose:

`AS_DISABLE_QUEUE_RUNNERS`

What do you think?

Mte90:
PR done :-)

barryhughes:
We haven't had any additional requests for this, and the desired outcome of disabling Action Scheduler is already possible via various hooks (this would just make it easier). Given those things and because the linked PR has stalled and closed, we'll close this issue.

PR #862, opened 2022-09-21
feat(disable): constant to disable it

Mte90:
Ref: https://github.com/woocommerce/action-scheduler/issues/857

I am not sure if it is enough to not initialize the class (check for tables etc) but for sure disable the cron and the queue.

### DIFF
--- a/classes/ActionScheduler_AsyncRequest_QueueRunner.php
+++ b/classes/ActionScheduler_AsyncRequest_QueueRunner.php
@@ -49,7 +49,7 @@ class ActionScheduler_AsyncRequest_QueueRunner extends WP_Async_Request {
 	 * if there are still pending actions after completing a queue in this request.
 	 */
 	protected function handle() {
-		if ( !defined( 'AS_DISABLE_QUEUE_RUNNERS' ) || !AS_DISABLE_QUEUE_RUNNERS ) {
+		if ( ! defined( 'AS_DISABLE_QUEUE_RUNNERS' ) || ! AS_DISABLE_QUEUE_RUNNERS ) {
 			do_action( 'action_scheduler_run_queue', 'Async Request' ); // run a queue in the same way as WP Cron, but declare the Async Request context
 		}
 
--- a/classes/ActionScheduler_AsyncRequest_QueueRunner.php
+++ b/classes/ActionScheduler_AsyncRequest_QueueRunner.php
@@ -49,7 +49,9 @@ class ActionScheduler_AsyncRequest_QueueRunner extends WP_Async_Request {
 	 * if there are still pending actions after completing a queue in this request.
 	 */
 	protected function handle() {
-		do_action( 'action_scheduler_run_queue', 'Async Request' ); // run a queue in the same way as WP Cron, but declare the Async Request context
+		if ( !defined( 'AS_DISABLE_QUEUE_RUNNERS' ) || !AS_DISABLE_QUEUE_RUNNERS ) {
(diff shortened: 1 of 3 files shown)


classes/ActionScheduler_AsyncRequest_QueueRunner.php
```diff
@@ -49,7 +49,9 @@ public function __construct( ActionScheduler_Store $store ) {
 	 * if there are still pending actions after completing a queue in this request.
 	 */
 	protected function handle() {
-		do_action( 'action_scheduler_run_queue', 'Async Request' ); // run a queue in the same way as WP Cron, but declare the Async Request context
+		if ( !defined( 'AS_DISABLE_QUEUE_RUNNERS' ) || !AS_DISABLE_QUEUE_RUNNERS ) {
```
barryhughes:
```suggestion
		if ( ! defined( 'AS_DISABLE_QUEUE_RUNNERS' ) || ! AS_DISABLE_QUEUE_RUNNERS ) {
```

docs/faq.md
```diff
@@ -6,6 +6,14 @@ Action Scheduler is designed to be used and released in plugins. It avoids redec
 
 To use it in your plugin (or theme), simply require the `action-scheduler/action-scheduler.php` file. Action Scheduler will take care of the rest. __Note:__ Action Scheduler is only loaded from a theme if it is not included in any active plugins.
 
+### How to disable Action Scheduler?
+
+Disabled it is helpful for automated testing and it is enough to add a constant in your `wp-config.php` as:
+
+```
+define( 'AS_DISABLE_QUEUE_RUNNERS', true );
+```
```
barryhughes, line 15: ✍🏽 Once we're happy with the change (and can be confident about the version it will become available) we could consider adding a further note here, such as:

```md
This is available from Action Scheduler x.y.z onwards.
```

barryhughes (changes requested):
Thanks! This looks good, though there are some further aspects I think we should consider.

### Usage

Once we make this available it will become a general purpose tool used in various scenarios (not just automated testing). For example, a site owner may hire a developer to work on their site in some way, and the developer may decide to use the constant during development or troubleshooting: flagging this as it is the background for the following scenarios.

### Messaging

With the constant in place and set to `true`, we will still see admin notices such as the following when viewing the **Tools ▸ Scheduled Actions** admin screen:

Admin notice: 'a new queue has begun processing'

Admin notice: 'next queue will begin in approx x seconds'

This could be confusing to the site owner (suppose here that the developer they hired has moved on to another project but inadvertently left the constant in place), since it implies queue processing is happening or is about to happen, even though the constant is in place and nothing will actually happen. Hiding these when appropriate seems like a sensible measure, and it could also be worth adding a new admin notice making it clear that queue runners have been disabled. Possible example:

Proposed admin notice: 'The queue is not currently being processed, as AS_DISABLE_QUEUE_RUNNERS is enabled. Learn more (link).'

### WP CLI

This impacts both WP Cron and HTTP-spawned queue runners, but not the WP CLI queue runner. We'd probably also want to cater to the WP CLI based runner, because on some sites that may be how actions are processed (ie, via a crontab rule). Also, in line with the previous point, we should probably add appropriate messaging if `wp action-scheduler run` is invoked while the constant is in place.

*Optionally,* we could also consider adding support for a `--force` flag, so that the queue can still be manually processed via WP CLI if desired.

### Custom Queue Runners

Third parties are free to develop and register their own queue runners. There are two takes on this: 1) it is up to them to check for the existence of this new constant and/or 2) we should help them out, possibly by adding a further condition inside the abstract queue runner's `process_action()` method.

Mte90:
Basically yes we need to disable AS and the code right now just disable the filter.

PR #863, opened 2022-09-23
Fix past-due with Async actions

### LINKED ISSUE #860: Past-due tab seems not to be working correctly with asynchronous actions
We use Action Scheduler to register many async actions at the same time. Since 3.5 update of Action Scheduler all our actions are set to `past-due` **just** after being scheduled and it's very confusing for our customers.

It looks like this update doesn't play well with async actions as the date for them can't be set:
```
array (
  'hook' => 'async_request_url',
  'status' => 'pending',
  'scheduled_date_gmt' => '0000-00-00 00:00:00',
  'scheduled_date_local' => '0000-00-00 00:00:00',
  'schedule' => 'O:28:"ActionScheduler_NullSchedule":0:{}',
  'group_id' => 10,
)
```
Is that something you are aware of, or maybe we're missing the point here?

engahmeds3ed:
Fixes: https://github.com/woocommerce/action-scheduler/issues/860

Here we created a new query argument `include_async_actions` with true as a default value not to break any current functionality as mentioned in the issue's description.

I passed this new argument three times as follows:
1. In AdminView class: in the part that controls the past-due admin notice.
2. In Store class: in the part of showing the status filter.
3. In ListTable class: in the part of showing the status filtered page.

Plz note that I fixed most of PHPCS formatting issues that were not related to this PR.

This is my first contribution to Action Scheduler and won't be the last one, I can see more rooms to enhance inside this amazing tool.

### DIFF
--- a/classes/ActionScheduler_AdminView.php
+++ b/classes/ActionScheduler_AdminView.php
@@ -2,25 +2,42 @@
 
 /**
  * Class ActionScheduler_AdminView
+ *
  * @codeCoverageIgnore
  */
 class ActionScheduler_AdminView extends ActionScheduler_AdminView_Deprecated {
 
-	private static $admin_view = NULL;
+	/**
+	 * AdminBiew single instance.
+	 *
+	 * @var ActionScheduler_AdminView|null
+	 */
+	private static $admin_view = null;
 
+	/**
+	 * AdminView screen ID.
+	 *
+	 * @var string
+	 */
 	private static $screen_id = 'tools_page_action-scheduler';
 
-	/** @var ActionScheduler_ListTable */
+	/**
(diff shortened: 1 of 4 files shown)


barryhughes:
Pausing here, since https://github.com/woocommerce/action-scheduler/pull/898 may change the approach we require.

barryhughes:
https://github.com/woocommerce/action-scheduler/pull/898 has been merged and should be released in `3.5.4`. I think that reduces the need for this change, so will go ahead and close unless you have objections. Don't hesitate to let me know if so.

barryhughes:
Closing per earlier note—feel free to create a new issue if this poses any problems/concerns. Thanks!

PR #864, opened 2022-09-27
Test/version update

roykho:
Here is a first take on making version bump automation work for AS using Woorelease. With these changes, I am able to put in logic on Woorelease side to account for "letter" based versions to be able to bump them.

The resulting diff changes after running Woorelease would look like so:

```
diff --git a/action-scheduler.php b/action-scheduler.php
index 7549dbc..c04c35d 100644
--- a/action-scheduler.php
+++ b/action-scheduler.php
@@ -5,7 +5,7 @@
  * Author: Automattic
- * Version: 3.5.2
+ * Version: 3.5.3
  * License: GPLv3
  *
  * Copyright 2019 Automattic, Inc.  (https://automattic.com/contact/)
@@ -26,29 +26,29 @@
  * @package ActionScheduler
  */

-if ( ! function_exists( 'action_scheduler_register_3_dot_5_dot_2' ) && function_exists( 'add_action' ) ) { // WRCS: DEFINED_VERSION.
+if ( ! function_exists( 'action_scheduler_register_3_dot_5_dot_3' ) && function_exists( 'add_action' ) ) { // WRCS: DEFINED_VERSION.

-	define( 'ACTION_SCHEDULER_VERSION', '3.5.2' ); // WRCS: DEFINED_VERSION.
+	define( 'ACTION_SCHEDULER_VERSION', '3.5.3' ); // WRCS: DEFINED_VERSION.

 	if ( ! class_exists( 'ActionScheduler_Versions', false ) ) {
 		require_once __DIR__ . '/classes/ActionScheduler_Versions.php';
 		add_action( 'plugins_loaded', array( 'ActionScheduler_Versions', 'initialize_latest_version' ), 1, 0 );
 	}

-	add_action( 'plugins_loaded', 'action_scheduler_register_3_dot_5_dot_2', 0, 0 ); // WRCS: DEFINED_VERSION.
+	add_action( 'plugins_loaded', 'action_scheduler_register_3_dot_5_dot_3', 0, 0 ); // WRCS: DEFINED_VERSION.

 	/**
 	 * Registers this version of Action Scheduler.
 	 */
-	function action_scheduler_register_3_dot_5_dot_2() { // WRCS: DEFINED_VERSION.
+	function action_scheduler_register_3_dot_5_dot_3() { // WRCS: DEFINED_VERSION.
-		$versions->register( ACTION_SCHEDULER_VERSION, 'action_scheduler_initialize_3_dot_5_dot_2' ); // WRCS: DEFINED_VERSION.
+		$versions->register( ACTION_SCHEDULER_VERSION, 'action_scheduler_initialize_3_dot_5_dot_3' ); // WRCS: DEFINED_VERSION.
 	}

 	/**
 	 * Initializes this version of Action Scheduler.
 	 */
-	function action_scheduler_initialize_3_dot_5_dot_2() { // WRCS: DEFINED_VERSION.
+	function action_scheduler_initialize_3_dot_5_dot_3() { // WRCS: DEFINED_VERSION.
 		// A final safety check is required even here, because historic versions of Action Scheduler
 		// followed a different pattern (in some unusual cases, we could reach this point and the
 		// ActionScheduler class is already defined—so we need to guard against that).
@@ -60,7 +60,7 @@ if ( ! function_exists( 'action_scheduler_register_3_dot_5_dot_2' ) && function_

 	if ( did_action( 'plugins_loaded' ) && ! doing_action( 'plugins_loaded' ) && ! class_exists( 'ActionScheduler', false ) ) {
-		action_scheduler_initialize_3_dot_5_dot_2(); // WRCS: DEFINED_VERSION.
+		action_scheduler_initialize_3_dot_5_dot_3(); // WRCS: DEFINED_VERSION.
 		do_action( 'action_scheduler_pre_theme_init' );
 		ActionScheduler_Versions::initialize_latest_version();
 	}
```

So basically I am adding `// WRCS: DEFINED_VERSION.` so Woorelease knows these are the lines to look for version bumps. Note this is not new. We've been using this code sniff for most of our other extensions that we need Woorelease to deploy.

### DIFF
--- a/action-scheduler.php
+++ b/action-scheduler.php
@@ -28,6 +28,8 @@
 
 if ( ! function_exists( 'action_scheduler_register_3_dot_5_dot_2' ) && function_exists( 'add_action' ) ) { // WRCS: DEFINED_VERSION.
 
+	define( 'ACTION_SCHEDULER_VERSION', '3.5.2' ); // WRCS: DEFINED_VERSION.
+
 	if ( ! class_exists( 'ActionScheduler_Versions', false ) ) {
 		require_once __DIR__ . '/classes/ActionScheduler_Versions.php';
 		add_action( 'plugins_loaded', array( 'ActionScheduler_Versions', 'initialize_latest_version' ), 1, 0 );
@@ -40,7 +42,7 @@ if ( ! function_exists( 'action_scheduler_register_3_dot_5_dot_2' ) && function_
 	 */
 	function action_scheduler_register_3_dot_5_dot_2() { // WRCS: DEFINED_VERSION.
 		$versions = ActionScheduler_Versions::instance();
-		$versions->register( '3.5.2', 'action_scheduler_initialize_3_dot_5_dot_2' ); // WRCS: DEFINED_VERSION.
+		$versions->register( ACTION_SCHEDULER_VERSION, 'action_scheduler_initialize_3_dot_5_dot_2' ); // WRCS: DEFINED_VERSION.
 	}
 
 	/**
--- a/action-scheduler.php
+++ b/action-scheduler.php
@@ -26,27 +26,27 @@
  * @package ActionScheduler
  */
 
-if ( ! function_exists( 'action_scheduler_register_3_dot_5_dot_2' ) && function_exists( 'add_action' ) ) {
+if ( ! function_exists( 'action_scheduler_register_3_dot_5_dot_2' ) && function_exists( 'add_action' ) ) { // WRCS: DEFINED_VERSION.
 
 	if ( ! class_exists( 'ActionScheduler_Versions', false ) ) {
 		require_once __DIR__ . '/classes/ActionScheduler_Versions.php';
 		add_action( 'plugins_loaded', array( 'ActionScheduler_Versions', 'initialize_latest_version' ), 1, 0 );
 	}
 
-	add_action( 'plugins_loaded', 'action_scheduler_register_3_dot_5_dot_2', 0, 0 );
+	add_action( 'plugins_loaded', 'action_scheduler_register_3_dot_5_dot_2', 0, 0 ); // WRCS: DEFINED_VERSION.
 
 	/**
 	 * Registers this version of Action Scheduler.
 	 */
-	function action_scheduler_register_3_dot_5_dot_2() {
+	function action_scheduler_register_3_dot_5_dot_2() { // WRCS: DEFINED_VERSION.
 		$versions = ActionScheduler_Versions::instance();
-		$versions->register( '3.5.2', 'action_scheduler_initialize_3_dot_5_dot_2' );
+		$versions->register( '3.5.2', 'action_scheduler_initialize_3_dot_5_dot_2' ); // WRCS: DEFINED_VERSION.
 	}
 
 	/**
 	 * Initializes this version of Action Scheduler.
 	 */
-	function action_scheduler_initialize_3_dot_5_dot_2() {
+	function action_scheduler_initialize_3_dot_5_dot_2() { // WRCS: DEFINED_VERSION.
 		// A final safety check is required even here, because historic versions of Action Scheduler
 		// followed a different pattern (in some unusual cases, we could reach this point and the
 		// ActionScheduler class is already defined—so we need to guard against that).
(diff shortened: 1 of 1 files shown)


action-scheduler.php
```diff
@@ -26,27 +26,29 @@
  * @package ActionScheduler
  */
 
-if ( ! function_exists( 'action_scheduler_register_3_dot_5_dot_2' ) && function_exists( 'add_action' ) ) {
+if ( ! function_exists( 'action_scheduler_register_3_dot_5_dot_2' ) && function_exists( 'add_action' ) ) { // WRCS: DEFINED_VERSION.
+
+	define( 'ACTION_SCHEDULER_VERSION', '3.5.2' ); // WRCS: DEFINED_VERSION.
```
barryhughes: Projecting forward a little, suppose we have Plugin A (loads first) with Action Scheduler 3.5.2 and Plugin B (loads second) with Action Scheduler 3.5.3:

- Plugin A will load Action Scheduler first.
- Function `action_scheduler_register_3_dot_5_dot_2` does not exist, so we enter this block.
- We define `ACTION_SCHEDULER_VERSION`.
- Later in the request, Plugin B loads its copy of Action Scheduler.
- Function `action_scheduler_register_3_dot_5_dot_3` does not exist, so we enter Plugin B's copy of this block.
- We attempt to define `ACTION_SCHEDULER_VERSION` a second time:
    - This leads to a notice, warning or error (depending on PHP version).
    - The constant will still be set to the initial value (`3.5.2` in this scenario, as defined earlier via Plugin A), which will make the call to `$versions->register()` incorrect, in the case of Plugin B.

roykho: Fair point. Perhaps add validation? `if ( ! defined( 'ACTION_SCHEDULER_VERSION' ) ) {...}`

barryhughes (changes requested):
Left a note about version resolution ...

I'm actually wondering if something like this might be viable (not comprehensively tested!):

```php
( function () {
	// Guard against premature loading.
	if ( ! function_exists( 'add_action' ) ) {
		return;
	}

	// Initialize ActionScheduler_Versions if needed.
	if ( ! class_exists( 'ActionScheduler_Versions', false ) ) {
		require_once __DIR__ . '/classes/ActionScheduler_Versions.php';
		add_action( 'plugins_loaded', array( 'ActionScheduler_Versions', 'initialize_latest_version' ), 1, 0 );
	}

	/**
	 * Initializes this version of the library (ActionScheduler_Versions will call this function
	 * if appropriate, based on the presence of other versions of the library).
	 */
	$initialize_this_version = function () {
		// A final safety check is required even here, because historic versions of Action Scheduler
		// followed a different pattern (in some unusual cases, we could reach this point and the
		// ActionScheduler class is already defined—so we need to guard against that).
		if ( ! class_exists( 'ActionScheduler', false ) ) {
			require_once __DIR__ . '/classes/abstracts/ActionScheduler.php';
			ActionScheduler::init( __FILE__ );
		}
	};

	/**
	 * Registers the existence of this version of the library.
	 */
	$register_this_version = function () use ( $initialize_this_version ) {
		ActionScheduler_Versions::instance()->register( '3.5.2', $initialize_this_version );
	};

	if ( did_action( 'plugins_loaded' ) && ! doing_action( 'plugins_loaded' ) && ! class_exists( 'ActionScheduler', false ) ) {
		$initialize_this_version();
		do_action( 'action_scheduler_pre_theme_init' );
		ActionScheduler_Versions::initialize_latest_version();
	}
	
	// Trigger the registration process.
	add_action( 'plugins_loaded', $register_this_version, 0, 0 );
} )();
```

I'm not sure how this would play with the WooRelease tool, but it would mean that we have only two places in the file where we reference the version number:

- The plugin header
- The call inside `$register_this_version`

It also means that, when we reference it, the version number is always in the same dot-separated format—instead of a mix of that plus the snake cased function names (ie, *"register_3_dot_5_dot_2")* we currently use.

(Again, this is just an early/provisional idea.)

action-scheduler.php
barryhughes: That would guard against the constant-already-defined notice/warning/error, but I *think* it would still lead to problems, because if an early version of Action Scheduler is loaded first then that is the version we capture in the constant (and so when a more recent version of Action Scheduler is also present, but loads second, we'd have essentially the same problem in that it registers itself as an earlier version than it really is).

roykho:
@barryhughes - as long as we can add `// WRCS: DEFINED_VERSION.` to the line with the version `ActionScheduler_Versions::instance()->register( '3.5.2', $initialize_this_version );`, WR can pick it up.

roykho:
@barryhughes so just a note that this PR change is really so that we can utilize Woorelease for automation without accidentally releasing with the issue we had in the past with versions. I wasn't thinking of refactoring to circumvent the issue altogether. As this is just a version bump, I feel at least for now, it might be better to stick with lower risk change.

With that said, I could easily not include this line of change `define( 'ACTION_SCHEDULER_VERSION', '3.5.3' ); // WRCS: DEFINED_VERSION.` and also bump this line `https://github.com/woocommerce/action-scheduler/blob/trunk/action-scheduler.php#L43` with `preg_replace`. That would make the changes needed only with comments which is low risk.

barryhughes:
Yep, all on board with a lower risk strategy. If we stick with a string literal instead of a constant, I think this will be A-OK :+1:

barryhughes (approved):
LGTM, all the relevant lines seem to be covered :+1:

PR #868, opened 2022-10-10
Prevent `undefined variable` warning for `$num_pastdue_actions`

### LINKED ISSUE #865: Using "action_scheduler_pastdue_actions_check_pre" causes PHP warnings
Using the following filter hook:

https://github.com/woocommerce/action-scheduler/blob/c2aaea225ac47d5654418223cbd4b0fdfc3910b5/classes/ActionScheduler_AdminView.php#L151

Returning a non-null, non-falsey value results in a PHP warning:

`Warning: Undefined variable $num_pastdue_actions in *****/action-scheduler/classes/ActionScheduler_AdminView.php on line 189`

This is happening because `$num_pastdue_actions` is defined **only if the filter is not in use**, but later in the code the variable is used regardless of the filter being used or not.

Can we have a fix for this? Thanks in advance.

stephymiehle:
Closes #865

> This is happening because `$num_pastdue_actions` is defined only if the filter is not in use, but later in the code the variable is used regardless of the filter being used or not.

Sets a fallback value of 0 to prevent the `Undefined variable $num_pastdue_actions` warnings when `$num_pastdue_actions` is referenced lower in the document after the initial `$check`

### Changelog

> Fix - Prevents an undefined variable error in relation to the past-due actions notice.

### DIFF
--- a/classes/ActionScheduler_AdminView.php
+++ b/classes/ActionScheduler_AdminView.php
@@ -147,6 +147,11 @@ class ActionScheduler_AdminView extends ActionScheduler_AdminView_Deprecated {
 		$threshold_seconds = ( int ) apply_filters( 'action_scheduler_pastdue_actions_seconds', DAY_IN_SECONDS );
 		$threshhold_min    = ( int ) apply_filters( 'action_scheduler_pastdue_actions_min', 1 );
 
+		# Set fallback value for past-due actions count.
+		if ( ! isset( $num_pastdue_actions ) ) {
+			$num_pastdue_actions = 0;
+		}
+
 		# Allow third-parties to preempt the default check logic.
 		$check = apply_filters( 'action_scheduler_pastdue_actions_check_pre', null );
 
--- a/classes/ActionScheduler_AdminView.php
+++ b/classes/ActionScheduler_AdminView.php
@@ -147,13 +147,16 @@ class ActionScheduler_AdminView extends ActionScheduler_AdminView_Deprecated {
 		$threshold_seconds = ( int ) apply_filters( 'action_scheduler_pastdue_actions_seconds', DAY_IN_SECONDS );
 		$threshhold_min    = ( int ) apply_filters( 'action_scheduler_pastdue_actions_min', 1 );
 
-		# Set fallback value for past-due actions count.
-		if ( ! isset( $num_pastdue_actions ) ) {
-			$num_pastdue_actions = 0;
-		}
+		// Set fallback value for past-due actions count.
+		$num_pastdue_actions = 0;
 
-		# Allow third-parties to preempt the default check logic.
+		// Allow third-parties to preempt the default check logic.
 		$check = apply_filters( 'action_scheduler_pastdue_actions_check_pre', null );
+
+		// If no third-party preempted and there are no past-due actions, return early.
+		if ( ! is_null( $check ) ) {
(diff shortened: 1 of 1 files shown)


classes/ActionScheduler_AdminView.php
```diff
@@ -147,6 +147,11 @@ protected function check_pastdue_actions() {
 		$threshold_seconds = ( int ) apply_filters( 'action_scheduler_pastdue_actions_seconds', DAY_IN_SECONDS );
 		$threshhold_min    = ( int ) apply_filters( 'action_scheduler_pastdue_actions_min', 1 );
 
+		# Set fallback value for past-due actions count.
```
barryhughes: A quick note that we use `//` for single-line comments *("Perl-style comments are not allowed").* I realize you are following the existing style within this method, but for new code we should stick to the project's current coding standard.

To run the linter for a specific file, you can do something like this (of course, you don't need to worry about existing issues, only problems with newly introduced or changed code):

`composer run phpcs classes/ActionScheduler_AdminView.php`

classes/ActionScheduler_AdminView.php
```diff
@@ -147,6 +147,11 @@ protected function check_pastdue_actions() {
...
+		if ( ! isset( $num_pastdue_actions ) ) {
```
barryhughes: I don't think the conditional is necessary. This close to the start of the method, we know that `$num_pastdue_actions` will not yet be set.

classes/ActionScheduler_AdminView.php
```diff
@@ -147,6 +147,11 @@ protected function check_pastdue_actions() {
...
+			$num_pastdue_actions = 0;
```
barryhughes: This successfully removes the potential for a PHP warning/error (relating to `$num_pastdue_actions` being undefined), but I'm not sure if it's the very best solution because, if it is used, we'll end up forming a notice like the following:

> **Action Scheduler:** 0 [past-due actions](#admin-link) found; something may be wrong. [Read documentation &raquo;](https://actionscheduler.org/faq/#my-site-has-past-due-actions-what-can-i-do)

There's no point emitting a warning at all in this case (if there are zero past due actions). So, it may make more sense to expand:

```php
$check = apply_filters( 'action_scheduler_pastdue_actions_check_pre', null );
```

To:

```php
$check = apply_filters( 'action_scheduler_pastdue_actions_check_pre', null );

if ( ! is_null( $check ) ) {
	return;
}
```

Or else we could consider using the non-null value of `$check` to set `$num_pastdue_actions` ... but if so we'd want to add a docblock for the `action_scheduler_pastdue_actions_check_pre` hook and clearly explain this. I hope that makes sense, but let me know if I can clarify anything.

barryhughes (changes requested):
Thanks for putting this fix together, and apologies for the delay in reviewing. I've left some notes explaining where we would need changes.

classes/ActionScheduler_AdminView.php
stephymiehle: Thanks so much for your thorough review! All updates made, and I agree that an early return here is a more straightforward opportunity.

barryhughes:
Thanks! Looks good, just running our test suite and barring no failures we can merge :+1:

barryhughes (approved):

PR #879, opened 2022-10-28
throwing an exception if as_schedule_recurring_action interval param is not of type integer

hsein-bitar:
While working on a project, I passed in a string ( 'twicedaily' in particular ) as the interval parameter of as_schedule_recurring_action. ( mistaken for wp_schedule_event params, an older implementation in our project )

That did not throw any errors, but it was silently causing many actions to be scheduled and filling the database very quickly.

Suggested/implemented fix: check/assert the parameter to be an integer and otherwise throw an exception to warn the developer of such behavior. 

If you have any feedback please let me know, this is my first contribution to an open-source project ever.

### Testing instructions

I recommend testing via WP CLI. Try creating recurring actions: if the interval is a valid value the action will be created, otherwise it will not be created and a suitable warning emitted. Example `wp shell` session:

```
>>> as_schedule_recurring_action( time(), 10, 'valid' )
=> 5220

>>> as_schedule_recurring_action( time(), 100.0, 'valid' )
=> 5221

>>> as_schedule_recurring_action( time(), '1000', 'valid' )
=> 5222

>>> as_schedule_recurring_action( time(), 'accidental', 'invalid' )
PHP Notice:  Function as_schedule_recurring_action was called <strong>incorrectly</strong>. An integer was expected but &quot;accidental&quot; (string) was received. Please see <a href="https://wordpress.org/support/article/debugging-in-wordpress/">Debugging in WordPress</a> for more information. (This message was added in version 3.5.5.) in /data/var/www/lab.wordpress/public/wp-includes/functions.php on line 5835

>>> as_schedule_recurring_action( time(), -123.456, 'invalid' )
PHP Notice:  Function as_schedule_recurring_action was called <strong>incorrectly</strong>. An integer was expected but &quot;-123.456&quot; (double) was received. Please see <a href="https://wordpress.org/support/article/debugging-in-wordpress/">Debugging in WordPress</a> for more information. (This message was added in version 3.5.5.) in /data/var/www/lab.wordpress/public/wp-includes/functions.php on line 5835
```

### Changelog

> Add safety checks to the procedural API, relating to the interval for recurring actions.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -122,7 +122,6 @@ function as_schedule_recurring_action( $timestamp, $interval_in_seconds, $hook, 
 		return 0;
 	}
 
-	// if you pass a string by mistake, it causes a flood of scheduled actions, recurring almost every minute or less
 	$interval = (int) $interval_in_seconds;
 
 	// We expect an integer and allow it to be passed using float and string types, but otherwise
@@ -130,13 +129,13 @@ function as_schedule_recurring_action( $timestamp, $interval_in_seconds, $hook, 
 	if ( ! is_numeric( $interval_in_seconds ) || $interval_in_seconds != $interval ) {
 		_doing_it_wrong(
 			__METHOD__,
-			sprintf( 
+			sprintf(
 				/* translators: 1: provided value 2: provided type. */
 				__( 'An integer was expected but "%1$s" (%2%s) was received.', 'action-scheduler' ),
 				$interval_in_seconds,
 				gettype( $interval_in_seconds )
 			),
-			ActionScheduler_Versions::instance()->latest_version()
+			'3.6.0'
 		);
 	}
 
--- a/functions.php
+++ b/functions.php
@@ -122,6 +122,11 @@ function as_schedule_recurring_action( $timestamp, $interval_in_seconds, $hook, 
 		return 0;
 	}
 
+	// if you pass a string by mistake, it causes a flood of scheduled actions, recurring almost every minute or less
+	if ( ! is_int( $interval_in_seconds ) ) {
+		throw new Exception( "Interval parameter should be of type integer, received" . gettype( $interval_in_seconds ) );
+	}
+
 	/**
 	 * Provides an opportunity to short-circuit the default process for enqueuing recurring
 	 * actions.
--- a/functions.php
+++ b/functions.php
@@ -128,7 +128,16 @@ function as_schedule_recurring_action( $timestamp, $interval_in_seconds, $hook, 
 	// We expect an integer and allow it to be passed using float and string types, but otherwise
 	// should reject unexpected values.
 	if ( ! is_numeric( $interval_in_seconds ) || $interval_in_seconds != $interval ) {
-		throw new Exception( "Interval parameter should be of type integer, received" . gettype( $interval_in_seconds ) );
+		_doing_it_wrong(
+			__METHOD__,
+			sprintf( 
+				/* translators: 1: provided value 2: provided type. */
+				__( 'An integer was expected but "%1$s" (%2%s) was received.', 'action-scheduler' ),
+				$interval_in_seconds,
+				gettype( $interval_in_seconds )
+			),
+			ActionScheduler_Versions::instance()->latest_version()
+		);
 	}
 
 	/**
--- a/functions.php
+++ b/functions.php
@@ -123,7 +123,11 @@ function as_schedule_recurring_action( $timestamp, $interval_in_seconds, $hook, 
 	}
 
 	// if you pass a string by mistake, it causes a flood of scheduled actions, recurring almost every minute or less
-	if ( ! is_int( $interval_in_seconds ) ) {
+	$interval = (int) $interval_in_seconds;
+
+	// We expect an integer and allow it to be passed using float and string types, but otherwise
+	// should reject unexpected values.
+	if ( ! is_numeric( $interval_in_seconds ) || $interval_in_seconds != $interval ) {
 		throw new Exception( "Interval parameter should be of type integer, received" . gettype( $interval_in_seconds ) );
 	}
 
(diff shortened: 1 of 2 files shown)


functions.php
```diff
@@ -53,6 +53,10 @@ function as_schedule_single_action( $timestamp, $hook, $args = array(), $group =
  * @return int The action ID.
  */
 function as_schedule_recurring_action( $timestamp, $interval_in_seconds, $hook, $args = array(), $group = '', $unique = false ) {
+	// if you pass a string by mistake, it causes a flood of scheduled actions, recurring almost every minute or less
+	if ( ! is_int( $interval_in_seconds ) ) {
```
barryhughes: For backwards-compatibility reasons, a less strict approach may be preferable here.

Strings and floats should probably still be allowed, so long as they are used to convey an integer. Examples of this might include `(string) '60'` or `(float) 60.0`.

Possible iteration:

```suggestion
	$interval = (int) $interval_in_seconds;

	// We expect an integer and allow it to be passed using float and string types, but otherwise
	// should reject unexpected values.
	if ( ! is_numeric( $interval_in_seconds ) || $interval_in_seconds != $interval ) {
```

Examples of allowed values, using the above code:

- `(int) 60`
- `(float) 60.0`
- `(string) 60`

Examples of values that will be rejected:

- `(float) 60.5`
- `(string) 'twicedaily'`

Of course, if we follow this approach we would also ideally update the call to `ActionScheduler::factory()->recurring_unique()` a few lines down (and pass `$interval` rather than `$interval_in_seconds`).

barryhughes: To explain; if we were introducing a new function then we could be much stricter—but it's entirely possible there is code out there that passes in a int contained in a string, or similar, and we would not wish to inadvertently break it.

functions.php
```diff
@@ -53,6 +53,10 @@ function as_schedule_single_action( $timestamp, $hook, $args = array(), $group =
...
+		throw new Exception( "Interval parameter should be of type integer, received" . gettype( $interval_in_seconds ) );
```
barryhughes: Similar to the earlier comment, we may wish to take a softer stance. If there is existing code out there that is inadvertently making a mistake, we don't necessarily want to end the request with an exception. Possible iteration:

```suggestion
		_doing_it_wrong(
			__METHOD__,
			sprintf( 
				/* translators: 1: provided value 2: provided type. */
				__( 'An integer was expected but "%1$s" (%2%s) was received.', 'action-scheduler' ),
				$interval_in_seconds,
				gettype( $interval_in_seconds )
			),
			ActionScheduler_Versions::instance()->latest_version()
		);
```

This would also make the exception message translatable.

barryhughes (changes requested):
Thank you for submitting this, and our apologies for the delay in reviewing.

I left some requests for changes—this is the basis of a really nice improvement but, primarily for backward-compatibility reasons, we would need a few tweaks before we can ship it (unfortunately we cannot always be as strict as we might be if we were writing completely new code).

Please do let me know if you have any questions.

barryhughes:
Hi @hsein-bitar—just checking to see if you have any questions/concerns or are still interested in moving this one forward? Thanks!

hsein-bitar:
> Hi @hsein-bitar—just checking to see if you have any questions/concerns or are still interested in moving this one forward? Thanks!

Thanks a lot @barryhughes Sorry I did not notice the comments before. I did commit your changes. Thanks for the improvements and the thorough explanation. Kindly let me know if you have any additional feedback.

functions.php
```diff
@@ -53,6 +53,22 @@ function as_schedule_single_action( $timestamp, $hook, $args = array(), $group =
...
+	$interval = (int) $interval_in_seconds;
+
+	// We expect an integer and allow it to be passed using float and string types, but otherwise
+	// should reject unexpected values.
+	if ( ! is_numeric( $interval_in_seconds ) || $interval_in_seconds != $interval ) {
+		_doing_it_wrong(
+			__METHOD__,
+			sprintf( 
+				/* translators: 1: provided value 2: provided type. */
+				__( 'An integer was expected but "%1$s" (%2%s) was received.', 'action-scheduler' ),
```
barryhughes: Need to fix the second placeholder. Also, we can add a measure of safety by escaping the string:

```suggestion
				esc_html__( 'An integer was expected but "%1$s" (%2$s) was received.', 'action-scheduler' ),
```

Apologies for missing those things in the initial suggestion.

functions.php
```diff
@@ -53,6 +53,22 @@ function as_schedule_single_action( $timestamp, $hook, $args = array(), $group =
...
+				$interval_in_seconds,
+				gettype( $interval_in_seconds )
+			),
+			ActionScheduler_Versions::instance()->latest_version()
+		);
+	}
 	if ( ! ActionScheduler::is_initialized( __FUNCTION__ ) ) {
 		return 0;
 	}
```
barryhughes: I propose we keep this block at the top of the method.

That is, if Action Scheduler is not initialized, then bail out early *(before* testing if `$interval_in_seconds` is valid). It occurs that this may be safer since we are making use of `ActionScheduler_Versions::instance()` in our `_doing_it_wrong()` call.

barryhughes (dismissed):
Thanks! Just a couple more tweaks and I think we can move ahead.

barryhughes:
Added basic testing instructions. Requesting a fresh review since I've committed some modifications of my own.

barryhughes:
The magic reviewer selector wasn't doing it's thing, perhaps because I was already active in the PR. Peter, I selected you manually—hope that's ok.

barryhughes:
Rebased, and updated the version number used in the `_doing_it_wrong()` call.

jorgeatorres (approved):
LGTM. Thank you @hsein-bitar for your contribution and @barryhughes for the last few fixes.

PR #883, opened 2022-11-03
Correctly order the parameters for `ActionScheduler_ActionFactory`'s calls to `single_unique`

### LINKED ISSUE #881: The function recurring_unique() calls single_unique() with the $args argument in place of the $first argument
In ActionScheduler_ActionFactory.php, there are [these lines](https://github.com/woocommerce/action-scheduler/blob/trunk/classes/ActionScheduler_ActionFactory.php#L147-L155):

```php
public function recurring_unique( $hook, $args = array(), $first = null, $interval = null, $group = '', $unique = true ) {
    if ( empty( $interval ) ) {
        return $this->single_unique( $hook, $unique, $args, $first, $group );
    }
    $date     = as_get_datetime_object( $first );
    $schedule = new ActionScheduler_IntervalSchedule( $date, $interval );
    $action   = new ActionScheduler_Action( $hook, $args, $schedule, $group );
    return $unique ? $this->store_unique_action( $action ) : $this->store( $action );
}
```

Then `single_unique` reads [like this](https://github.com/woocommerce/action-scheduler/blob/6633378283d33746eec7314586783f58deee5375/classes/ActionScheduler_ActionFactory.php#L113-L118):

```php
public function single_unique( $hook, $args = array(), $when = null, $group = '', $unique = true ) {
    $date     = as_get_datetime_object( $when );
    $schedule = new ActionScheduler_SimpleSchedule( $date );
    $action   = new ActionScheduler_Action( $hook, $args, $schedule, $group );
    return $unique ? $this->store_unique_action( $action ) : $this->store( $action );
}
```

I had a user of the plugin I maintain report this as a potential issue. Here's what they said:

> I believe the issue is a difference in the ActionScheduler_ActionFactory.php class between 2.1.2 and 2.2.5. The function recurring_unique() calls single_unique() with the $args argument in place of the $first argument (hence an error when trying to parse an empty array as a string)

For further context, see [this forum thread](https://wordpress.org/support/topic/scheduling-prompts-a-critical-error/). I've never seen the issue the user is reporting, but it does indeed look to me as though `recurring_unique` is passing parameters in a different order than `single_unique` is expecting. I'm interested to see if this is in fact an error that should be fixed, or if we're all missing something.

Update: to clarify, version 2.1.2 of the plugin I maintain was running Action Scheduler 3.4.0, and 2.2.5 is running 3.5.2.

jonathanstegall:
I found three calls to the `single_unique` method and two of them had parameters that didn't match the documented order. This fixes them.

In doing a search, I only found one other use of this object's `single_unique` method, and its parameters were already in the documented order. If that's correct, I think this fixes #881.

### DIFF
--- a/classes/ActionScheduler_ActionFactory.php
+++ b/classes/ActionScheduler_ActionFactory.php
@@ -146,7 +146,7 @@ class ActionScheduler_ActionFactory {
 	 */
 	public function recurring_unique( $hook, $args = array(), $first = null, $interval = null, $group = '', $unique = true ) {
 		if ( empty( $interval ) ) {
-			return $this->single_unique( $hook, $unique, $args, $first, $group );
+			return $this->single_unique( $hook, $args, $first, $group, $unique );
 		}
 		$date     = as_get_datetime_object( $first );
 		$schedule = new ActionScheduler_IntervalSchedule( $date, $interval );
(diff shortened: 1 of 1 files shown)


MannyAdumbire:
I traced an issue to the `single_unique`  method also, but was at a loss as to why it was misbehaving. 
This patch fixed it for me. Hope it can be merged soon!

jorgeatorres (approved):
LGTM. Thanks @jonathanstegall for your contribution! 💯

PR #891, opened 2022-11-25
PHP 8.2: fix "Using ${var} in strings is deprecated, use {$var} instead"

rodrigoprimo:
PHP 8.2 deprecated using ${var} in strings (https://kinsta.com/blog/php-8-2/#deprecate--string-interpolation). {$var} should be used instead. This commit fixes the occurrences of ${var} inside ActionScheduler_StoreSchema.php to prepare for the next version of PHP which is scheduled to be released early in December.

I found this in the context of testing MailPoet with PHP 8.2. I'm not familiar with the Action Scheduler so I might be missing something. I tested these changes briefly in a MailPoet environment that uses AS.

### Changelog

> Dev - Update code using a deprecated style of variable interpolation, for improved compatibility with PHP 8.2.


barryhughes (approved):
Thank you!

PR #892, opened 2022-11-29
PHP 8.2: declare property to fix creation of dynamic property warning

rodrigoprimo:
This PR fixes the following deprecation warning that is being thrown by PHP 8.2 since this version of PHP deprecated dynamic properties:

```
Creation of dynamic property ActionScheduler_NullSchedule::$scheduled_date is deprecated /wp-core/wp-content/plugins/mailpoet/vendor/woocommerce/action-scheduler/classes/schedules/ActionScheduler_NullSchedule.php:14
```

I opted to declare the property as public to avoid introducing a backward incompatible change, but I'm not sure if this is the best approach or not.

### Changelog

> Fix - Prevent dynamic property warnings in relation to Null Schedule instances.

### DIFF
--- a/classes/schedules/ActionScheduler_NullSchedule.php
+++ b/classes/schedules/ActionScheduler_NullSchedule.php
@@ -6,7 +6,7 @@
 class ActionScheduler_NullSchedule extends ActionScheduler_SimpleSchedule {
 
 	/** @var DateTime|null */
-	public $scheduled_date;
+	protected $scheduled_date;
 
 	/**
 	 * Make the $date param optional and default to null.
--- a/classes/schedules/ActionScheduler_NullSchedule.php
+++ b/classes/schedules/ActionScheduler_NullSchedule.php
@@ -5,6 +5,9 @@
  */
 class ActionScheduler_NullSchedule extends ActionScheduler_SimpleSchedule {
 
+	/** @var DateTime|null */
+	public $scheduled_date;
+
 	/**
 	 * Make the $date param optional and default to null.
 	 *


classes/schedules/ActionScheduler_NullSchedule.php
```diff
@@ -5,6 +5,9 @@
  */
 class ActionScheduler_NullSchedule extends ActionScheduler_SimpleSchedule {
 
+	/** @var DateTime|null */
+	public $scheduled_date;
```
barryhughes: There are quite a few layers of inheritance here, but I have a suspicion that the constructor is assuming it can write to the property of the same name that exists in superclass [`ActionScheduler_Abstract_Schedule`](https://github.com/woocommerce/action-scheduler/blob/3.5.3/classes/abstracts/ActionScheduler_Abstract_Schedule.php#L13), which is private (and perhaps this was possible in much earlier versions of PHP?).

With that in mind, I think it may make more sense to resolve this by making `ActionScheduler_Abstract_Schedule::scheduled_date` protected, vs exposing a new public property. What do you think?

barryhughes (changes requested):

classes/schedules/ActionScheduler_NullSchedule.php
barryhughes: ...Also, our apologies for the delay in getting to this PR: if it's helpful/if you're focused on other things, we can drive this forward.

rodrigoprimo: Thanks for the review, @barryhughes. I wasn't sure what would be the best in terms of the visibility of the property. I went ahead an pushed a new commit making it protected. Can you please check this PR again?

barryhughes: Thanks, @rodrigoprimo! I originally meant updating the property already defined in `ActionScheduler_Abstract_Schedule` (which this class inherits from), but this also works and is perhaps a safer/more conservative change to make. Looks good :+1:

barryhughes (approved):

PR #893, opened 2022-12-08
Fetch action in memory first before releasing claim to avoid deadlock.

vedanshujain:
Earlier, we used to release claim in a atomic query, i.e. we would update all actions belonging to a particular claim_id with claim_id = 0. While this was functionally correct, it would cause deadlock since the update query will hold a lock on the claim_id_.. index on the action table.

This allowed the possibility of a race condition, if the claimer query is also running at the same time, then the claimer query will also try to acquire a lock on the claim_id_.. index, and in this case, if claim release query has already progressed to the point of acquiring the lock, but have not updated yet, it would cause a deadlock.

We resolve this by getting all the actions_id that we want to release the claim in a separate query and then releasing the claim on each of them. This way, our lock is acquired on the action_id index instead of the claim_id index. Note that the lock on claim_id will still be acquired, but only when we make the update rather than when we select the actions.

### DIFF
--- a/classes/data-stores/ActionScheduler_DBStore.php
+++ b/classes/data-stores/ActionScheduler_DBStore.php
@@ -943,14 +943,19 @@ AND `group_id` = %d
 		 * We resolve this by getting all the actions_id that we want to release claim from in a separate query, and then releasing the claim on each of them. This way, our lock is acquired on the action_id index instead of the claim_id index. Note that the lock on claim_id will still be acquired, but it will only when we actually make the update, rather than when we select the actions.
 		 */
 		$action_ids = $wpdb->get_col( $wpdb->prepare( "SELECT action_id FROM {$wpdb->actionscheduler_actions} WHERE claim_id = %d", $claim->get_id() ) );
-		if ( 0 === count( $action_ids ) ) {
-			$wpdb->delete( $wpdb->actionscheduler_claims, array( 'claim_id' => $claim->get_id() ), array( '%d' ) );
+
+		if ( count( $action_ids ) > 0 ) {
+			$action_id_string = implode( ',', array_map( 'absint', $action_ids ) );
+			$row_updates = $wpdb->query( "UPDATE {$wpdb->actionscheduler_actions} SET claim_id = 0 WHERE action_id IN ({$action_id_string})" ); // phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared
+			if ( $row_updates < count( $action_ids ) ) {
+				throw new RuntimeException(
+					sprintf(
+						__( 'Unable to release actions from claim id %d.', 'woocommerce' ),
+						$claim->get_id()
+					)
+				);
+			}
 			return;
-		}
-		$action_id_string = implode( ',', array_map( 'absint', $action_ids ) );
-		$row_updates = $wpdb->query( "UPDATE {$wpdb->actionscheduler_actions} SET claim_id = 0 WHERE action_id IN ({$action_id_string})" ); // phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared
-		if ( $row_updates < count( $action_ids ) ) {
-			throw new RuntimeException( __( 'Unable to release actions from claim. Database error.', 'woocommerce' ) );
 		}
 		$wpdb->delete( $wpdb->actionscheduler_claims, array( 'claim_id' => $claim->get_id() ), array( '%d' ) );
 	}
--- a/classes/data-stores/ActionScheduler_DBStore.php
+++ b/classes/data-stores/ActionScheduler_DBStore.php
@@ -944,20 +944,22 @@ AND `group_id` = %d
 		 */
 		$action_ids = $wpdb->get_col( $wpdb->prepare( "SELECT action_id FROM {$wpdb->actionscheduler_actions} WHERE claim_id = %d", $claim->get_id() ) );
 
+		$row_updates = 0;
 		if ( count( $action_ids ) > 0 ) {
 			$action_id_string = implode( ',', array_map( 'absint', $action_ids ) );
 			$row_updates = $wpdb->query( "UPDATE {$wpdb->actionscheduler_actions} SET claim_id = 0 WHERE action_id IN ({$action_id_string})" ); // phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared
-			if ( $row_updates < count( $action_ids ) ) {
-				throw new RuntimeException(
-					sprintf(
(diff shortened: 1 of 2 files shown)


WPprodigy (commented):
There is also the option to use `$claim->get_actions()` for the list of IDs rather than the first SELECT.

I can't say for sure if it's better though, as the extra SELECT is possibly a bit more protection against actions that may have been reclaimed already - although I'm not sure if that's very likely for a reclaim to happen before here given the time limits involved.

In either case, we may want to note the decision on why we chose one way vs the other.

WPprodigy:
Also, slight refactor suggestion to avoid the duplicate claim delete:

```
function release_claim( ActionScheduler_ActionClaim $claim ) {
	global $wpdb;

	$action_ids  = $wpdb->get_col( $wpdb->prepare( "SELECT action_id FROM {$wpdb->actionscheduler_actions} WHERE claim_id = %d", $claim->get_id() ) );
	$row_updates = 0;

	if ( 0 !== count( $action_ids ) ) {
		$action_id_string = implode( ',', array_map( 'absint', $action_ids ) );
		$row_updates = $wpdb->query( "UPDATE {$wpdb->actionscheduler_actions} SET claim_id = 0 WHERE action_id IN ({$action_id_string})" );
	}

	$wpdb->delete( $wpdb->actionscheduler_claims, array( 'claim_id' => $claim->get_id() ), array( '%d' ) );

	if ( $row_updates < count( $action_ids ) ) {
		throw new RuntimeException( __( 'Unable to release all actions from claim. Database error.', 'woocommerce' ) );
	}
}
```

(should still delete from actionscheduler_claims before throwing error)

WPprodigy:
@vedanshujain just checking back in on this. Would be helpful to have this merged in so a few customers could run the standalone version from `trunk` here to get all the deadlock patches at once in an upgrade-safe way.

classes/data-stores/ActionScheduler_DBStore.php
```diff
@@ -935,7 +935,23 @@ public function find_actions_by_claim_id( $claim_id ) {
...
+		 * While this was functionally correct, it would cause deadlock, since this update query will hold a lock on the claim_id_.. index on the action table.
+		 * This allowed the possibility of a race condition, where the claimer query is also running at the same time, then the claimer query will also try to acquire a lock on the claim_id_.. index, and in this case if claim release query has already progressed to the point of acquiring the lock, but have not updated yet, it would cause a deadlock.
+		 *
+		 * We resolve this by getting all the actions_id that we want to release claim from in a separate query, and then releasing the claim on each of them. This way, our lock is acquired on the action_id index instead of the claim_id index. Note that the lock on claim_id will still be acquired, but it will only when we actually make the update, rather than when we select the actions.
+		 */
+		$action_ids = $wpdb->get_col( $wpdb->prepare( "SELECT action_id FROM {$wpdb->actionscheduler_actions} WHERE claim_id = %d", $claim->get_id() ) );
```
barryhughes, line 945: Very minor nit, but we call `$claim->get_id()` twice (once here, then one more time depending on which branch we follow) ... could be worth placing it in a variable instead.

classes/data-stores/ActionScheduler_DBStore.php
```diff
@@ -935,7 +935,23 @@ public function find_actions_by_claim_id( $claim_id ) {
...
+		if ( 0 === count( $action_ids ) ) {
+			$wpdb->delete( $wpdb->actionscheduler_claims, array( 'claim_id' => $claim->get_id() ), array( '%d' ) );
+			return;
+		}
+		$action_id_string = implode( ',', array_map( 'absint', $action_ids ) );
+		$row_updates = $wpdb->query( "UPDATE {$wpdb->actionscheduler_actions} SET claim_id = 0 WHERE action_id IN ({$action_id_string})" ); // phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared
+		if ( $row_updates < count( $action_ids ) ) {
+			throw new RuntimeException( __( 'Unable to release actions from claim. Database error.', 'woocommerce' ) );
```
```suggestion
			throw new RuntimeException( 
				sprintf(
					/* translators: %d claim ID. */
					__( 'Unable to release all actions from claim %d.', 'woocommerce' ),
					$claim->get_id()
				);
```

Could something like this be more helpful in the event the exception is thrown?

barryhughes (commented):
Left a couple of minor notes. Beyond those things, I think @WPprodigy flags a good point here:

> There is also the option to use $claim->get_actions() for the list of IDs rather than the first SELECT.
> 
> I can't say for sure if it's better though, as the extra SELECT is possibly a bit more protection against actions that may have been reclaimed already.

My own take: I think the fresh `SELECT` is worthwhile, since though it is unlikely that actions will have prematurely been released from the claim they may indeed have been deleted (ie, by another action that was part of the same claim).

classes/data-stores/ActionScheduler_DBStore.php
vedanshujain: Thanks for the suggestion, addressed.

classes/data-stores/ActionScheduler_DBStore.php
vedanshujain: It should be fine I think, since get_id also just looks up the value from variable (and not fetch it from database)

vedanshujain:
@WPprodigy @barryhughes I have addressed both the comments; note that I am fetching the actions again using the select clause to preserve as much atomicity as possible. We can use the in-memory list, but this behavior is more consistent with what it was doing earlier.

classes/data-stores/ActionScheduler_DBStore.php
```diff
@@ -935,7 +935,28 @@ public function find_actions_by_claim_id( $claim_id ) {
...
+				throw new RuntimeException(
+					sprintf(
+						__( 'Unable to release actions from claim id %d.', 'woocommerce' ),
+						$claim->get_id()
+					)
+				);
+			}
+			return;
+		}
 		$wpdb->delete( $wpdb->actionscheduler_claims, array( 'claim_id' => $claim->get_id() ), array( '%d' ) );
```
WPprodigy, line 953: this should be called even if the Exception is going to be thrown

vedanshujain: should be resolved in https://github.com/woocommerce/action-scheduler/pull/893/commits/07f4dd0bf32ad9d8eea7aeca023f2de10ec9af94

barryhughes (approved):
LGTM! @WPprodigy just checking in for any other feedback before merging.

WPprodigy (approved):
Looks good!

PR #894, opened 2022-12-09
Calculate timeouts based on total actions

WPprodigy:
### Description

Before `run()` starts up a new batch, it checks if the timeout is likely to be exceeded based on how many actions have been processed so far in the given amount of time. So if 120 actions are processed in 60minutes, it thinks 2 seconds per action. It'll decide if it's safe to do another batch based on that. This is sound.

But after each action is processed inside of `do_batch()`, it does the same calculation. Except it tells the logic how many actions that batch itself has processed. So if we're on the 1st action the logic will see that 1 action has been processed in the given 60seconds. Clearly, a much different result and it will determine to bail inside of `do_batch()` far before it wants to bail inside of `run()`.

As a result, typically once you get towards the end of a queue's life, `run()` will keep creating new batches/claims but after the first action in the claim is processed it will bail and unclaim. This process can repeat A LOT before `run()` also decides it is time to bail.

This means there are tons of claim and unclaim queries flying about, heavily increasing the odds of deadlocks currently. So when things get really mess and the unclaim query starts failing, you can end up with 100s of different claim_ids that are each holding onto their own set of actions. So the system ends up blocked up until the cleanup is triggered once the claims grow stale. Overall, not great.

Preventing the deadlocks altogether is another issue, but here we help avoid it and just make the logic smarter so that if do_batch() aborts due to resources/time that means run() will also abort rather than starting off the vicious cycle.

### Testing

Can really showcase the issue if you set batch size to 10, set timeout to 120seconds, and register a few thousand dummy actions that will complete immediately. Then watch the queue process, and note that after a few batches in, `do_batch()` will start to abort after the 1st action whilst `run()` will just keep looping and making new claims.

### Considerations

This is going to result in queues running closer to the configured timeout limit. That's ideal, but incorrectly configured sites may find that problematic. The effects are particularly more notable on environments with larger time limits that are set to process multiple batches per queue/run (which I would wager is not super common).

### Changelog

> Fix - Prevent batches from ending prematurely, which can result in additional unnecessary claims.

### DIFF
--- a/classes/ActionScheduler_QueueRunner.php
+++ b/classes/ActionScheduler_QueueRunner.php
@@ -129,7 +129,7 @@ class ActionScheduler_QueueRunner extends ActionScheduler_Abstract_QueueRunner {
 		if ( false === $this->has_maximum_concurrent_batches() ) {
 			$batch_size = apply_filters( 'action_scheduler_queue_runner_batch_size', 25 );
 			do {
-				$processed_actions_in_batch = $this->do_batch( $batch_size, $context );
+				$processed_actions_in_batch = $this->do_batch( $batch_size, $context, $processed_actions );
 				$processed_actions         += $processed_actions_in_batch;
 			} while ( $processed_actions_in_batch > 0 && ! $this->batch_limits_exceeded( $processed_actions ) ); // keep going until we run out of actions, time, or memory
 		}
@@ -147,9 +147,10 @@ class ActionScheduler_QueueRunner extends ActionScheduler_Abstract_QueueRunner {
 	 * @param int $size The maximum number of actions to process in the batch.
 	 * @param string $context Optional identifer for the context in which this action is being processed, e.g. 'WP CLI' or 'WP Cron'
 	 *        Generally, this should be capitalised and not localised as it's a proper noun.
+	 * @param int $total_processed_actions The number of total actions processed so far.
 	 * @return int The number of actions processed.
 	 */
-	protected function do_batch( $size = 100, $context = '' ) {
+	protected function do_batch( $size = 100, $context = '', $total_processed_actions = 0 ) {
 		$claim = $this->store->stake_claim($size);
 		$this->monitor->attach($claim);
 		$processed_actions = 0;
@@ -162,7 +163,7 @@ class ActionScheduler_QueueRunner extends ActionScheduler_Abstract_QueueRunner {
 			$this->process_action( $action_id, $context );
 			$processed_actions++;
 
-			if ( $this->batch_limits_exceeded( $processed_actions ) ) {
+			if ( $this->batch_limits_exceeded( $processed_actions + $total_processed_actions ) ) {
 				break;
 			}
 		}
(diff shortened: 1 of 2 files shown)


vedanshujain (commented):
Thanks so much for this patch! Much appreciate and great work.

Approach works well, but I don't think we can add a new param to method even with a default value, since it will break inheriting classes after PHP8, so suggesting reframing the code a little bit (but keeping the same approach)

From your explanation, the `do_batch` does not really need its own context, so we can store it in a private variable of the class and use it. that way

Our run function will become:

```php
public function run( $context = 'WP Cron' ) {
    ActionScheduler_Compatibility::raise_memory_limit();
    ActionScheduler_Compatibility::raise_time_limit( $this->get_time_limit() );
    do_action( 'action_scheduler_before_process_queue' );
    $this->run_cleanup();
    $this->processed_actions_count = 0;
    if ( false === $this->has_maximum_concurrent_batches() ) {
	    $batch_size = apply_filters( 'action_scheduler_queue_runner_batch_size', 25 );
	    do {
		    $processed_actions_in_batch = $this->do_batch( $batch_size, $context );
	    } while ( $processed_actions_in_batch > 0 && ! $this->batch_limits_exceeded( $this->processed_actions_count ) ); // keep going until we run out of actions, time, or memory
    }
    
    do_action( 'action_scheduler_after_process_queue' );
    return $this->processed_actions_count;
}
```

and then do_batch will use the $this->processed_actions_count like so:

```php
protected function do_batch( $size = 100, $context = '' ) {
    $claim = $this->store->stake_claim($size);
    $this->monitor->attach($claim);
    $processed_actions_before_this_batch = $this->processed_actions_count;
    foreach ( $claim->get_actions() as $action_id ) {
	    // bail if we lost the claim
	    if ( ! in_array( $action_id, $this->store->find_actions_by_claim_id( $claim->get_id() ) ) ) {
		    break;
	    }
	    $this->process_action( $action_id, $context );
	    $this->processed_actions_count++;
    
	    if ( $this->batch_limits_exceeded( $this->processed_actions_count ) ) {
		    break;
	    }
    }
    $this->store->release_claim($claim);
    $this->monitor->detach();
    $this->clear_caches();
    return $this->processed_actions_count - $processed_actions_before_this_batch;
}
```


wdyt?

WPprodigy:
> since it will break inheriting classes after PHP8

<details>
  <summary>A small tangent :)</summary>
  
Seems unfortunate if we're forever locked out of altering the signatures of `protected` methods.
  
Nothing in the plugin is inheriting from this `ActionScheduler_QueueRunner` class atm, so it'd have to be some custom code from a user that makes their own class AND overrides the protected `do_batch()` method. That feels like a lot of IFs, and anything extending a class should probably anticipate dealing with compatibility fixes during updates.

I only say this because switching to a class property / more "global" state does feel like a bit of a complexity increase that could bite us in the future.
</details>

All in all though, happy to make this change still - especially since this route means we can add it in right away vs waiting for a major version increment to signify backwards-compat changes.

WPprodigy:
Failing tests seem unrelated, not sure what's up? I don't have permissions in this repo to re-run them.

vedanshujain (approved):
LGTM!

vedanshujain:
Asked for a quick second look before merging from @barryhughes

barryhughes:
- Love it ❤️ 
- Added changelog to description 🗒️ 
- Merging (provisionally targeting 3.5.4) ⏬

PR #897, opened 2022-12-09
Add pre filters during action registration

WPprodigy:
Add some filters that will allow interjecting yourself into the registration of certain actions.

Main use case is that this will give a level of control over to sites to "nerf" certain plugins that are registering actions they would like to handle differently. For example, the new filter would allow a site to just handle `wc-admin_import_customers` inline rather than having it schedule an action.

Kind of mirror-ed from core's cron api: https://github.com/WordPress/WordPress/blob/6b06c73cd7b8af0e0b92a19b0bad370a5747750c/wp-includes/cron.php#L267-L282, except keeping the api the same and ensuring the filters are going to return under the same contract that is currently in place. Also unlike core, we don't need the `pre_` filters under all the public methods because AS already provides a mean for registering a custom data store.

### Changelog

> Dev - Added a set of action hooks that can be used to short circuit the scheduling process.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -98,6 +98,22 @@ function as_schedule_recurring_action( $timestamp, $interval_in_seconds, $hook, 
 		return 0;
 	}
 
+	/**
+	 * Provides an opportunity to short-circuit the default process for enqueuing recurring
+	 * actions.
+	 *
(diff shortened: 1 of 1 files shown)


barryhughes (approved):
Thanks!

Once checks pass, we can merge :+1:

PR #902, opened 2023-01-18
Include PHP 8.2 in the test matrix.

barryhughes:
PHP 8.2 was released in December 2022. This change adds it to the test matrix so we can more readily detect problems when Action Scheduler is running against that version.


rrennick:
@barryhughes Maybe we should update the WP version test matrix as well.

barryhughes:
Definitely. I haven't had space to revisit this so far, but kept it open as a reminder to take action (probably should have created an issue) 🤷🏼

PR #904, opened 2023-01-31
Do not process an action if it cannot be set to `in-progress`

### LINKED ISSUE #896: Not safe to process an action if the "in-progress" status update fails
The `log_execution()` db store method doesn't throw an error if the update fails: https://github.com/woocommerce/action-scheduler/blob/trunk/classes/data-stores/ActionScheduler_DBStore.php#L989-L996

Logic inside of `process_action()` and throughout the plugin really depend on actions having their status updated to `in-progress`. Otherwise, some of the checks and balances fall apart and it's possible for the same action to be processed twice if you are running an action whilst its status is still `pending`.

One of the safest ways to ensure concurrency is working as intended would be to run a query like:

 ```
 UPDATE wp_actionscheduler_actions SET status = 'in-progress' WHERE action_id = 123 AND status = 'pending':
 ```
 
 If it passes, your thread has won any sort of race and is the sole owner of this action. If it fails though, the thread is not safe to proceed because another may have won.
 
 ___
 
 As an aside, it may be smart to retry the query if it failed due to a deadlock here:
 
 ```
$result = $wpdb->query( $sql );
if ( false !== strpos( $wpdb->last_error, 'Deadlock found' ) ) {
  // Try query again. 
}
```

### PR DESCRIPTION
**Pull request by barryhughes**
Immediately before we begin to process an action, we try to change its status to running ('in-progress') [[1]](https://github.com/woocommerce/action-scheduler/blob/3.5.4/classes/abstracts/ActionScheduler_Abstract_QueueRunner.php#L64) [[2]](https://github.com/woocommerce/action-scheduler/blob/3.5.4/classes/data-stores/ActionScheduler_DBStore.php#L1012-L1019). However, we do not check that the status update succeeded.

This is potentially risky, as identified in the linked issue. In this change we add a check to ensure the status update was successful, throwing an exception if it is unsuccessful (this exception is handled by existing logic [[3]](https://github.com/woocommerce/action-scheduler/blob/3.5.4/classes/data-stores/ActionScheduler_DBStore.php#L1012-L1019) which will then mark the action as having failed).

Closes #896.

### Changelog

> Fix - Guard against problems when an action cannot be set to 'in-progress'.

### Testing instructions

We can optionally test to ensure that, under 'normal conditions', things still work as previously (ie, an action is still successfully set to 'in-progress' when it is run). Start by adding the following snippet to a suitable location, such as `wp-content/mu-plugins/pr-904.php`:

```php
<?php

	if ( as_has_scheduled_action( 'test_pr_904' ) ) {
		return;
	}

	as_schedule_single_action( time() + HOUR_IN_SECONDS, 'test_pr_904' );
} );

add_action( 'test_pr_904', function () {
	sleep( 20 );
} );
```

Now visit **Tools ▸ Scheduled Actions ▸ Pending** and you should see our test action (if you have a *lot* of pending actions, you may need to perform a search for `test_or_904` to surface it):

pending

Run it manually (ie, hover over it and click on the run link) then, *as quickly as you can,* open **Tools ▸ Scheduled Actions** in a new tab. There should be an `In-Progress` tab/link and if you open this you should see the test action is running:

running

After another 20 seconds have elapsed, that action should non longer be running and will be marked as complete.

☝🏽 To clarify, the goal of these testing instructions is imply to ensure normal operations are not impeded.

### DIFF
--- a/classes/data-stores/ActionScheduler_DBStore.php
+++ b/classes/data-stores/ActionScheduler_DBStore.php
@@ -1019,7 +1019,9 @@ AND `group_id` = %d
 		$sql = $wpdb->prepare( $sql, self::STATUS_RUNNING, current_time( 'mysql', true ), current_time( 'mysql' ), $action_id ); // phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared
 
 		// phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared
-		if ( ! $wpdb->query( $sql ) ) {
+		$status_updated = $wpdb->query( $sql );
+
+		if ( ! $status_updated ) {
 			throw new Exception(
 				sprintf(
 					/* translators: 1: action ID. 2: status slug. */
--- a/classes/data-stores/ActionScheduler_DBStore.php
+++ b/classes/data-stores/ActionScheduler_DBStore.php
@@ -1005,6 +1005,8 @@ AND `group_id` = %d
 	/**
 	 * Add execution message to action log.
 	 *
+	 * @throws Exception If the action status cannot be updated to self::STATUS_RUNNING ('in-progress').
+	 *
 	 * @param int $action_id Action ID.
 	 *
 	 * @return void
@@ -1015,7 +1017,18 @@ AND `group_id` = %d
 
 		$sql = "UPDATE {$wpdb->actionscheduler_actions} SET attempts = attempts+1, status=%s, last_attempt_gmt = %s, last_attempt_local = %s WHERE action_id = %d";
 		$sql = $wpdb->prepare( $sql, self::STATUS_RUNNING, current_time( 'mysql', true ), current_time( 'mysql' ), $action_id ); // phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared
-		$wpdb->query( $sql ); // phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared
+
+		// phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared
+		if ( ! $wpdb->query( $sql ) ) {
+			throw new Exception(
+				sprintf(
+					/* translators: 1: action ID. 2: status slug. */
+					__( 'Unable to update the status of action %1$d to %2$s.', 'action-scheduler' ),
+					$action_id,
+					self::STATUS_RUNNING
+				)
+			);
+		}
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


classes/data-stores/ActionScheduler_DBStore.php
```diff
@@ -1015,7 +1017,18 @@ public function log_execution( $action_id ) {
 
 		$sql = "UPDATE {$wpdb->actionscheduler_actions} SET attempts = attempts+1, status=%s, last_attempt_gmt = %s, last_attempt_local = %s WHERE action_id = %d";
 		$sql = $wpdb->prepare( $sql, self::STATUS_RUNNING, current_time( 'mysql', true ), current_time( 'mysql' ), $action_id ); // phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared
-		$wpdb->query( $sql ); // phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared
+
+		// phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared
+		if ( ! $wpdb->query( $sql ) ) {
```
coreymckrill: More of a personal style thing, I suppose, but I usually try to avoid executing logic within a conditional's expression, as it makes it harder for me to find where something happened when scanning the code. 🤷

coreymckrill (approved):
👍 Works! In addition to the code in the testing instructions, I added a snippet to make the query fail:

```php
add_filter( 'query', function( $query ) {
	if ( false !== strpos( $query, 'SET attempts = attempts+1, status=' ) ) {
		return '';
	}

	return $query;
} );
```

When I then manually ran the `test_pr_904` action, I got the desired error log:

> action failed via Admin List Table: Unable to update the status of action 7 to in-progress.

Unit test also passes for me.

classes/data-stores/ActionScheduler_DBStore.php
barryhughes: Thanks, @coreymckrill: more than happy to update it—to confirm, is this what you'd prefer to see?

```php
$status_updated = $wpdb->query( $sql );

if ( ! $status_updated ) {
```

coreymckrill: Yeah, I think that's easier to follow 👍

coreymckrill (approved):
@barryhughes this is still a "Draft". Is there anything left to do on this before merging?

barryhughes:
My fault ... I only meant to keep it as a draft until CI passed. All good, thanks for the review!

PR #905, opened 2023-02-06
add CLI exclude-group filter

### LINKED ISSUE #736: Option to ignore some groups when using the wp-cli command

rrennick:
Closes #736 

This PR is an alternative to #784. #784 is a great solution if adding group exclusion were the only filter we were looking to add to the queue runners. However, we are considering adding:
- hook exclusion
- priorities
- retries
- settings screen for managing the async and cron runners

#784 works around the limitation that we cannot modify the `claim_actions` function definition by embedding flags in existing parameters. While that works for one flag/parameter it isn't practical for adding several. This PR avoid that by adding a new function to the base tables DB store.

It also lays the groundwork for adding additional filters. For example, excluding hooks will only require a few additional lines of code. Secondly, all added filters could be implemented on the async and cron runners through a settings screen without using similar parameter tampering.

### Testing

- Disable the default runners.
- I used this hook to create test actions:

```
    $groups = [
      'automatewoo',
      'woocommerce',
      'woocommerce-admin',
      'internal',
    ];
    $hooks = [
      'alpha',
      'beta',
      'gamma',
      'delta',
    ];
    $group = rand( 0, 3 );
    $hook = rand( 0, 3 );
    $unique = rand( 1, 9 );

    as_schedule_single_action(
        time() + MINUTE_IN_SECONDS,
        'init_' . $hooks[ $hook ] . $unique,
        [],
        'test-' . $groups[ $group ]
    );
}, 99 );
```
- Allow some time for actions to be created
- Then ran the following a few minutes apart
```
wp action-scheduler run --group=alpha
// check that all pending past due alpha actions ran
wp action-scheduler run --exclude-group=beta
// check that beta is the only group with pending past due actions
```

---

### Changelog

> Add - Add support for excluding group(s) when using the WP CLI queue runner.

### DIFF
--- a/classes/WP_CLI/ActionScheduler_WPCLI_Scheduler_command.php
+++ b/classes/WP_CLI/ActionScheduler_WPCLI_Scheduler_command.php
@@ -56,7 +56,7 @@ class ActionScheduler_WPCLI_Scheduler_command extends WP_CLI_Command {
 	 * : Only run actions from the specified group. Omitting this option runs actions from all groups.
 	 *
 	 * [--exclude-group=<group>]
-	 * : Run actions from all groups except the specified group.
+	 * : Run actions from all groups except the specified group. This option is ignored when `--group` is used.
 	 *
 	 * [--free-memory-on=<count>]
 	 * : The number of actions to process between freeing memory. 0 disables freeing memory. Default 50.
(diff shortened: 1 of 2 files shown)


Konamiman:
There's something weird with the report of deleted actions:

image

If I specify 2000 it says "2 actions deleted". For numbers below 1000 it reports the correct number.

Konamiman:
By looking at the implementation it seems that when both `--group` and `--exclude-group` are specified, only `--group` is taken in account. That's fine (I guess combining both is out of scope for this pr) but maybe it's something that should be explicitly stated in the help.

rrennick:
@Konamiman Thanks for the suggestion. I've added that to the option description.

Konamiman (approved):

PR #915, opened 2023-02-24
Improve logic for flushing the runtime cache.

### LINKED ISSUE #909: Consider protections around use of `wp_cache_flush_runtime()`
Original report: https://wordpress.org/support/topic/wp_cache_flush_runtime-incorrectly-called/

> Function wp_cache_flush_runtime was called **incorrectly**. Your object cache implementation does not support flushing the in-memory runtime cache.

Before calling `wp_cache_flush_runtime()` (which we currently do [over here](https://github.com/woocommerce/action-scheduler/blob/3.5.4/classes/ActionScheduler_QueueRunner.php#L191)) we should additionally check if:

- Function `wp_cache_supports` is defined (noting it is only available as of WP 6.1).
- If it is defined, call it and check if `flush_runtime` is supported by the current caching implementation.

Context:

- `wp_cache_flush_runtime()` was introduced in WP 6.0.
- Consequently, we started using the new function as of Action Scheduler 3.4.2 (testing to make sure it exists, first of all).
- `wp_cache_supports()` was then introduced in WP 6.1 and can be used to test for various caching features, such as runtime caching ... however, we missed this change and, consequently, there is potential for a 'doing it wrong' notice to be generated.

barryhughes:
Before flushing the runtime cache and depending on the active version of WordPress, we should:

- Check if support for flushing the runtime cache is explicitly declared (possible from WP 6.1 onwards).
- Otherwise, check if the `wp_cache_flush_runtime()` function exists (possible with all versions of WP).

Calling `wp_cache_flush_runtime()` is essentially safe so long as we are using WP 6.0 or up, but with WP 6.1 and above it will potentially lead to a doing-it-wrong notice if the active cache does not support the concept (even though nothing will break).

Closes #909.

### Testing

This code will not be invoked when using the WP CLI queue runner, and there isn't a super easy 'user test' you can apply. Code review should be sufficient in this case.

### Changelog

> Fix - Avoid throwing errors when clearing the runtime cache with WP 6.1 and higher.

### DIFF
--- a/classes/ActionScheduler_QueueRunner.php
+++ b/classes/ActionScheduler_QueueRunner.php
@@ -186,8 +186,14 @@ class ActionScheduler_QueueRunner extends ActionScheduler_Abstract_QueueRunner {
 		/*
 		 * Calling wp_cache_flush_runtime() lets us clear the runtime cache without invalidating the external object
 		 * cache, so we will always prefer this when it is available (but it was only introduced in WordPress 6.0).
+		 *
+		 * Additionally, the preferred way of detecting support changed in WordPress 6.1 so we use two different
+		 * methods here.
 		 */
-		if ( function_exists( 'wp_cache_flush_runtime' ) ) {
+		$flushing_runtime_cache_explicitly_supported = function_exists( 'wp_cache_supports' ) && wp_cache_supports( 'flush_runtime' );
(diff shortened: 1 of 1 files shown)


coreymckrill (approved):
👍

PR #916, opened 2023-02-24
exclude past-due from list table all filter count

### LINKED ISSUE #913: List table All count incorrect on past-due actions
The list table is adding the past-due action count to the action total. Past-due actions are already counted in the Pending count. 

Screenshot 2023-02-23 at 2 51 31 PM

rrennick:
Closes #913 

This PR fixes the list table `All` action count by deducting the past due count which is already included in the pending count.

### Testing

- Disable the default runner
- Create an async action
- Allow some time for it to become past due
- Go to the `All` filter in the list table
- Verify that the `All` count  on the filter matches the paging count on the right.

---

### Changelog

> Fix - Corrects the `All` action count.


jorgeatorres (approved):
LGTM. Thank you @rrennick!

PR #920, opened 2023-03-01
Filter view labels (status names) should be translatable | #919

### LINKED ISSUE #919: Name of statuses in filters does not display translated
In the list table, the name of statuses in filters does not display translated, as does the "All" element:

action-scheduler

I think the list is generated by this line:

https://github.com/woocommerce/action-scheduler/blob/833f3bde9dac71d78559a756670fd329151fc47b/classes/abstracts/ActionScheduler_Abstract_ListTable.php#L693

Thanks.

barryhughes:
The status/filter view names found in the **Tools ▸ Scheduled Actions** screen should be translatable:

filter-status-labels

Currently, as highlighted in the linked issue, they are not—what we are actually doing is echoing the status *slugs*, after passing them through [ucfirst()](https://www.php.net/manual/en/function.ucfirst.php), rather than using the translated status labels. 

Closes #919.

### Testing instructions

Start by adding the following snippet to a suitable location. I recommend `wp-content/mu-plugins/test-919.php` (which you can remove when you have finished testing):

```php
<?php

function test_fix_for_919( $translation, $text ) {
	$status_labels = array( 'All', 'Complete', 'Pending', 'In-progress', 'Failed', 'Canceled', 'Past-due' );
	return in_array( $text, $status_labels ) ? "Translated '$text'" : $translation;
}

add_filter( 'gettext_action-scheduler', 'test_fix_for_919', 10, 2 );
add_filter( 'gettext_with_context_action-scheduler', 'test_fix_for_919', 10, 2 );
```

With that in place, if you visit the **Tools ▸ Scheduled Actions** screen:

- Without this change, there will be no change to the status labels (they will render in English).
- With this change, each status name should render as `Translated '<slug>'` (simulating a situation where we have translations for these strings).

filter-status-labels-translated

For clarity, this should only impact the 'filter views' as in the above screenshot. I don't know if we surface the status slugs elsewhere, but if we do they are probably not covered by this change.

### Changelog

> Fix - Makes the filter names in the `Tools > Scheduled Actions` screen translatable.

### DIFF
--- a/classes/abstracts/ActionScheduler_Abstract_ListTable.php
+++ b/classes/abstracts/ActionScheduler_Abstract_ListTable.php
@@ -680,21 +680,27 @@ abstract class ActionScheduler_Abstract_ListTable extends WP_List_Table {
 			$this->status_counts = array( 'all' => $all_count ) + $this->status_counts;
 		}
 
-		foreach ( $this->status_counts as $status_name => $count ) {
+		// Translated status labels.
+		$status_labels             = ActionScheduler_Store::instance()->get_status_labels();
+		$status_labels['all']      = _x( 'All', 'status labels', 'action-scheduler' );
+		$status_labels['past-due'] = _x( 'Past-due', 'status labels', 'action-scheduler' );
+
+		foreach ( $this->status_counts as $status_slug => $count ) {
 
 			if ( 0 === $count ) {
 				continue;
 			}
 
-			if ( $status_name === $request_status || ( empty( $request_status ) && 'all' === $status_name ) ) {
+			if ( $status_slug === $request_status || ( empty( $request_status ) && 'all' === $status_slug ) ) {
 				$status_list_item = '<li class="%1$s"><a href="%2$s" class="current">%3$s</a> (%4$d)</li>';
 			} else {
 				$status_list_item = '<li class="%1$s"><a href="%2$s">%3$s</a> (%4$d)</li>';
 			}
 
-			$status_filter_url   = ( 'all' === $status_name ) ? remove_query_arg( 'status' ) : add_query_arg( 'status', $status_name );
+			$status_name         = isset( $status_labels[ $status_slug ] ) ? $status_labels[ $status_slug ] : ucfirst( $status_slug );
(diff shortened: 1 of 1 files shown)


classes/abstracts/ActionScheduler_Abstract_ListTable.php
```diff
@@ -680,21 +680,27 @@ protected function display_filter_by_status() {
 			$this->status_counts = array( 'all' => $all_count ) + $this->status_counts;
 		}
 
-		foreach ( $this->status_counts as $status_name => $count ) {
+		// Translated status labels.
+		$status_labels             = ActionScheduler_Store::instance()->get_status_labels();
+		$status_labels['all']      = _x( 'All', 'status labels', 'action-scheduler' );
+		$status_labels['past-due'] = _x( 'Past-due', 'status labels', 'action-scheduler' );
```
coreymckrill, line 686: 🤔 I get why the "All" label would be added here, but why wouldn't "Past-due" get added to the `get_status_labels()` method?

coreymckrill (approved):
👍 Works! Not merging yet in case you want to follow up on the question below.

classes/abstracts/ActionScheduler_Abstract_ListTable.php
barryhughes: I don't think we've defined a vocabulary for this, but both could possibly be said to be "meta-statuses" in the sense that neither can be found in the actions table (past-due actions are just a subset of those with pending status).

That's not necessarily a reason to exclude it from the `get_status_labels()` method, I was just slightly hesitant because the other statuses that method returns are 'concrete' statuses, each with a matching `ActionScheduler_Store::STATUS_*` constant. 

So ... I can't think of problems adding it would cause, but I wanted to err on the side of caution. Open to making a change, though.

coreymckrill: Ah ok, I didn't realize Past-due was also a "meta" status. I guess it probably makes sense to leave it out then, in case there's some other usage of `get_status_labels` that assumes they're all things you would save to the database or something.

PR #921, opened 2023-03-03
Add action priorities.

### LINKED ISSUE #910: Implement priorities
Implement action priorities. The core idea is that the order in which (over-)due scheduled actions will be processed will be determined by priority first, time-waiting second.

- Prior discussion: https://github.com/woocommerce/action-scheduler/issues/236
- Internal links: pcJTIi-20-p2 | p7bje6-3ms-p2

To do:

- [ ] Final decision on whether priorities should apply to groups, atomic actions or both

barryhughes:
Adds support for priorities to scheduled actions. 

- Given a set of scheduled actions that are due (or past-due) for processing, higher priority actions will be claimed and processed first.
- This gives library consumers the latitude to designate certain actions (an example could be those used for processing renewal payments) priority over others (those used to perform routine clean-up tasks, perhaps).
- Scheduled action priorities mimic regular WordPress action priorities in the sense the lower values such as 1 are a higher priority than large values such as 100.
    - On the other hand, they diverge from WordPress action priorities in that they must be integers.
    - They also need to be in the range 0-255. This seems sufficient for our needs, and means there is very little overhead.

Updates #910.

---

### Testing instructions

Please see the new and revised unit tests, though we can also test manually as described here. However, before you begin, I recommend disabling all other plugins and wiping the queue clean:

- If you are using a system scheduler (like `cron.d`) to trigger either WP Cron or Action Scheduler directly, then disable the relevant rules.
- Disable WP Cron by adding `define( 'DISABLE_WP_CRON', true );` to your wp-config.php file.
- Delete the existing queue: `wp db query "DELETE FROM $(wp db prefix)actionscheduler_actions"`
- Add the following code to a suitable location. I recommend `wp-content/mu-plugins/test-pr-921.php`:

```php
<?php

add_filter( 'action_scheduler_allow_async_request_runner', '__return_false' );
add_action( 'test_action', function( $arg ) { echo $arg . PHP_EOL; } );
```

Now create some past-due actions of mixed priority (all these timestamps are in the past—the early 2000s in fact):

```sh
wp eval "as_schedule_single_action( 1000000000, 'test_action', [ 'DEFAULT PRIORITY' ] );" \
wp eval "as_schedule_single_action( 1000000100, 'test_action', [ 'HIGH PRIORITY' ], '', false, 1 );" \
wp eval "as_schedule_single_action( 1000010000, 'test_action', [ 'LOW PRIORITY' ], '', false, 100 );" \
wp eval "as_schedule_single_action( 1010000000, 'test_action', [ 'DEFAULT PRIORITY' ] );"
```

Now run the queue using `wp action-scheduler run`. You should see output looking something like the following:

```
Found 4 scheduled tasks
Started processing action 2772
HIGH PRIORITY
Completed processing action 2772 with hook: test_action
Running 4 actions  25 % [=====================>                                                                 ] 0:00 / 0:00Started processing action 2773
DEFAULT PRIORITY
Completed processing action 2773 with hook: test_action
Started processing action 2774
DEFAULT PRIORITY
Completed processing action 2774 with hook: test_action
Started processing action 2775
LOW PRIORITY
Completed processing action 2775 with hook: test_action
Running 4 actions  100% [=======================================================================================] 0:00 / 0:00
1 batch executed.
Success: 4 scheduled tasks completed.
```

Specifically, what we want to see is that the priority labels are ordered just as you see above:

- `HIGH PRIORITY` processed first
- `DEFAULT PRIORITY` (x2) processed second
- `LOW PRIORITY` processed last

---

### Changelog

> Add - Priorities can now be set for scheduled actions.

### DIFF
--- a/classes/actions/ActionScheduler_Action.php
+++ b/classes/actions/ActionScheduler_Action.php
@@ -10,6 +10,19 @@ class ActionScheduler_Action {
 	protected $schedule = NULL;
 	protected $group = '';
 
+	/**
+	 * Priorities are conceptually similar to those used for regular WordPress actions.
+	 * Like those, a lower priority takes precedence over a higher priority and the default
+	 * is 10.
+	 *
+	 * Unlike regular WordPress actions, the priority of a scheduled action is strictly an
+	 * integer and should be kept within the bounds 0-255 (anything outside the bounds will
+	 * be brought back into the acceptable range).
+	 *
+	 * @var int
+	 */
+	protected $priority = 10;
+
 	public function __construct( $hook, array $args = array(), ActionScheduler_Schedule $schedule = NULL, $group = '' ) {
 		$schedule = empty( $schedule ) ? new ActionScheduler_NullSchedule() : $schedule;
 		$this->set_hook($hook);
@@ -92,5 +105,31 @@ class ActionScheduler_Action {
 	 */
 	public function is_finished() {
 		return FALSE;
+	}
+
+	/**
+	 * Sets the priority of the action.
+	 *
+	 * @param int $priority Priority level (lower is higher priority). Should be in the range 0-255.
+	 *
+	 * @return void
+	 */
+	public function set_priority( $priority ) {
+		if ( $priority < 0 ) {
+			$priority = 0;
+		} elseif ( $priority > 255 ) {
+			$priority = 255;
+		}
+
+		$this->priority = $priority;
+	}
+
+	/**
+	 * Gets the action priority.
+	 *
+	 * @return int
+	 */
+	public function get_priority() {
+		return $this->priority;
 	}
 }
--- a/classes/actions/ActionScheduler_Action.php
+++ b/classes/actions/ActionScheduler_Action.php
@@ -121,7 +121,7 @@ class ActionScheduler_Action {
 			$priority = 255;
 		}
 
-		$this->priority = $priority;
+		$this->priority = (int) $priority;
 	}
 
 	/**
--- a/classes/data-stores/ActionScheduler_DBStore.php
+++ b/classes/data-stores/ActionScheduler_DBStore.php
@@ -92,6 +92,7 @@ class ActionScheduler_DBStore extends ActionScheduler_Store {
 				'scheduled_date_local' => $this->get_scheduled_date_string_local( $action, $date ),
 				'schedule'             => serialize( $action->get_schedule() ), // phpcs:ignore WordPress.PHP.DiscouragedPHPFunctions.serialize_serialize
 				'group_id'             => $this->get_group_id( $action->get_group() ),
+				'priority'             => $action->get_priority(),
 			);
 
 			$args = wp_json_encode( $action->get_args() );
@@ -362,7 +363,7 @@ AND `group_id` = %d
 		}
 		$group = $data->group ? $data->group : '';
 
-		return ActionScheduler::factory()->get_stored_action( $data->status, $data->hook, $args, $schedule, $group );
+		return ActionScheduler::factory()->get_stored_action( $data->status, $data->hook, $args, $schedule, $group, $data->priority );
 	}
 
 	/**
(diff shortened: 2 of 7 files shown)


classes/actions/ActionScheduler_Action.php
```diff
@@ -93,4 +106,30 @@ public function get_group() {
...
+	 * @return void
+	 */
+	public function set_priority( $priority ) {
+		if ( $priority < 0 ) {
+			$priority = 0;
+		} elseif ( $priority > 255 ) {
+			$priority = 255;
+		}
+
+		$this->priority = $priority;
```
vedanshujain: probably cast to `int` here as well

classes/data-stores/ActionScheduler_DBStore.php
```diff
@@ -928,7 +929,7 @@ protected function claim_actions( $claim_id, $limit, \DateTime $before_date = nu
 		 *
 		 * @param string $order_by_sql
 		 */
-		$order    = apply_filters( 'action_scheduler_claim_actions_order_by', 'ORDER BY attempts ASC, scheduled_date_gmt ASC, action_id ASC' );
+		$order    = apply_filters( 'action_scheduler_claim_actions_order_by', 'ORDER BY priority ASC, attempts ASC, scheduled_date_gmt ASC, action_id ASC' );
```
vedanshujain, line 932: not sure about this query, this query already causes bunch of issues for high traffic sites, adding another order clause will probably worsen it. any ideas around this?

vedanshujain: perhaps we can use tools like `mysqlslap` to simulate deadlock conditions and see what can we do to improve these

vedanshujain (commented):
Left couple of comments, otherwise looks good.

classes/actions/ActionScheduler_Action.php
barryhughes: Good catch, addressed in https://github.com/woocommerce/action-scheduler/pull/921/commits/fbffde233c0157215d47370ed89c01035a47b4e5.

classes/data-stores/ActionScheduler_DBStore.php
barryhughes: Looking at this (benchmarking, also alternatives).

Though I think it's worth noting that while this introduces additional overhead of its own, its potential value only becomes apparent in those situations where the system is under pressure (because when things are running smoothly, priorities will not be needed at all).

barryhughes: Hi @vedanshujain. Coming back to this one...

It doesn't seem to have a *huge* impact. These are averaged results from multiple test runs, showing the number of successfully processed actions (and, in the second chart, the occurrence of deadlocks) alongside the number of concurrent queue runners:

- Blue line shows how 3.5.4 performs with the current schema.
- Red line shows how this branch performs (actions all share the same priority).
- Yellow line shows how this branch performs (mixed priorities: 10% high, 10% low, 80% default).

<div align="center">Concurrent Queue Runners vs Throughput</div>

<div align="center">Concurrent Queue Runners vs Deadlocks]</div>

It does look like 3.5.4 has a slight edge in some cases, but this wasn't consistently the case across all test runs.

> This query already causes bunch of issues for high traffic sites, adding another order clause will probably worsen it. Any ideas around this?

A few ideas:

- Performance is not conclusively worse. Added to which, this feature offers a potential advantage for those situations where the queue gets backed up (because, of course, it will allow higher priority actions to move forward):
    - We could merge as is, release and then monitor for community feedback.
    - An existing filter hook is in place, making it relatively trivial to roll back the additional `orderby` field.
- It's possible we could test internally on a large operating WooCommerce site, and gather additional real-world metrics via that process.
- Alternatively, or additionally, we could look at the rest of the `orderby` clause:
    - We currently look at 3 fields (4 with this change). Can some be dropped?
    - I wonder in particular if `attempts` is generally useful in real-world cases.
    - Experimenting by dropping `action_id` could similarly be interesting.

vedanshujain (approved):
LGTM! Great work!

PR #922, opened 2023-03-05
Add $unique parameter to function signatures

siliconforks:
In commit c5211c2c19be1e8bfe48e909ccffa14c93852294 documentation for the $unique parameter was added for several functions; however, the parameter was not actually added to the function signatures.


barryhughes (approved):
Thanks for catching this!

PR #923, opened 2023-03-06
Add a cast-to-int for extra safety before forming new DateTime object.

### LINKED ISSUE #917: PHP 8 Failing DateTime
Thanks for the work on this.

I am running PHP 8 
AS Plugin: 3.5.4 version

I am getting the Failing DateTime. The issue seems to be with `/www/com_643/public/wp-content/plugins/action-scheduler/functions.php(421): DateTime->__construct(' seconds ago', `. We might need to have the time as an integer.
or somethig else.

```
2023/02/27 17:46:34 [error] 52135#52135: *17723 FastCGI sent in stderr: "PHP message: PHP Fatal error: Uncaught Exception: DateTime::__construct(): Failed to parse time string ( seconds ago) at position 0 (s): The timezone could not be found in the database in /www/com_643/public/wp-content/plugins/action-scheduler/functions.php:421
Stack trace:
#0 /www/com_643/public/wp-content/plugins/action-scheduler/functions.php(421): DateTime->__construct(' seconds ago', Object(DateTimeZone))
#1 /www/com_643/public/wp-content/plugins/action-scheduler/classes/ActionScheduler_QueueCleaner.php(34): as_get_datetime_object(' seconds ago')
#2 /www/com_643/public/wp-content/plugins/action-scheduler/classes/ActionScheduler_QueueCleaner.php(139): ActionScheduler_QueueCleaner->delete_old_actions()
#3 /www/com_643/public/wp-content/plugins/action-scheduler/classes/abstracts/ActionScheduler_Abstract_QueueRunner.php(160): ActionScheduler_QueueCleaner->clean(300)
```

barryhughes:
Adds a small piece of defensive coding for the `action_scheduler_retention_period` hook. 

We use this hook to allow an integer value (representing a number of seconds) to be modified by other plugins, but it is possible they will return a non-integer and, in some cases—such as if `null` or `(string) ''` are returned—they will trigger a fatal error.

Closes https://github.com/woocommerce/action-scheduler/issues/917.

### Testing

Try this multi-liner:

```sh
wp eval "\
  add_filter( 'action_scheduler_retention_period', '__return_false' ); \
  ( new ActionScheduler_QueueRunner )->run(); \
"
```

With this branch, there should be no problem. Without this branch, you should find a fatal error similar to the following is reported:

```
Error: There has been a critical error on this website.Learn more about troubleshooting WordPress.Error details:Array
(
    [type] => 1
    [message] => Uncaught Exception: DateTime::__construct(): Failed to parse time string ( seconds ago) at position 0 (s)
    ...
)
```

### Changelog

> Fix - Improves safety around the `action_scheduler_retention_period` filter hook.

### DIFF
--- a/classes/ActionScheduler_QueueCleaner.php
+++ b/classes/ActionScheduler_QueueCleaner.php
@@ -49,7 +49,23 @@ class ActionScheduler_QueueCleaner {
 		 * @param int $retention_period Minimum scheduled age in seconds of the actions to be deleted.
 		 */
 		$lifespan = (int) apply_filters( 'action_scheduler_retention_period', $this->month_in_seconds );
-		$cutoff = as_get_datetime_object( $lifespan . ' seconds ago' );
+
+		try {
+			$cutoff = as_get_datetime_object( $lifespan . ' seconds ago' );
+		} catch ( Exception $e ) {
+			_doing_it_wrong(
+				__METHOD__,
+				sprintf(
+					/* Translators: %s is the exception message. */
+					esc_html__( 'It was not possible to determine a valid cut-off time: %s.', 'action-scheduler' ),
+					esc_html( $e->getMessage() )
+				),
+				'3.5.5'
+			);
+
+			return array();
+		}
+
 
 		/**
 		 * Filter the statuses when cleaning the queue.
--- a/classes/ActionScheduler_QueueCleaner.php
+++ b/classes/ActionScheduler_QueueCleaner.php
@@ -40,7 +40,7 @@ class ActionScheduler_QueueCleaner {
 	/**
 	 * Default queue cleaner process used by queue runner.
 	 *
-	 * @return void
+	 * @return array
 	 */
 	public function delete_old_actions() {
 		/**
@@ -58,7 +58,7 @@ class ActionScheduler_QueueCleaner {
 		 */
 		$statuses_to_purge = (array) apply_filters( 'action_scheduler_default_cleaner_statuses', $this->default_statuses_to_purge );
 
-		$this->clean_actions( $statuses_to_purge, $cutoff, $this->get_batch_size() );
+		return $this->clean_actions( $statuses_to_purge, $cutoff, $this->get_batch_size() );
 	}
 
 	/**
--- a/classes/ActionScheduler_QueueCleaner.php
+++ b/classes/ActionScheduler_QueueCleaner.php
@@ -48,7 +48,7 @@ class ActionScheduler_QueueCleaner {
 		 *
 		 * @param int $retention_period Minimum scheduled age in seconds of the actions to be deleted.
 		 */
-		$lifespan = (int) apply_filters( 'action_scheduler_retention_period', $this->month_in_seconds );
+		$lifespan = apply_filters( 'action_scheduler_retention_period', $this->month_in_seconds );
 
 		try {
 			$cutoff = as_get_datetime_object( $lifespan . ' seconds ago' );
--- a/classes/ActionScheduler_QueueCleaner.php
+++ b/classes/ActionScheduler_QueueCleaner.php
@@ -48,7 +48,7 @@ class ActionScheduler_QueueCleaner {
 		 *
 		 * @param int $retention_period Minimum scheduled age in seconds of the actions to be deleted.
 		 */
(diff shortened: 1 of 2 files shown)


coreymckrill (commented):
👍 Works via command line testing.

Worth adding a unit test for this? Maybe [here](https://github.com/woocommerce/action-scheduler/blob/8bf7ceba4bbb3af1b0987b6c184c6d89710e1d15/tests/phpunit/runner/ActionScheduler_QueueCleaner_Test.php).

barryhughes:
Hi @coreymckrill,

> Worth adding a unit test for this? Maybe [here](https://github.com/woocommerce/action-scheduler/blob/8bf7ceba4bbb3af1b0987b6c184c6d89710e1d15/tests/phpunit/runner/ActionScheduler_QueueCleaner_Test.php).

Yep, I didn't because, honestly, I was struggling to think of a graceful way of doing this:

```php
add_filter( 'action_scheduler_retention_period', '__return_null' );
$cleaner = new ActionScheduler_QueueCleaner( $store );

# This is where the failure would happen; but it returns void so we
# cannot currently test the return value:
$cleaner->delete_old_actions();

# Simply asserting for the sake of it seems inelegant and bad form,
# though maybe I'm overthinking it/maybe we don't *need* the assertion:
$this->assertTrue( true, 'A queue cleaner will not fail even if a non-integer value like null is supplied as the retention period.');

remove_filter( 'action_scheduler_retention_period', '__return_null' );
```

A possibility is we could change `$cleaner->delete_old_actions()` to return the result (an array) of underlying call to the `clean_actions()` method, and since in this case we don't actually need to change the method signature to do this it would be backwards-compatible: then we can at least assert that an array was returned.

What do you think?

coreymckrill:
@barryhughes 

> A possibility is we could change $cleaner->delete_old_actions() to return the result (an array) of underlying call to the clean_actions() method

It seems like it would be fine to do this, and would just generally make the method more testable.

One other possible approach might be to do a try/catch within the unit test:

```php
add_filter( 'action_scheduler_retention_period', '__return_null' );
$cleaner = new ActionScheduler_QueueCleaner( $store );

try {
	$cleaner->delete_old_actions();
	$this->assertTrue( true );
} catch ( \Exception $e ) {
	$this->assertTrue( false );
}

remove_filter( 'action_scheduler_retention_period', '__return_null' );
```

Although maybe this is still a little bit hacky?

barryhughes:
Perhaps we've found a real world case for `doNotExpectException()` 🧐 (jk) ... I'll make a change and add _something_ to cover this.

barryhughes:
Sorry, hitting pause: I think this can be improved some more.

classes/ActionScheduler_QueueCleaner.php
```diff
@@ -40,16 +40,32 @@ public function __construct( ActionScheduler_Store $store = null, $batch_size =
...
 		 * @param int $retention_period Minimum scheduled age in seconds of the actions to be deleted.
 		 */
-		$lifespan = apply_filters( 'action_scheduler_retention_period', $this->month_in_seconds );
-		$cutoff = as_get_datetime_object( $lifespan . ' seconds ago' );
+		$lifespan = (int) apply_filters( 'action_scheduler_retention_period', $this->month_in_seconds );
+
+		try {
+			$cutoff = as_get_datetime_object( $lifespan . ' seconds ago' );
+		} catch ( Exception $e ) {
+			_doing_it_wrong(
```
barryhughes, line 56: I added this for completeness, in but didn't cover it in our unit tests because it seems tricky to reliably trigger it across different PHP runtimes (the behavior when casting to int from an unsupported type is [undefined](https://www.php.net/manual/en/language.types.integer.php#language.types.integer.casting)).

barryhughes:
Apologies for the flood of changes 😬 ... to summarize:

- I removed the cast to int.
- This maintains the same level of strictness as previously (so, instead of suddenly allowing `null` as a retention period, which was previously disallowed, and assuming it should be treated as zero, we continue to disallow it).
- However, to prevent fatal errors, we now catch any exception that is raised when creating the `DateTime` and raise attention via `_doing_it_wrong`. 

This should maintain the same strictness, without the site breaking fatal.

coreymckrill (approved):
Looks great now! Doing the try/catch within the method instead of casting seems like a better approach.

PR #924, opened 2023-03-07
Add a hook allowing exceptions for consistently failing recurring actions.

### LINKED ISSUE #908: Notify when task fails and retries are exhausted
I am a user of a plugin that implements a data exchange/sync between WP and an external service. The plugin polls the external service and when it detects a worthy event, it pulls the data from external service over to WP. The plugin uses AS to implement polling and data retrieval by scheduling a series of AS tasks. The recurrence is via WP Cron. 

Occasionally, an AS task generated by the plugin fails to contact the external service. After the task is retried and the max number of retries is reached, the plugin logs this error from AS:

```This action appears to be consistently failing. A new instance will not be scheduled.```

We'd like to be notified when the AS task fails and the retries are exhausted. My attempt:

```php
add_action('action_scheduler_failed_execution', my_failed_action_handler_v1, 10, 3);
add_action('action_scheduler_failed_execution', my_failed_action_handler_v2, 10, 3);

function my_failed_action_handler_v2(const string $action_id, $ex, $context) {
		$action = ActionScheduler::store()->fetch_action($action_id);
		if (
			ActionScheduler_Store::STATUS_FAILED === ActionScheduler::store()->get_status( $action_id )
			&& ActionScheduler::store()->recurring_action_is_consistently_failing( $action, $action_id )
		) 
		{
			 wp_mail(...);
		}
}

function my_failed_action_handler_v1(const string $action_id, $ex, $context ) {
	
    $error_message = $ex->get_error_message();
    if ( strpos( $error_message, 'new instance will not be scheduled' ) !== false ) {
        wp_mail( ... );
    }
}
```

My questions:

1. Is `action_scheduler_failed_execution` the right hook

2. If possible, I would prefer not to rely on string matches (the v1 handler). Is the condition in v2 correct in terms of catching the dead task? I based it on [QueueRunner](https://github.com/woocommerce/action-scheduler/blob/trunk/classes/abstracts/ActionScheduler_Abstract_QueueRunner.php)

I did read the docs and AS source code but I might have missed something. Pointers to docs and/or examples are welcome.

barryhughes:
Since the [3.5.0](https://github.com/woocommerce/action-scheduler/releases/tag/3.5.0) release, if a recurring action consistently fails then it will not be rescheduled.

However, there may be cases where that is undesirable and an exception should be made, or else where plugins simply want to be able to observe the fact that an action has been deemed to be consistently failing. To accomplish these things, this PR introduces a new filter hook:

`action_scheduler_recurring_action_is_consistently_failing`

This can be used both to observe that an action has been determined to be consistently failing, and to make an exemption if appropriate.

Closes #908.

---

### Testing instructions

New assertions/code review are probably sufficient, but we can also mimic the same thing manually. First, to make it easier to test, I recommend deleting all existing actions from the queue before you begin. Via WP CLI:

```sh
wp db query "DELETE FROM $(wp db prefix)actionscheduler_actions"
```

Next, add the following snippet to a suitable location. I recommend `wp-content/mu-plugins/test-908.php`:

```php
<?php

add_filter( 'action_scheduler_recurring_action_is_consistently_failing', function( $is_failing, $action ) {
	return 'foo' === $action->get_hook() ? false : $is_failing;
}, 10, 2 );
```

Now create two recurring actions (for hooks `foo` and `bar`):

```sh
wp eval "as_schedule_recurring_action( time(), 1, 'foo' );" \
wp eval "as_schedule_recurring_action( time(), 1, 'bar' );"
```

Now let's wait for the queue runner to do its thing. You can hurry things along via WP CLI in most shells by using the following:

```sh
for i in `seq 8`; wp action-scheduler run; done
```

That done, if you observe the list of scheduled actions via the **Tools ▸ Scheduled Actions** screen, you should see something close to the following:

foo-bar-threshold

Principally:

* Scheduled actions for `bar` stop being rescheduled after 5 failed attempts.
* Scheduled actions for `foo` continue to be rescheduled (because our filter function is making an exception for these).

### Changelog

> Add - New hook `action_scheduler_recurring_action_is_consistently_failing`.

### DIFF
--- a/classes/abstracts/ActionScheduler_Abstract_QueueRunner.php
+++ b/classes/abstracts/ActionScheduler_Abstract_QueueRunner.php
@@ -143,12 +143,22 @@ abstract class ActionScheduler_Abstract_QueueRunner extends ActionScheduler_Abst
 			return false;
 		}
 
-		// Now let's fetch the first action (having the same hook) of *any status*ithin the same window.
+		// Now let's fetch the first action (having the same hook) of *any status* within the same window.
 		unset( $query_args['status'] );
 		$first_action_id_with_the_same_hook = $this->store->query_actions( $query_args );
 
-		// If the IDs match, then actions for this hook must be consistently failing.
-		return $first_action_id_with_the_same_hook === $first_failing_action_id;
+		/**
+		 * If a recurring action is assessed as consistently failing, it will not be rescheduled. This hook provides a
+		 * way to observe and optionally override that assessment.
+		 *
+		 * @param bool                   $is_consistently_failing If the action is considered to be consistently failing.
+		 * @param ActionScheduler_Action $action                  The action being assessed.
+		 */
+		return (bool) apply_filters(
+			'action_scheduler_recurring_action_is_consistently_failing',
+			$first_action_id_with_the_same_hook === $first_failing_action_id,
+			$action
+		);
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


jorgeatorres (approved):
Looks good to me. Thank you @barryhughes!

PR #926, opened 2023-03-07
Always raise the time limit.

### LINKED ISSUE #767: Make sure you are always "raising the time limit"
In [`ActionScheduler_Compatibility::raise_time_limit()`](https://github.com/woocommerce/action-scheduler/blob/3.2.1/classes/ActionScheduler_Compatibility.php#L81-L98)

I would suggest you always call `set_time_limit()` (unless already infinite) using either the current max_execution_time or the `$limit` which ever is bigger that way you are always "raising the time limit".

_Originally posted by @dgoring in https://github.com/woocommerce/action-scheduler/issues/755#issuecomment-921589673_

barryhughes:
Method [`ActionScheduler_Compatibility::raise_time_limit()`](https://github.com/woocommerce/action-scheduler/blob/3.5.4/classes/ActionScheduler_Compatibility.php#L88-L108) is intended to only ever raise and never lower the time limit. However, in some scenarios a time extension is requested but will be ignored. For example:

- Suppose the max execution time is set to 60 seconds.
- Then, with 10 seconds left to run, [`raise_time_limit()`](https://github.com/woocommerce/action-scheduler/blob/3.5.4/classes/ActionScheduler_Compatibility.php#L88-L108) is called with a value of `20`. This should effectively mean the request can run for a further 20 seconds (giving us 70 seconds of execution time in total).
- As currently written, it will make no adjustment at all in this case because the [max execution time is higher than the new value](https://github.com/woocommerce/action-scheduler/blob/3.5.4/classes/ActionScheduler_Compatibility.php#L98).

The problem here is that the current logic ignores the fact that a successful call to [`set_time_limit()`](https://www.php.net/manual/en/function.set-time-limit.php) actually resets the timer. In this PR, we apply the suggestion from the linked issue to always raise the limit—using either the max execution time or the new limit, whichever is higher.

Closes https://github.com/woocommerce/action-scheduler/issues/767.

---

### Testing instructions

Code review should be sufficient in this case.

### Changelog

> Fix - Always raise and never lower the time limit.

### DIFF
--- a/classes/ActionScheduler_Compatibility.php
+++ b/classes/ActionScheduler_Compatibility.php
@@ -93,12 +93,7 @@ class ActionScheduler_Compatibility {
 			return;
 		}
 
-		/*
-		 * Whichever of $max_execution_time or $limit is higher is the amount by which we raise the time limit.
-		 *
-		 * This prevents a scenario where the time limit is inadvertently lowered (such as if $limit is less than
-		 * $max_execution_time, and where the time left to run exceeds $limit).
-		 */
+		// Whichever of $max_execution_time or $limit is higher is the amount by which we raise the time limit.
 		$raise_by = 0 === $limit || $limit > $max_execution_time ? $limit : $max_execution_time;
 
 		if ( function_exists( 'wc_set_time_limit' ) ) {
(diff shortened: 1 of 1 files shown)


vedanshujain (approved):
LGTM!

PR #927, opened 2023-03-08
Improve data-store initialization flow.

### LINKED ISSUE #839: as_enqueue_async_action shows error
Checking in my WooCommerce extension if an action has already been scheduled using `as_has_scheduled_action` like this `as_has_scheduled_action('run_my_function')` shows the following error: 

```
Notice: Undefined property: wpdb::$actionscheduler_actions in /Users/install/dev/apps/wordpress/wp-includes/wp-db.php on line 740
WordPress database error: [Table 'wordpress.a' doesn't exist]
SELECT a.action_id FROM a WHERE 1=1 AND a.hook='run_my_function' AND a.status IN ('in-progress', 'pending') LIMIT 0, 1
```

barryhughes:
We have safeguards within our procedural API that try to stop developers from using those functions too early (ie, before the active data-store has finished setting itself up). A good example of this can be found in [`as_schedule_single_function()`](https://github.com/woocommerce/action-scheduler/blob/3.5.4/functions.php#L18-L43), which has the following check:

```php
if ( ! ActionScheduler::is_initialized( __FUNCTION__ ) ) {
	return 0;
}
```

The problem is that, *currently,* we may [declare the data-store as having successfully initialized](https://github.com/woocommerce/action-scheduler/blob/3.5.4/classes/abstracts/ActionScheduler.php#L175) even though, in most cases, actual initialization has been [deferred until a future action](https://github.com/woocommerce/action-scheduler/blob/3.5.4/classes/abstracts/ActionScheduler.php#L153). This creates a pocket of time in which it is unsafe to call the procedural API. One such example, as described in the linked issue, is during `woocommerce_init` (effectively, this means during `init:0`), which can lead to a database error.

This PR changes things so that:

- If we wait until `init:1` before initializing the data-store, we also wait until after that before declaring the store is initialized.
- It additionally corrects a small oversight in a related [`error_log()`](https://www.php.net/manual/en/function.error-log.php) call, which could itself lead to problems. Specifically, we were passing `E_WARNING` as the second param which seems to break in most cases (and, in fact, doesn't seem like it is intended to be passed to that function at all).

Closes https://github.com/woocommerce/action-scheduler/issues/839.

---

### Testing instructions

- You will need some means of watching for errors. [Query Monitor](https://wordpress.org/plugins/query-monitor/) is a good choice, but you could also simply set the `WP_DEBUG` and `WP_DEBUG_DISPLAY` constants to true in your `wp-config.php` file if you prefer, and/or some other method you prefer.
- Add the following code to a suitable location. I would recommend `wp-content/mu-plugins/839-init.php`, which you can then remove after testing:

```php
<?php

	as_has_scheduled_action( 'foo' );
}, 0 );
```

With `trunk` checked out (ie, without this branch) you should be able to refresh any page or screen and see a database error complaining that `table *.a doesn't exist`:

db-error

Now checkout this branch: the above error should be cleared. Additionally, an attempt is made to write an entry to your system log, looking something like this:

```
as_has_scheduled_action() was called before the Action Scheduler data store was initialized
```

### Notes

- Given when these mechanics run (very early, and before the test suite actually starts to run) I could not think of an elegant way to write a corresponding unit test. I think that's fine, but am open to ideas and alternatives.
- This shouldn't introduce any breakages because, if we detect a function from our procedural API is called too early, we still return a valid value per each function signature.
- I thought about replacing the `error_log()` call with a `_doing_it_wrong()` call, but while that feels like a better choice in some ways, switching to a 'noisier' function like that one can also be problematic in some cases.

---

### Changelog

> Fix - Ensure warnings are issued if the procedural API is used too early.

### DIFF
--- a/classes/abstracts/ActionScheduler.php
+++ b/classes/abstracts/ActionScheduler.php
@@ -153,11 +153,27 @@ abstract class ActionScheduler {
 			add_action( 'init', array( $store, 'init' ), 1, 0 );
 			add_action( 'init', array( $logger, 'init' ), 1, 0 );
 			add_action( 'init', array( $runner, 'init' ), 1, 0 );
+
+			add_action(
+				'init',
+				/**
+				 * Runs after the active store's init() method has been called.
+				 *
+				 * It would probably be preferable to have $store->init() (or it's parent method) set this itself,
+				 * once it has initialized, however that would cause problems in cases where a custom data store is in
+				 * use and it has not yet been updated to follow that same logic.
+				 */
+				function () {
+					self::$data_store_initialized = true;
+				},
+				1
+			);
 		} else {
 			$admin_view->init();
 			$store->init();
 			$logger->init();
 			$runner->init();
+			self::$data_store_initialized = true;
 		}
 
 		if ( apply_filters( 'action_scheduler_load_deprecated_functions', true ) ) {
@@ -172,8 +188,6 @@ abstract class ActionScheduler {
 				$command->register();
 			}
 		}
-
-		self::$data_store_initialized = true;
 
 		/**
 		 * Handle WP comment cleanup after migration.
@@ -193,8 +207,12 @@ abstract class ActionScheduler {
 	 */
 	public static function is_initialized( $function_name = null ) {
 		if ( ! self::$data_store_initialized && ! empty( $function_name ) ) {
-			$message = sprintf( __( '%s() was called before the Action Scheduler data store was initialized', 'action-scheduler' ), esc_attr( $function_name ) );
-			error_log( $message, E_WARNING );
+			$message = sprintf(
+				/* translators: %s function name. */
+				__( '%s() was called before the Action Scheduler data store was initialized', 'action-scheduler' ),
+				esc_attr( $function_name )
+			);
+			error_log( $message );
 		}
 
 		return self::$data_store_initialized;


coreymckrill (approved):
👍 Works in my test environment. Although, for some reason the error log message was not consistently output to my error log file.

> I thought about replacing the error_log() call with a _doing_it_wrong() call, but while that feels like a better choice in some ways, switching to a 'noisier' function like that one can also be problematic in some cases.

`_doing_it_wrong` does seem like the right thing here. Given that it still only outputs an error when `WP_DEBUG` is enabled, where do you think it could be problematic?

barryhughes:
> `_doing_it_wrong` does seem like the right thing here. Given that it still only outputs an error when `WP_DEBUG` is enabled, where do you think it could be problematic?

Well, I'm a little sensitive to the fact that (in the context of WooCommerce) we've previously attempted to [deprecate a function](https://github.com/woocommerce/woocommerce/pull/31113) only to decide to [revert](https://github.com/woocommerce/woocommerce/pull/31672) due to the noise created by the resulting error messages and the breakages to ajax flows, etc.

As with `_doing_it_wrong`, this is only a 'problem' when `WP_DEBUG` is enabled, but it did clearly create enough of a disturbance that we changed course. So, though I agree it would be the better path in general, I also wanted to tread carefully and pause for thought.

barryhughes:
> Although, for some reason the error log message was not consistently output to my error log file.

Hmm. I'm not sure why that would be, unless it was when you were testing with the current `trunk` code (because the `$message_type` was/is invalid).

barryhughes:
> We've previously attempted to deprecate a function only to decide to revert due to the noise created by the resulting error messages and the breakages to ajax flows, etc.

Though, ruminating on this one some more, it's probably less common for premature calls to Action Scheduler's procedural API to take place, and, project maintainers have an opportunity to test things out before bundling the latest version of Action Scheduler 🤔

barryhughes:
...Will move ahead with this iteration, and we can revisit the choice of system error log (which is what we were already doing) vs a WP-style *doing it wrong* approach at a later point if required,

PR #928, opened 2023-03-10
Improve error handling across all supported PHP versions.

### LINKED ISSUE #391: Fatal error on order save in PHP 7.0+
First, I'll say this is definitely a 3rd party plugin issue, however I think Action Scheduler could handle this scenario gracefully. Maybe there's a way to prevent a fatal error in this scenario.

Second, I want to say I don't understand why this is happening. Someone smarter than me might be able to shed some light on that.

Third, the issue is not consistent. It doesn't happen every time I save/update an order from the admin area. I did manage to find a way to reproduce it consistently, but it's not really a situation that would happen often.

In order to produce the error I'm seeing, one could do the following:
1. Clean install of WP
2. Install WooCommerce, activate and setup
3. Install a plugin that makes use of the Action Scheduler and has hooks for actions on a post save. A good example of this would be to install Mailchimp for Woocommerce or WooCommerce Admin. Do NOT yet activate this plugin.
4. Install a plugin which is incorrectly calling `new WC_Order( $post_id )` instead of `wc_get_order( $post_id )` and has that in a function which gets called on `save_post`. An example plugin is attached below. The plugin does absolutely nothing, but does call `new WC_Order` every time a shop_order is saved.
5. Create a new order and save it. Don't need to enter any info, just need a created and saved blank order.
6. Activate either Mailchimp for WooCommerce or the WooCommerce Admin plugin (or presumably any other plugin which makes use of Action Scheduler on the update of an order).
7. Go back to the order you created and update it. Don't need to change anything, just click update. This will call `new WC_Order` and the fatal error shown relates to Action Scheduler.

Now, I'm fully aware the easy answer to this is that its a 3rd party plugin not using `wc_get_order` when it should be. But how specifically is that causing Action Scheduler to fail? And is it possible to fail more gracefully? And why does it only happen right after the plugin is activated, and not consistently?

EXAMPLE of code that is poorly written but definitely does exist in the plethora of available WooCommerce plugins out there:

```
<?php
/*
Plugin Name: WooCommerce Save Order Bug
*/

add_action( 'save_post', function( $post_id ) {

	//set some early returns. we only want this to happen when saving a shop order 
	if ( defined( 'DOING_AUTOSAVE' ) && DOING_AUTOSAVE ) return;
	if ( ! isset( $_POST['post_type'] ) ) return;
	if ( 'shop_order' != $_POST['post_type'] ) return;
	if ( ! current_user_can( 'edit_post', $post_id ) ) return;

	//now we're presumably saving a shop order
	
	//lets create a WC_Order from this post_id
	$order = new WC_Order($post_id);
	
	//we don't even need to do anything with $order, the error will happen (or not)
	//but an example of something someone might be doing in this instance...
	//maybe we want to log every time one of our users saves an order
	//$order->add_order_note( "Order was updated by {$user_name}" );

});
```

barryhughes:
To the greatest possible extent, we want to prevent problematic actions (those that raise errors, exceptions or throwables) from preventing other actions in the current batch from being processed. Since we continue to support a wide range of PHP versions, this requires special consideration (error handling having evolved considerably across the versions we support).

Props to @ju1ius for the idea behind this change. The general idea being that raised errors are converted into exceptions by a custom error handler (which we remove when we're finished). Additionally, we add support for catching throwables—which is PHP 5.6 compatible even though they do not exist in that version.

Closes #391.

---

### Testing instructions

- Code review is likely sufficient here, but I encourage you to simulate a range of different error conditions across the supported range of PHP versions (5.6 through to 8.2).
- Note that the new test will be executed across the usual matrix of PHP and WP combinations (which includes PHP 5.6).

### Changelog

> Fix - Improve error handing across all supported PHP versions.

### DIFF
--- a/classes/abstracts/ActionScheduler_Abstract_QueueRunner.php
+++ b/classes/abstracts/ActionScheduler_Abstract_QueueRunner.php
@@ -48,34 +48,93 @@ abstract class ActionScheduler_Abstract_QueueRunner extends ActionScheduler_Abst
 	 *        Generally, this should be capitalised and not localised as it's a proper noun.
 	 */
 	public function process_action( $action_id, $context = '' ) {
+		// Temporarily override the error handler while we process the current action.
+		set_error_handler(
+			/**
+			 * Temporary error handler which can catch errors and convert them into exceptions. This faciliates more
+			 * robust error handling across all supported PHP versions.
+			 *
+			 * @throws Exception
+			 *
+			 * @param int    $type    Error level expressed as an integer.
+			 * @param string $message Error message.
+			 */
+			function ( $type, $message ) {
+				throw new Exception( $message );
+			},
+			E_USER_ERROR | E_RECOVERABLE_ERROR
+		);
+
+		/*
+		 * The nested try/catch structure is required because we potentially need to convert thrown errors into
+		 * exceptions (and an exception thrown from a catch block cannot be caught be a later catch block in the same
+		 * structure).
+		 */
 		try {
-			$valid_action = false;
-			do_action( 'action_scheduler_before_execute', $action_id, $context );
+			try {
+				$valid_action = false;
+				do_action( 'action_scheduler_before_execute', $action_id, $context );
 
-			if ( ActionScheduler_Store::STATUS_PENDING !== $this->store->get_status( $action_id ) ) {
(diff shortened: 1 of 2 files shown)


vedanshujain (approved):
Interesting PR, let's ship in a new version as soon as possible.

PR #929, opened 2023-03-10
Add init hook

### LINKED ISSUE #749: Feature request: 'ready' hook
[Action Scheduler documentation](https://actionscheduler.org/usage/) is telling me to hook to WordPress's `init` to use its functions.

If I do this and then disable Action Scheduler plugin, there will be errors when I call the missing functions.

For now I'll use

```
if( ! function_exists( 'as_next_scheduled_action' ) ) {
    return;
}
```

but a neater way would be for Action Scheduler to fire its own init, then I can hook to that and be sure the functions will exist.

So something like this inside Action Scheduler:

```
add_action( 'init', function() { do_action( 'actionshecduler_init' ); }, 1 );
```

I searched the code for all `do_action`s and didn't find any that suited this idea.

It seems this would be implemented in [`ActionScheduler::init()`](https://github.com/woocommerce/action-scheduler/blob/92f4146de3dcf7b01967652cd141d6b14214cd36/classes/abstracts/ActionScheduler.php#L155) with an `add_action` in the first part of the conditional calling the function with the `do_action`, and a direct call to that new, one line function in the second.

barryhughes:
As suggested in the linked issue, introduces new hook `action_scheduler_init`. It is safe to use the procedural API from this action onwards.

Closes https://github.com/woocommerce/action-scheduler/issues/749.

---

### Notes

- Code review should be sufficient, there are no functional changes.
- Builds on https://github.com/woocommerce/action-scheduler/pull/927 (which tweaks the initialization procedure). Keeping as a draft until that is merged.

---

### Changelog

> Add - New `action_scheduler_init` hook.

### DIFF
--- a/classes/abstracts/ActionScheduler.php
+++ b/classes/abstracts/ActionScheduler.php
@@ -165,6 +165,13 @@ abstract class ActionScheduler {
 				 */
 				function () {
 					self::$data_store_initialized = true;
+
+					/**
+					 * Fires when Action Scheduler is ready: it is safe to use the procedural API after this point.
+					 *
+					 * @since 3.5.5
+					 */
+					do_action( 'action_scheduler_init' );
 				},
 				1
 			);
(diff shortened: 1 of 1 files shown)


barryhughes:
#927 is merged, so this can now be reviewed.

Konamiman (approved):

PR #938, opened 2023-04-14
Document new optional `$priority` arg for various API functions.

barryhughes:
Updates [actionscheduler.org/api](https://actionscheduler.org/api/) with information about the new `$priority` argument added in https://github.com/woocommerce/action-scheduler/pull/921/.

Target branch is `docs/3.6.0`, the idea being that once this (and related PRs are approved) we have a convenient way to merge and deploy once 3.6.0 goes live.

PR #939, opened 2023-04-14
Document the new `action_scheduler_init` hook.

barryhughes:
Adds information about the new `action_scheduler_init` hook, added in https://github.com/woocommerce/action-scheduler/pull/929/.

Though otherwise unrelated, this PR 'stacks' on top of https://github.com/woocommerce/action-scheduler/pull/938 since it touches one of the same files (basic idea being to keep all doc changes together, then merge and deploy when 3.6.0 goes live).

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -88,7 +88,7 @@ Action Scheduler will register its version on `'plugins_loaded'` with priority `
 
 It is recommended to load it _when the file including it is included_. However, if you need to load it on a hook, then the hook must occur before `'plugins_loaded'`, or you can use `'plugins_loaded'` with negative priority, like `-10`.
 
-Action Scheduler will later initialize itself on `'init'` with priority `1`.  Action Scheduler APIs should not be used until after `'init'` with priority `1`.
+Action Scheduler will later initialize itself on `'init'` with priority `1`.  Action Scheduler APIs should not be used until after `'init'` with priority `1`. As described in [API Function Availability](/api/#api-function-availability), you can also use the `'action_scheduler_init'` action hook for this purpose.
 
 ### Usage in Themes
 
--- a/docs/api.md
+++ b/docs/api.md
@@ -22,9 +22,10 @@ Functions return similar values and accept similar arguments to their WP-Cron co
 
 As mentioned in the [Usage - Load Order](usage.md#load-order) section, Action Scheduler will initialize itself on the `'init'` hook with priority `1`. While API functions are loaded prior to this and can be called, they should not be called until after `'init'` with priority `1`, because each component, like the data store, has not yet been initialized.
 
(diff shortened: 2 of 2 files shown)


docs/usage.md
```diff
@@ -88,7 +88,7 @@ Action Scheduler will register its version on `'plugins_loaded'` with priority `
 
 It is recommended to load it _when the file including it is included_. However, if you need to load it on a hook, then the hook must occur before `'plugins_loaded'`, or you can use `'plugins_loaded'` with negative priority, like `-10`.
 
-Action Scheduler will later initialize itself on `'init'` with priority `1`.  Action Scheduler APIs should not be used until after `'init'` with priority `1`.
+Action Scheduler will later initialize itself on `'init'` with priority `1`.  Action Scheduler APIs should not be used until after `'init'` with priority `1`. As described in [API Function Availability](/api/#api-function-availability), you can also use the `'action_scheduler_init'` action hook for this purpose.
```
coreymckrill, line 91: The "API Function Availability" link goes to a 404 when I click it...

coreymckrill (changes requested):

docs/usage.md
barryhughes: If you are previewing here in GitHub, but outside of Jekyll—such as via [this URL](https://github.com/woocommerce/action-scheduler/blob/5fe09ad996befae7539cfbe1a211001c512c1ee0/docs/usage.md)—it won't work, but it should work when rendered via Jekyll. As a confidence measure, if you don't want to setup Jekyll locally, you can take a peek at a very similar link [on this page](https://actionscheduler.org/usage/):

> ### Scheduling an Action
>
> To schedule an action, call the [API function](https://actionscheduler.org/api/) for the desired schedule type passing in the required parameters.

It's effectively the same link, albeit missing the `#fragment` and in the source looks like `[API function](/api/)`. I preferred this to using a full URL with scheme and domain for portability, but in practice it probably doesn't matter a whole bunch so can update it to use the fully qualified URL if you like.

barryhughes: I should have added a link to this in the PR description, but there is guidance [over here](https://docs.github.com/en/pages/setting-up-a-github-pages-site-with-jekyll/testing-your-github-pages-site-locally-with-jekyll) on locally testing GitHub Pages sites.

coreymckrill: Ah ok. I forgot that these files were used on a separate site. Too bad the format can't work for both the Jekyll site and when viewing as rendered Markdown here on GitHub 🤔

coreymckrill (approved):

docs/usage.md
barryhughes: Yes ... I think the problem is it doesn't know that the `docs/` directory is special (I mean, of course it *could* know that it is our assigned GH Pages directory, but it seems not to), and I guess it's handling would be correct if it were literally just another directory in which we happen to stash documentation, vs being a web root.

PR #946, opened 2023-05-01
Remove SHOW TABLES usage when checking if tables exist.

tommyshellberg:
ActionScheduler_Abstract_Schema::get_tables() fails on some hosts(WordPress.com for instance) because [HyperDB](https://github.com/Automattic/HyperDB) can't [find a table from the query](https://github.com/Automattic/HyperDB/blob/09c962470332f4dee70d57282bd67d56dd4864ab/db.php#L304). HyperDB allows for splitting a table across multiple servers so it needs to find the table based on the query.
It's failing here because the query is trying to return _all_ of the tables which HyperDB doesn't know what to do with.

`[06-Apr-2023 11:39:59 UTC] WordPress database error Unable to determine dataset (for table: ) for query SHOW TABLES made by ActionScheduler_Abstract_Schema::tables_exist`

This change goes through the expected tables and performs a SHOW TABLES LIKE query on each, and if one fails, marks the return of the `tables_exist()` function as `false`.

### Changelog

> Add - Improved compability when running alongside HyperDB.

### DIFF
--- a/classes/abstracts/ActionScheduler_Abstract_Schema.php
+++ b/classes/abstracts/ActionScheduler_Abstract_Schema.php
@@ -25,7 +25,7 @@ abstract class ActionScheduler_Abstract_Schema {
 	/**
 	 * @var array Names of tables that will be registered by this class.
 	 */
-	protected $tables = [];
+	protected $tables = array();
 
 	/**
 	 * Can optionally be used by concrete classes to carry out additional initialization work
@@ -90,10 +90,10 @@ abstract class ActionScheduler_Abstract_Schema {
 			$plugin_option_name = 'schema-';
 
 			switch ( static::class ) {
-				case 'ActionScheduler_StoreSchema' :
+				case 'ActionScheduler_StoreSchema':
 					$plugin_option_name .= 'Action_Scheduler\Custom_Tables\DB_Store_Table_Maker';
 					break;
-				case 'ActionScheduler_LoggerSchema' :
+				case 'ActionScheduler_LoggerSchema':
 					$plugin_option_name .= 'Action_Scheduler\Custom_Tables\DB_Logger_Table_Maker';
 					break;
 			}
@@ -129,7 +129,7 @@ abstract class ActionScheduler_Abstract_Schema {
 	 * @return void
 	 */
 	private function update_table( $table ) {
-		require_once( ABSPATH . 'wp-admin/includes/upgrade.php' );
+		require_once ABSPATH . 'wp-admin/includes/upgrade.php';
 		$definition = $this->get_table_definition( $table );
 		if ( $definition ) {
 			$updated = dbDelta( $definition );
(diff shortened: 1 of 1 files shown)


barryhughes (commented):
Curious: instead of running multiple individual queries, would HyperDB support doing this via a single query that mentions each table name? That is, using a query shaped something like:

```sql
SHOW TABLES WHERE `tables_in_dbname` IN ( 'table_1', 'table_2', 'table_3' )
```

If viable, this might be preferable.

tommyshellberg:
Thanks for the review! I don't see a way to make your suggestion work, unfortunately.
1. wpdb [doesn't have a parser for the IN clause](https://github.com/WordPress/WordPress/blob/fe6b65c44b929ee5583f6146d92388ef43061f9a/wp-includes/class-wpdb.php#L3768).
2. We don't know the dbname to put in as `tables_in_dbname` in your example because HyperDB figures out the database name from the table; a table could be spread across multiple databases.

classes/abstracts/ActionScheduler_Abstract_Schema.php
```diff
@@ -159,14 +159,25 @@ protected function get_full_table_name( $table ) {
...
+			$pattern        = str_replace( '_', '\\_', $table_name );
+			$existing_table = $wpdb->get_var( $wpdb->prepare( 'SHOW TABLES LIKE %s', $pattern ) );
+
+			if ( $existing_table !== $table_name ) {
+				$tables_exist = false;
+				break;
+			}
+		}
+
+		return $tables_exist;
```
barryhughes: Instead of looping twice (`array_map` followed by a `foreach`), seems like we could use a single loop? We could also switch to using `esc_like`.

```php
public function tables_exist() {
	global $wpdb;

	foreach ( $this->tables as $table_name ) {
		$table_name     = $wpdb->prefix . $table_name;
		$existing_table = $wpdb->get_var( $wpdb->prepare( 'SHOW TABLES LIKE %s', $wpdb->esc_like( $table_name ) ) );

		if ( $existing_table !== $table_name ) {
			return false;
		}
	}

	return true;
}
```

barryhughes (changes requested):
Thanks, HyperDB isn't something I'm particularly familiar with, so I appreciate you looking into that. 

Left one other note, and beyond that I think this is good to go :+1:

classes/abstracts/ActionScheduler_Abstract_Schema.php
tommyshellberg: Thanks for the suggestions; I think we're good to go now 👍 
> We could also switch to using esc_like.

Unfortunately, [this breaks HyperDB](https://core.trac.wordpress.org/ticket/57149) 😬
Which is why the `str_replace()` is being performed rather than use `esc_like()`.

barryhughes (approved):
Thanks!

barryhughes:
Added changelog text.

Issue #798, opened 2022-02-05
$as_args only passing the first element of the array


btolle89:
I schedule an action using as_schedule_single_action and pass an array $as_args
I can see that the array made it into Action Scheduler in the UI within woo. (see attached)
![image](https://user-images.githubusercontent.com/29187982/152645298-f839d49f-15c3-471a-b48e-3048463866e3.png)

However, when the action runs, and the args do not seem to pass to the function properly. 
Only the first element of the array appears as $as_args

I have 
`add_action( "add_tag_and_user_callback" , "add_tag_and_user", 10, 1 ); `
and
```
function add_tag_and_user( $as_args ){

	error_log(print_r('add tag and user ran!',true));
	error_log(print_r($as_args,true));
}
```

but I can see in the error log, that the full array has not passed. it's only the first element
```
[05-Feb-2022 14:01:15 UTC] add tag and user ran!
[05-Feb-2022 14:01:15 UTC] 1938
```

what am I doing wrong?

barryhughes:
Hi @btolle89,

The array represents a set of arguments for your callback (in other words, rather than supplying the array in its entirety, it is broken into individual elements and each acts as a single argument). So, if you do:

```php
as_enqueue_async_action( 'my_task', array( 'foo-foo', 'bar-bar' ) );
```

You would then do something like this, taking care to set the number of `$accepted_args` appropriately:

```php
add_action( 'my_task', 'my_callback', 10, 2 );

function my_callback( $foo, $bar ) {
    # $foo := (string) 'foo-foo'
    # $bar := (string) 'bar-bar'
}
```

Does that help?

mischnu:
I’m having the same issue as @btolle89, only the first item of the array is passed as argument to the callback. The rest of the array seems to be omitted.

This is what my callback looks like:
```PHP
add_action( 'as_set_customer_status', 'set_customer_status' );
function set_customer_status( $customer_id, $status ) {
    wp_update_post( array(
        'ID'          =>  $customer_id,
        'post_status' =>  $status,
    ) );
}
```

Now when I schedule a single action like this:
```PHP
as_schedule_single_action( strtotime( '+1 min' ), 'as_set_customer_status', array( 128, 'trial-expired' ), 'set_customer_status' );
```
the task fails and throws a PHP fatal error:
[...]

PR #969, opened 2023-07-11
Add guidance about passing arguments.

barryhughes:
- Updates our usage documentation to provide some (very high-level) guidance about passing arguments.
- The need for this is based on the discussion in https://github.com/woocommerce/action-scheduler/issues/798.
- No changelog required.

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -125,7 +125,7 @@ add_action( 'eg_midnight_log', 'eg_log_action_data' );
 
 Note that the `as_has_scheduled_action()` function was added in 3.3.0: if you are using an earlier version, you should use `as_next_scheduled_action()` instead. For more details on all available API functions, and the data they accept, refer to the [API Reference](/api/).
 
-## Passing arguments
+### Passing arguments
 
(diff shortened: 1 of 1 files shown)


docs/usage.md
```diff
@@ -124,3 +124,29 @@ add_action( 'eg_midnight_log', 'eg_log_action_data' );
 ```
 
 Note that the `as_has_scheduled_action()` function was added in 3.3.0: if you are using an earlier version, you should use `as_next_scheduled_action()` instead. For more details on all available API functions, and the data they accept, refer to the [API Reference](/api/).
+
+## Passing arguments
```
coreymckrill: Should this be a level 3 header? Seems like it's still related to "Scheduling an action" above...

coreymckrill (approved):
Approving, but not merging in case you have thoughts about the one comment below.

docs/usage.md
barryhughes: 🤔 I initially felt it was different—but I see your point.

barryhughes: Updated in https://github.com/woocommerce/action-scheduler/pull/969/commits/c4420712d82a3d6f0a41c1ad15bf9bf47e99c422.

PR #992, opened 2023-09-20
Add extended indexes for hook_status_scheduled_date_gmt and status_sheduled_date_gmt

### LINKED ISSUE #978: Poor query performance in as_has_scheduled_action() and as_next_scheduled_action()
Some of the functions provided to check if an action is already scheduled have poor performance.  Specifically, `as_has_scheduled_action()` and `as_next_scheduled_action()`.  The format of these queries are as follows:

```
SELECT a.action_id FROM wp_actionscheduler_actions a LEFT JOIN wp_actionscheduler_groups g ON g.group_id=a.group_id WHERE 1=1 AND g.slug='%s' AND a.hook='%s' AND a.args='[]' AND a.status IN ('in-progress', 'pending') LIMIT 0, 1
```

When a site has been running for a while and has a large number of completed actions in the table, the EXPLAIN starts to look like the following:

```
+----+-------------+-------+-------------+---------------------------+-----------+---------+---------------------+------+-----------------------------------------+
| id | select_type | table | type        | possible_keys             | key       | key_len | ref                 | rows | Extra                                   |
+----+-------------+-------+-------------+---------------------------+-----------+---------+---------------------+------+-----------------------------------------+
| 1  | SIMPLE      | a     | index_merge | hook,status,args,group_id | args,hook | 194,193 |                     | 245  | Using intersect(args,hook); Using where |
| 1  | SIMPLE      | g     | eq_ref      | PRIMARY,slug              | PRIMARY   | 8       | wp.a.group_id | 1    | Using where                             |
+----+-------------+-------+-------------+---------------------------+-----------+---------+---------------------+------+-----------------------------------------+
```

Note: a filesort will be used as well for `as_next_scheduled_action()`.

As these tend to get used to determine if an action is already scheduled to prevent creating a duplicate action, these checks often are setup to run on every page load by other plugins - though an upgrade/install script would be better suited.  Either way we should look at improving these queries.

prettyboymp:
Fixes #978

The current `hook` and `status` indexes can be pretty useless as there isn't enough uniqueness in the generated index to add much value to query performance.  Extending these keys to include the `status` and `scheduled_date_gmt` make them much more useful for functions like `as_next_scheduled_action()` and `ActionScheduler_Store::has_pending_actions_due()`.

### Changelog

> Fix - Improve indicies within the actions table.

### DIFF
--- a/classes/schema/ActionScheduler_StoreSchema.php
+++ b/classes/schema/ActionScheduler_StoreSchema.php
@@ -38,6 +38,7 @@ class ActionScheduler_StoreSchema extends ActionScheduler_Abstract_Schema {
 		$table_name       = $wpdb->$table;
 		$charset_collate  = $wpdb->get_charset_collate();
 		$max_index_length = 191; // @see wp_get_db_schema()
+		$hook_status_scheduled_date_gmt_max_index_length = $max_index_length - 20 - 8; // - status, - scheduled_date_gmt
 		$default_date     = self::DEFAULT_DATE;
 		switch ( $table ) {
 
@@ -59,8 +60,8 @@ class ActionScheduler_StoreSchema extends ActionScheduler_Abstract_Schema {
 				        claim_id bigint(20) unsigned NOT NULL default '0',
 				        extended_args varchar(8000) DEFAULT NULL,
 				        PRIMARY KEY  (action_id),
-				        KEY hook (hook($max_index_length)),
-				        KEY status (status),
+				        KEY hook_status_scheduled_date_gmt (hook($hook_status_scheduled_date_gmt_max_index_length), status, scheduled_date_gmt),
+				        KEY status_scheduled_date_gmt (status, scheduled_date_gmt),
 				        KEY scheduled_date_gmt (scheduled_date_gmt),
 				        KEY args (args($max_index_length)),
 				        KEY group_id (group_id),


lsinger:
Hello @prettyboymp! You assigned @peterfabian to this PR -- did you mean to request a review from him? 

In any case he has switched teams -- if this PR is ready for review then please request a review from `@woocommerce/proton`. ☺️

barryhughes (commented):
This won't take effect for existing installations without some migration code: we don't necessarily need that to merge (there are valid reasons not to), but I wanted to check and make sure that's also fine from your perspective, @prettyboymp?

prettyboymp:
> This won't take effect for existing installations without some migration code: we don't necessarily need that to merge (there are valid reasons not to), but I wanted to check and make sure that's also fine from your perspective, @prettyboymp?

🤔 This is what I wasn't sure about.  I haven't looked into the update process for Action Scheduler yet.  I wanted to get some initial feedback before diving in too deep with it.  So Action Scheduler tables are excluded from the dbDelta queries during a Woo update?

Either way, it may be a good idea to add a schema update step for this so the previous indexes can be dropped to save disk space.  I can work on that.

barryhughes:
The table definitions still pass through dbDelta, but my memory is that key changes will not be applied if the tables already exist.

...That said, we've had some discussion in the past as to whether it is even a good idea to modify keys for existing installs (since, especially for sites of a certain size, they may anyway have been tweaked and fine-tuned in a project-specific way).

prettyboymp:
Looking through the dbDelta code, I believe that new indexes will be automatically created, but existing indexes won't be modified, i.e. if the name stays the same, but the parts change, dbDelta won't modify the index.

In this case, both indexes had the names changed as well, so it would add them on the next dbDelta run but won't remove the previous indexes.  That should be fine for most use cases but allow those that specifically need to drop the old indexes for space have that separate control.

barryhughes (approved):
Thanks (and sorry for the delay in updating this)!

PR #993, opened 2023-09-20
Fix: Use orderby => 'none' when bulk cancelling actions.

prettyboymp:
The orderby clause when querying actions to cancel can result in a filesort operation for larger tables as there is no corresponding index that provides the where clause followed by the action_id.  Since any actions being cancelled transition to another status, we don't need to have a sort in the initial query to read the IDs.

### DIFF
--- a/classes/abstracts/ActionScheduler_Store.php
+++ b/classes/abstracts/ActionScheduler_Store.php
@@ -347,7 +347,7 @@ abstract class ActionScheduler_Store extends ActionScheduler_Store_Deprecated {
 					'hook'     => $hook,
 					'status'   => self::STATUS_PENDING,
 					'per_page' => 1000,
-					'orderby'  => 'action_id',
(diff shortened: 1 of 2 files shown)


coreymckrill (approved):
👍 Looks good to me. I did a quick smoke test to make sure that `as_unschedule_all_actions` still works as expected. Thanks @prettyboymp 

Side note: it seems... not ideal that `ActionScheduler_DBStore::bulk_cancel_actions()` has a different method signature than the parent `ActionScheduler_Store::bulk_cancel_actions()`.

rrennick:
> Side note: it seems... not ideal that ActionScheduler_DBStore::bulk_cancel_actions() has a different method signature than the parent ActionScheduler_Store::bulk_cancel_actions().

@coreymckrill An issue should be created for that as it is problematic in PHP 8.1+.

coreymckrill:
Done: #998

PR #995, opened 2023-09-22
Fix option lock test

barryhughes:
Fixes a couple of items (thank you @lsinger for raising awareness!):

- The lock tests were not part of the main test suite.
- One of the lock tests made the wrong assumption about the shape of an update query, and was failing.

Worth calling out: I'm deliberately suppressing database errors (but only temporarily). Without the suppression, if you run the test suite locally, you will notice unwanted noise in the output:

Screenshot 2023-09-22 at 09 17 39

If there are any concerns with this, let me know.

### DIFF
--- a/tests/phpunit.xml.dist
+++ b/tests/phpunit.xml.dist
@@ -23,6 +23,7 @@
 		<testsuite name="Action Scheduler">
 			<directory>./phpunit/helpers</directory>
 			<directory>./phpunit/jobs</directory>
+			<directory>./phpunit/lock</directory>
 			<directory>./phpunit/procedural_api</directory>
 			<directory>./phpunit/runner</directory>
 			<directory>./phpunit/schedules</directory>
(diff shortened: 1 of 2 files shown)


lsinger (approved):
Looks good to me. :shipit: 

Two of the unit test runs are failing because [WordPress 6.3 requires PHP 7+](https://github.com/WordPress/wordpress-develop/commit/f7dbb2462ba35362be2ab632f466ba1a4d760282), which is of course unrelated to this PR. I'll open a separate issue for that.

barryhughes:
> Two of the unit test runs are failing because [WordPress 6.3 requires PHP 7+](https://github.com/WordPress/wordpress-develop/commit/f7dbb2462ba35362be2ab632f466ba1a4d760282), which is of course unrelated to this PR. I'll open a separate issue for that.

Thanks! Yes, and we also need to make it possible to run the tests under 8.2 (but that's yet another problem...).

Issue #996, opened 2023-09-25
Unit test matrix needs updating for WordPress 6.3 dropping PHP 5 support


lsinger:
Action Scheduler's [unit test matrix](https://github.com/woocommerce/action-scheduler/blob/trunk/.github/workflows/pr-unit-tests.yml#L11) tries to run tests with PHP 5.6 and WordPress latest. As of 6.3, [WordPress supports only PHP versions 7+](https://github.com/WordPress/wordpress-develop/commit/f7dbb2462ba35362be2ab632f466ba1a4d760282), so the "PHP 5.6 / WordPress latest" tests have been failing very quickly. 

- [x] **1.** We should start by excluding that combination from the test matrix. #997
- [x] **2.** Further, we should look into our support policy for Action Scheduler or define one if we don't have one yet -- in the matrix, we list several WordPress versions up to 5.6, leaving a large gap until the current 6.3. One option would be to simply copy [WooCommerce's policy](https://developer.woocommerce.com/2023/06/05/new-requirement-for-woocommerce-8-2-php-7-4/), but maybe that's too simple because Action Scheduler is used in several other plugins as well? (To update: [test matrix](https://github.com/woocommerce/action-scheduler/blob/trunk/.github/workflows/pr-unit-tests.yml#L11), [plugin file](https://github.com/woocommerce/action-scheduler/blob/trunk/action-scheduler.php#L11), [readme.txt](https://github.com/woocommerce/action-scheduler/blob/trunk/readme.txt#L6), [composer.json](https://github.com/woocommerce/action-scheduler/blob/trunk/composer.json#L21), and possibly other places.)
- [x] **3.** Once we have a policy, we should add keeping it current to one of the team's rotations or hats.

lsinger:
After an internal discussion (p7bje6-5o3-p2) we now have a team assigned to keep track of version updates. Closing. 

/cc @barryhughes

PR #997, opened 2023-09-25
Update PR unit tests matrix

lsinger:
Action Scheduler's [unit test matrix](https://github.com/woocommerce/action-scheduler/blob/trunk/.github/workflows/pr-unit-tests.yml#L11) tries to run tests with PHP 5.6 and WordPress latest. As of 6.3, [WordPress supports only PHP versions 7+](https://github.com/WordPress/wordpress-develop/commit/f7dbb2462ba35362be2ab632f466ba1a4d760282), so the "PHP 5.6 / WordPress latest" tests have been failing very quickly.

This PR excludes that combination from the test matrix (and also updates the `.gitignore` file by picking the most relevant bits from WooCommerce's).

Also see https://github.com/woocommerce/action-scheduler/issues/996 for additional tasks around this issue.

### DIFF
--- a/.github/workflows/pr-unit-tests.yml
+++ b/.github/workflows/pr-unit-tests.yml
@@ -25,6 +25,9 @@ jobs:
             wp: 5.4
           - php: 8.0
             wp: 5.5
+          # WordPress 6.3+ requires PHP 7.0+
+          - php: 5.6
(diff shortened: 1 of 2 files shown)


barryhughes (approved):
Thank you!

Issue #973, opened 2023-08-02
Problems when db user cannot create or alter tables


thomasgaillard:
Hi there

Some of my Wordpress sites didn't have admin rights on the SQL user side when I upgraded Woocommerce.

I now have fatal errors because the "priority" column in the wp_actionscheduler_actions table was not created when I updated Woocommerce (and therefore ActionScheduler).

How can I manually replay the last two ActionScheduler core updates please?

Manual creation using the sql command given on another workstation doesn't work:
ALTER TABLE `wp_actionscheduler_actions` ADD `priority` TINYINT(3) UNSIGNED NOT NULL DEFAULT '10' AFTER `extended_args`;

And I want to repatch the whole thing.
I don't have access to WP CLI.

Thanks for your help, I've already spent many hours

<img width="1067" alt="Capture d’écran 2023-08-02 à 17 48 02" src="https://github.com/woocommerce/action-scheduler/assets/681454/16f1569a-2a6d-471c-8690-37191cab0899">

thomasgaillard:
I finally success to install WP CLI and run the command for manually update ActionScheduler : and everything works !!

ULHI-xin:
Same here. My whole WP site was down because the SQL trying inserting a new row with field `priority` but the column was not added yet.

It seems no straightforward way to fix it, including disabling all plugin and installing Action Scheduler plugin directly.

But I finally got it fixed by running a manual DB DDL (actually just `alter table add column`). If you have privilege to run DDL you might try this way.

ULHI-xin:
> I finally success to install WP CLI and run the command for manually update ActionScheduler : and everything works !!

Hi @thomasgaillard would you mind sharing the CLI command to do the manual update? I believe it would be very helpful.

> Manual creation using the sql command given on another workstation doesn't work:
ALTER TABLE wp_actionscheduler_actions ADD priority TINYINT(3) UNSIGNED NOT NULL DEFAULT '10' AFTER extended_args;

Also I'm a little bit curious why the `alter table` is not working for you. In my case, it solve the problem as well. Just FYI.

thomasgaillard:
Helo @ULHI-xin,

I had an sql error while running the command `alter table` .. and I don't know if this "Priority column" was the only one missing
So the idea was to manually update/repair with these following commands :

1. Ssh connexion
2. WP CLi install `curl -O https://raw.githubusercontent.com/wp-cli/builds/gh-pages/phar/wp-cli.phar`
[...]

PR #1000, opened 2023-10-09
Catch and log exceptions thrown when actions can't be created, e.g. under a corrupt database schema

lsinger:
This PR attempts to address parts of #973 by catching and logging the exceptions that are thrown when actions can't be created. If, e.g., a database schema migration did not complete properly, scheduling an action has been able to render a site inaccessible because exceptions just bubbled up to the user. 

This PR explicitly does not solve completing or fixing incomplete schema migrations -- that's a different task. The first step implemented here is to lessen the impact of such issues. 

To test this manually, you could take a working site with AS running fine and then drop the priority column. Without this PR the exceptions thrown when trying to schedule actions bubble up to the user, with this PR applied they're merely logged. AS isn't functional then but at least we limit the damage. 

Alternatively apply the commits from this branch sequentially and run the test suite in between: 
- ff2c680bc0fd48ca790359ab8ebd86343ddf9f77 makes the test suite fail from the uncaught exceptions
- 25a43f2e47cbc662fdc8c1d0ac3131e41274696f catches the exceptions and makes the test suite pass
- e1bd2e9d9d4edc3a228e0286c0f125ce06991ec8 adds some requirements around how we log the exceptions to the tests and makes the test suite fail again
- e5c427f942f8259e52123e3c1a67169e6f2be464 adds the exception message to the log entry and thus makes the test suite pass again
- (27bf7496c3fb176f8dac2c0ff23d3353d4a2181f and b0864e020f69265fdafeefad3d86799c6b31015b are not covered by tests)

### Changelog 

> Fix - Reduce impact when the database schema cannot be modified following an update.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -53,8 +53,7 @@ function as_enqueue_async_action( $hook, $args = array(), $group = '', $unique =
 			)
 		);
 	} catch ( Exception $exception ) {
-		ActionScheduler::logger()->log(
-			0,
+		error_log(
 			sprintf(
 				/* translators: %1$s is the name of the hook to be enqueued, %2$s is the exception message. */
 				__( 'Caught exception while enqueuing action "%1$s": %2$s', 'action-scheduler' ),
@@ -117,8 +116,7 @@ function as_schedule_single_action( $timestamp, $hook, $args = array(), $group =
 			)
 		);
 	} catch ( Exception $exception ) {
-		ActionScheduler::logger()->log(
-			0,
+		error_log(
 			sprintf(
 				/* translators: %1$s is the name of the hook to be enqueued, %2$s is the exception message. */
 				__( 'Caught exception while enqueuing action "%1$s": %2$s', 'action-scheduler' ),
@@ -203,8 +201,7 @@ function as_schedule_recurring_action( $timestamp, $interval_in_seconds, $hook, 
 			)
 		);
 	} catch ( Exception $exception ) {
-		ActionScheduler::logger()->log(
-			0,
+		error_log(
 			sprintf(
 				/* translators: %1$s is the name of the hook to be enqueued, %2$s is the exception message. */
 				__( 'Caught exception while enqueuing action "%1$s": %2$s', 'action-scheduler' ),
@@ -282,8 +279,7 @@ function as_schedule_cron_action( $timestamp, $schedule, $hook, $args = array(),
 			)
 		);
 	} catch ( Exception $exception ) {
-		ActionScheduler::logger()->log(
-			0,
+		error_log(
 			sprintf(
 				/* translators: %1$s is the name of the hook to be enqueued, %2$s is the exception message. */
 				__( 'Caught exception while enqueuing action "%1$s": %2$s', 'action-scheduler' ),
--- a/functions.php
+++ b/functions.php
@@ -14,7 +14,7 @@
  * @param bool   $unique Whether the action should be unique.
  * @param int    $priority Lower values take precedence over higher values. Defaults to 10, with acceptable values falling in the range 0-255.
  *
- * @return int The action ID.
+ * @return int The action ID. Zero if there was an error scheduling the action.
  */
 function as_enqueue_async_action( $hook, $args = array(), $group = '', $unique = false, $priority = 10 ) {
 	if ( ! ActionScheduler::is_initialized( __FUNCTION__ ) ) {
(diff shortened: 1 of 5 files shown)


functions.php
```diff
@@ -91,17 +104,30 @@ function as_schedule_single_action( $timestamp, $hook, $args = array(), $group =
...
-			'hook'      => $hook,
-			'arguments' => $args,
-			'when'      => $timestamp,
-			'group'     => $group,
-			'unique'    => $unique,
-			'priority'  => $priority,
-		)
-	);
+	try {
+		return ActionScheduler::factory()->create(
```
barryhughes: I don't think there's a right or wrong answer here, but what are your thoughts on any of the following permutations:

- Moving the try/catch structures, complete with logging code, inside `ActionScheduler::factory()->create()`?
- Or, moving the logging code inside `ActionScheduler::factory()->create()` but still catch the exception here?

Primarily thinking this would allow us to reduce the degree of repetition across these function definitions.

functions.php
```diff
@@ -91,17 +104,30 @@ function as_schedule_single_action( $timestamp, $hook, $args = array(), $group =
...
+		ActionScheduler::logger()->log(
+			0,
+			sprintf(
+				/* translators: %1$s is the name of the hook to be enqueued, %2$s is the exception message. */
+				__( 'Caught exception while enqueuing action "%1$s": %2$s', 'action-scheduler' ),
+				$hook,
+				$exception->getMessage()
+			)
+		);
+		return 0;
```
barryhughes: It was already possible for these functions to return zero, but I wonder if we should take this opportunity to highlight this via the docblock? Something like:

```diff
- * @return int The action ID.
+ * @return int The action ID. If zero, the action was not created and an error may have been logged in the database.
```

functions.php
```diff
@@ -41,16 +41,29 @@ function as_enqueue_async_action( $hook, $args = array(), $group = '', $unique =
...
+				'hook'      => $hook,
+				'arguments' => $args,
+				'group'     => $group,
+				'unique'    => $unique,
+				'priority'  => $priority,
+			)
+		);
+	} catch ( Exception $exception ) {
+		ActionScheduler::logger()->log(
+			0,
```
barryhughes: By logging to action zero, the logged message won't be surfaced anywhere (that I'm aware of).

- Should we support this with a change to the admin UI, to make it easier for site operators to find the problem?
- Or, as a minimum, support this change with notes in our docs?
- Or, perhaps consider using [error_log()](https://www.php.net/manual/en/function.error-log.php) (on the basis that the logged errors would be more findable).

functions.php
lsinger: `ActionScheduler::factory()->create()` throws at least one exception itself so I wouldn't want to change that behavior too much in case other parts depend on this behavior. 

If we move only the logging code there but not the exception catching we'd not yet have the exception details available ...? (Not sure if I understand correctly?)

I agree that the repetition isn't ideal. One other option I could see is adding a function to the factory that does the exception catching and the logging, allowing us to reduce duplication but at the same time keep `create`'s method signature. 🤔

functions.php
lsinger: Great catch. For now I'll go with 2 and 3, but longer term I think 1 would also be desirable.

lsinger: Opened an issue for 1 here: #1002.

functions.php
barryhughes: > ActionScheduler::factory()->create() throws at least one exception itself so I wouldn't want to change that behavior too much in case other parts depend on this behavior.

I don't think we have to? That is, if you're referring to [this exception](https://github.com/woocommerce/action-scheduler/blob/3.6.4/classes/ActionScheduler_ActionFactory.php#L310):

```php
throw new Exception( "Unknown action type '{$options['type']}' specified when trying to create an action for '{$options['hook']}'." );
```

We could leave that in place and only wrap the calls to the [store action methods](https://github.com/woocommerce/action-scheduler/blob/3.6.4/classes/ActionScheduler_ActionFactory.php#L310), something like:

```diff
-		return $options['unique'] ? $this->store_unique_action( $action ) : $this->store( $action );
+		try {
+			return $options['unique'] ? $this->store_unique_action( $action ) : $this->store( $action );
+		catch ( Exception $e ) {
+			// Add code to log the error here, then:
+			return 0;
+		}
```

barryhughes: > If we move only the logging code there but not the exception catching we'd not yet have the exception details available ...? (Not sure if I understand correctly?)

My description was a little vague, sorry. *If* we did this (and I was just noting it as a possibility, I much prefer the earlier option) we'd need to catch the exception first, to get the error details, but then re-throw it.

lsinger: @barryhughes I made the changes we discussed, so this is ready for another review. 👍

barryhughes (approved):
This is great, thank you.

- Confirmed a (potentially site-breaking) exception occurs without this change (when the priority column is missing).
- Confirmed an exception does not occur when this branch is checkout out, and that appropriate error messages are logged.

PR #1001, opened 2023-10-11
Release/3.6.4

alopezari:
This is the PR for the 3.6.4 release.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,12 @@
 *** Changelog ***
 
+= 3.6.4 - 2023-10-11 =
+* 3.6.3 release.
+* Fix option lock test.
+* Fix: Use orderby => 'none' when bulk cancelling actions.
+* Tweak - WP 6.3 compatibility.
+* Update PR unit tests matrix.
+
 = 3.6.3 - 2023-09-13 =
 * Use `_doing_it_wrong` in initialization check.
 
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,11 +1,8 @@
 *** Changelog ***
 
 = 3.6.4 - 2023-10-11 =
-* 3.6.3 release.
-* Fix option lock test.
-* Fix: Use orderby => 'none' when bulk cancelling actions.
-* Tweak - WP 6.3 compatibility.
-* Update PR unit tests matrix.
+* Performance improvements when bulk cancelling actions.
+* Dev-related fixes.
 
 = 3.6.3 - 2023-09-13 =
 * Use `_doing_it_wrong` in initialization check.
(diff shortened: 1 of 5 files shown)


changelog.txt
```diff
@@ -1,5 +1,12 @@
 *** Changelog ***
 
+= 3.6.4 - 2023-10-11 =
+* 3.6.3 release.
+* Fix option lock test.
+* Fix: Use orderby => 'none' when bulk cancelling actions.
+* Tweak - WP 6.3 compatibility.
+* Update PR unit tests matrix.
```
barryhughes: Not the reviewer, just passing through 👋🏼 ... but thinking we *could* condense this a little:

```suggestion
* Performance improvements when bulk cancelling actions.
* Dev-related fixes.
```

Here's my thinking:

- We probably shouldn't have *3.6.3 release* here in the list of 3.6.4 changes.
- I think *Tweak - WP 6.3 compatibility* and *Update PR unit tests matrix* are referencing the same thing, and they don't really impact users (hence my idea of combining with the *fix option lock test* entry, and reducing to just *dev-related fixes).*

Beyond this ... maybe we need to think about how we handle and communicate changelog needs for Action Scheduler. Our processes here are not as robust as in the WooCommerce mono-repo, and perhaps it's time to align them?

alopezari: The suggested changes sound good to me, @barryhughes, thanks for sharing! It's my first time committing to the Action Scheduler so I just stuck to the default changelog. If @Konamiman agrees with the suggested changes too, I'm happy to commit them 👍 

> Beyond this ... maybe we need to think about how we handle and communicate changelog needs for Action Scheduler. Our processes here are not as robust as in the WooCommerce mono-repo, and perhaps it's time to align them?

Sounds great too! This will streamline the release process - mostly for people new to the process - as it takes time to find out how to announce the changes 💯

Konamiman: Yes, I think these changes are good.

Konamiman (approved):

alopezari:
Thanks for approving this @Konamiman!

@barryhughes I committed your suggested changes, thanks!

Seems one of you needs to merge this PR as I don't have enough permission to do it.

readme.txt
```diff
@@ -45,6 +45,13 @@ Collaboration is cool. We'd love to work with you to improve Action Scheduler. [
 
 == Changelog ==
 
+= 3.6.4 - 2023-10-11 =
+* 3.6.3 release.
+* Fix option lock test.
+* Fix: Use orderby => 'none' when bulk cancelling actions.
+* Tweak - WP 6.3 compatibility.
+* Update PR unit tests matrix.
```
barryhughes:
```suggestion
* Performance improvements when bulk cancelling actions.
* Dev-related fixes.
```

☝🏼 Applying the same change as before.

PR #1004, opened 2023-10-23
Update unit tests for upcoming dependency version policy

lsinger:
This PR updates our CI to ensure unit tests run with the WordPress and PHP versions that we'll require in our upcoming L-2 dependency version policy. 

To test:
- ensure CI passes
- ensure the versions in the file conform to the L-2 policy -- also [see here](https://make.wordpress.org/core/handbook/references/php-compatibility-and-wordpress-versions/), we want to support WordPress latest minus 2 versions and the respective minimum PHP versions


coreymckrill (approved):
👍 Looks good to me, and the tests have all passed.

PR #1016, opened 2023-12-11
Try: Make code coverage work

lsinger:
I had to make a few changes to let us use a relatively current and, most importantly, stock phpunit. Other changes I implemented to enable the process suggested by Codecov. 

The one change that seems to break about 25 assertions is [using the @before and @after annotations instead of magic method names](https://github.com/woocommerce/action-scheduler/pull/1016/commits/9b76f9e4d4356e2e3ce4e54652590c9a2ed76dad) -- but my suspicion is that this is because those method now actually work. Previously we called `parent::setUp` even though the parent class was providing a `set_up` method ... and one missing the `@before` annotation at that. 

So the next step here is to see whether the failing assertions actually make sense (as in, whether our expectations are actually correct) or else what is causing the results to be different from what we expect. Maybe the annotated `set_up` and `tear_down` methods aren't running properly at all.

### DIFF
--- a/.github/workflows/pr-coverage.yml
+++ b/.github/workflows/pr-coverage.yml
@@ -34,7 +34,7 @@ jobs:
           key: ${{ runner.os }}-${{ hashFiles('./composer.lock') }}
 
       - name: Install dependencies
-        run: composer self-update && composer require --dev --no-update phpunit/phpunit:^9.5 && composer install && composer dump-autoload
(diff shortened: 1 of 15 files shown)


lsinger:
Pinging @barryhughes for awareness -- my suspicion right now is that not all `set_up` and `tear_down` methods are running as intended and possibly might never have. I need to focus back on other things for the time being so this PR here is a sort of save point.

barryhughes:
Good find!

Issue #1011, opened 2023-11-15
WordPress.org plugin review warning/issue about internationalization


remcotolsma:
We received the following feedback from the WordPress.org plugin review team (plugins@wordpress.org) about a new plugin in which we included this library:

> Hi!
> 
> Good work with the corrections!
> 
> ## Internationalization: Don't use variables or defines as text, context or text domain parameters.
> 
> Please **do not use variable names** for the text, context or text domain portion of a gettext function, all of them NEED to be strings, otherwise they could not be read by the translation parser.
> 
> For example, this is a bad call: `esc_html__( 'Translate me.' , $plugin_slug );` , the fix here would be like this `esc_html__( 'Translate me.' , 'plugin-slug' );`
> 
> This is also wrong: `esc_html__( 'Your city is ' . $city . '.' , 'plugin-slug' );` , the fix here would be using a placeholder and change it later using `printf` , like this:
> 
> ```php
> printf(
> 
>       /* translators: %s: Name of a city */
> 
>       esc_html__( 'Your city is %s.', 'plugin-slug' ),
> 
>       esc_html($city)
> 
> );
> ```
> 
> Also have in mind that your text domain has to be the same as your plugin's slug (also known as permalink).
> 
> The text domain also needs to be added to the plugin header. WordPress uses it to internationalize your plugin meta-data even when the plugin is disabled.
> 
> You can read https://developer.wordpress.org/plugins/internationalization/how-to-internationalize-your-plugin/#text-domains for more information.
> 
> Example(s) from your plugin:
> 
> ```
> packages/woocommerce/action-scheduler/classes/ActionScheduler_ListTable.php:237 _n(self::$time_periods[$time_period_index]['names'][0], self::$time_periods[$time_period_index]['names'][1], $periods_in_interval, 'pronamic-pay-with-mollie-for-woocommerce');
> ```
> 
> --
> WordPress Plugin Review Team | [plugins@wordpress.org](mailto:plugins@wordpress.org)
> https://make.wordpress.org/plugins/
> https://developer.wordpress.org/plugins/wordpress-org/detailed-plugin-guidelines/

We don't think this is actually an issue, but for future plugin reviews it might be useful to address this.

https://github.com/woocommerce/action-scheduler/blob/10c96381d2ae6162654b370cbb944ec6bb42dbc8/classes/ActionScheduler_ListTable.php#L237

[...]

PR #1026, opened 2024-01-25
No longer user variables in `_n()` translation function

remcotolsma:
Fixes issue:

- https://github.com/woocommerce/action-scheduler/issues/1011

### DIFF
--- a/classes/ActionScheduler_ListTable.php
+++ b/classes/ActionScheduler_ListTable.php
@@ -234,7 +234,7 @@ class ActionScheduler_ListTable extends ActionScheduler_Abstract_ListTable {
 				if ( ! empty( $output ) ) {
 					$output .= ' ';
 				}
-				$output .= sprintf( _n( self::$time_periods[ $time_period_index ]['names'][0], self::$time_periods[ $time_period_index ]['names'][1], $periods_in_interval, 'action-scheduler' ), $periods_in_interval );
+				$output .= sprintf( translate_nooped_plural( self::$time_periods[ $time_period_index ]['names'], $periods_in_interval, 'action-scheduler' ), $periods_in_interval );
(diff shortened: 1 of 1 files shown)


jorgeatorres (changes requested):
Hi @remcotolsma!

Thanks for your help with this one 💯. Much appreciated.
Your changes make sense, but I wonder if such a refactor is necessary. Given we were already using `_n_noop()` in theory the only thing we would've had to change was that `_n()` call inside of `human_interval()` to use `translate_nooped_plural()` instead.

So we would change this line:

https://github.com/woocommerce/action-scheduler/blob/2ca161ef7b199f0b690d90f91116b82e37fff3ad/classes/ActionScheduler_ListTable.php#L237

to

```
$output .= sprintf( translate_nooped_plural( self::$time_periods[ $time_period_index ]['names'], $periods_in_interval, 'action-scheduler' ), $periods_in_interval );
```

It should achieve the same thing in a compatible way and we wouldn't have to make any more changes.

Please let me know what your thoughts are and thanks again for taking a stab at this!

remcotolsma:
@jorgeatorres Thanks, my colleague @rvdsteege also suggested this to me yesterday. This indeed makes the PR a lot smaller, I had not interpreted your comment from Nov 16 correctly:

> Thanks for the report. We reviewed the code and we were in fact using `_n_noop()` (as we should) but then passed that string to `_n()` instead of `translate_nooped_plural()` as we should have. We'll take care of this in a future release. Thanks again!

_Originally posted by @jorgeatorres in https://github.com/woocommerce/action-scheduler/issues/1011#issuecomment-1814431802_

Apologies, I updated the PR: https://github.com/woocommerce/action-scheduler/pull/1026/commits/b36b2a0c73d30ec27e2a8ac98a8087cdbc684013.

jorgeatorres (approved):
LGTM. Thanks @remcotolsma! 🙇

PR #1038, opened 2024-02-27
update version in package-lock.json

lsinger:
The `3.7.2` branch has `version` set to `3.7.2.` only in the top-level `version` field ([see here](https://github.com/woocommerce/action-scheduler/blob/3.7.2/package-lock.json)), but not in the nested `action-scheduler` package. 

Running `npm install` in `trunk` updates that version, too. 

I'm not sure if this was done on purpose or not, so I'm putting out this PR to update the version if necessary. I'd especially like @nigeljamesstevenson to chime in in case there's something about the release process that I'm overlooking here. Thanks!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "action-scheduler",
-      "version": "3.7.0",
+      "version": "3.7.2",
       "license": "GPL-3.0+",
       "devDependencies": {
         "grunt": "1.5.3",


coreymckrill (approved):
I don't fully understand why the package-lock file would include a reference to itself, but it seems logical that the version should be updated there, too. Approving, but I'll leave it unmerged so @nigeljamesstevenson can also take a look. Maybe we need to update the docs for releasing this?

nigeljamesstevenson (approved):
Thanks for pointing this out @coreymckrill and @lsinger !

This 'reference to itself' was missed in the release process, but not intentionally.

I will update the process doc to ensure this gets updated in future - approving!

PR #1049, opened 2024-04-05
Bump "Tested up to" version to 6.5

jorgeatorres:
This was already done in the readme.txt, but now we're updating the plugin file to match that.


Stojdza (approved):
LGTM 👍

Issue #1059, opened 2024-05-14
Review and revise exception handling within Action Scheduler


alewolf:
There are quite a few occasions when Action Scheduler experiences an unexpected condition, and (mostly for safety) it stops. It communicates about this using exceptions, but many of these exceptions:

- Are not handled by Action Scheduler.
- The design does not make it easy for parent plugins to handle them.

In most cases, this is 'fine' in the sense that the queue is processed in requests that are independent of user-facing requests. However, they could impact other scheduled events when running in the context of WP Cron.

Additionally, there is the problem noted below which is that only one version of Action Scheduler can be active at any one time. That is, only one parent plugin 'wins' this particular race, and if things go wrong—irrespective of the underlying reason—the backtraces often lead users back to that plugin's support team.

Let's review and revise when we throw exceptions, and when we should add handling for them.

- We could consider logging deadlocks etc in Action Scheduler's own logging table (perhaps by referencing an `action_id` of `0` to indicate they are not related to a specific action). If we do this, we would also want to add:
    - A means of viewing any such global log entries.
    - Possibly, hooks that fire when these conditions are encountered.
- There are other cases where we bail out for safety, but possibly are being 'too safe'. For instance, when cleaning up expired actions: if the action being removed is concurrently removed by another process, we currently bail out of the clean-up process. This is probably overkill.
- Decisions like the above could also be made configurable/overrideable.

Original report follows...

---

I bundle the ActionScheduler version `3.7.4` in the [Pixel Manager for WooCommerce](http://wordpress.org/plugins/woocommerce-google-adwords-conversion-tracking-tag/). 

As of recently we started to receive critical error reports like this: 

```
2024-05-13T08:24:09+00:00 Critical Uncaught InvalidArgumentException: Unidentified action 19061390 in /nas/content/live/nck/wp-content/plugins/pixel-manager-pro-for-woocommerce/vendor/woocommerce/action-scheduler/classes/data-stores/ActionScheduler_DBStore.php:1086  
Additional context
{
    "error": {
        "type": 1,
        "file": "\/nas\/content\/live\/nck\/wp-content\/plugins\/pixel-manager-pro-for-woocommerce\/vendor\/woocommerce\/action-scheduler\/classes\/data-stores\/ActionScheduler_DBStore.php",
        "line": 1086
    },
    "backtrace": [
        "",
[...]

PR #1060, opened 2024-05-21
Improve the messaging in our unidentified action exceptions.

barryhughes:
We trigger `InvalidArgumentExceptions` from a number of locations, all with the same basic message:

```
Unidentified action <ID>
```

This is not particularly helpful, and conveys no information about what was happening at the time the exception was thrown. This change adds some additional context to a number of these exception messages.

Updates #1059 (but does not close).

---

### Testing instructions

This PR changes the text of various exception messages, but there are no *functional* changes to the way the library works. Manual testing will therefore have limited value, but I've still provided some basic instructions to see the change in action.

✍🏼 Note that you need to have access to the system/PHP error log, or some other way to view errors logged via `error_log()` calls:

1. Open **Tools ‣ Scheduled Actions**
2. Open the very same screen in a second browser tab.
3. Identify and check the box for the same action in both browser tabs.
4. Using the bulk actions, delete the action first in one tab and then the other.
5. You should see something similar to the following being logged:

```
Action Scheduler was unable to delete action 12345. Reason: Unidentified action 12345: we were unable to delete this action. It may may have been deleted by another process.
```

If that seems a little verbose, note that the first half (everything before `Reason:`) is added by the List Table implementation when it catches the exception.

### Changelog

> Fix - Provide more context when throwing `InvalidArgumentException` exceptions.

### DIFF
--- a/classes/data-stores/ActionScheduler_DBStore.php
+++ b/classes/data-stores/ActionScheduler_DBStore.php
@@ -650,7 +650,7 @@ AND `group_id` = %d
 		);
 		if ( false === $updated ) {
 			/* translators: %s: action ID */
-			throw new \InvalidArgumentException( sprintf( __( 'Unidentified action %s', 'action-scheduler' ), $action_id ) );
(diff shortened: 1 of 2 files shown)


jorgeatorres (approved):
LGTM. Thank you @barryhughes!

PR #1061, opened 2024-05-22
Merge Release/3.8.0 into trunk

rodelgc:


changelog.txt
```diff
@@ -1,5 +1,12 @@
 *** Changelog ***
 
+= 3.8.0 - 2024-05-22 =
+* Bump "Tested up to" version to 6.5.
+* Fix typos in perf.md.
+* Merge Release/3.7.4 into trunk.
+* Tweak - WP 6.5 compatibility.
+* Update minimum requirements for WordPress (and PHP). Fix test matrix.
```
barryhughes:
```suggestion
* Documentation - Fixed typos in perf.md.
* Update - We now require WordPress 6.3 or higher.
* Update - We now require PHP 7.0 or higher.
```

I wonder if this is a clearer summary?

- We don't need to tell people about the merge of the release branch in the changelog, it's not really a change from the perspective of our users.
- The bump in 'tested up to' also isn't really a change, and is any way communicated via the relevant readme field.
- Fixing the test matrix also isn't really a shipped change, it's something that really only matters here (in CI).
- If we make adjustments for the above, I figured we may as well break the requirement updates into two separate bullets for readability reasons.

I do realize the changelog is formed by an automation 🙌🏼 ... but I do think for it to be genuinely useful it could require a bit of finesse.

coreymckrill: I agree that this change makes the changelog more useful. Looks like we might want to go back and edit the entry for 3.7.3 as well, as it has some of the same entries.

readme.txt
```diff
@@ -47,6 +47,13 @@ Collaboration is cool. We'd love to work with you to improve Action Scheduler. [
 
 == Changelog ==
 
+= 3.8.0 - 2024-05-22 =
+* Bump "Tested up to" version to 6.5.
+* Fix typos in perf.md.
+* Merge Release/3.7.4 into trunk.
+* Tweak - WP 6.5 compatibility.
+* Update minimum requirements for WordPress (and PHP). Fix test matrix.
```
coreymckrill:
```suggestion
* Documentation - Fixed typos in perf.md.
* Update - We now require WordPress 6.3 or higher.
* Update - We now require PHP 7.0 or higher.
```

Updating to match the changelog.txt file

coreymckrill (approved):

PR #1071, opened 2024-06-19
WP-CLI

### LINKED ISSUE #255: WP CLI: Schedule new action(s)

crstauf:
Port of [crstauf/action-scheduler-cli](https://github.com/crstauf/action-scheduler-cli) into Action Scheduler.

Closes #255.

Hope I'm not disappointed.

### Changelog

> Add - A number of new WP CLI commands have been added, making it easier to manage actions in the terminal and from scripts.


classes/WP_CLI/Action/Create.php
```diff
@@ -0,0 +1,126 @@
...
+		$callback_args  = get_flag_value( $this->assoc_args, 'args', array() );
+		$group          = get_flag_value( $this->assoc_args, 'group', '' );
+		$interval       = absint( get_flag_value( $this->assoc_args, 'interval', 0 ) );
+		$cron           = get_flag_value( $this->assoc_args, 'cron', '' );
+
+		if ( ! empty( $callback_args ) ) {
+			$callback_args = json_decode( $callback_args, true );
+		}
+
+		$function_args = array(
```
vedanshujain: let's add support for `uniq` and `priority` param too

classes/WP_CLI/Action/Create.php
```diff
@@ -0,0 +1,126 @@
...
+
+		} else if ( ! empty( $cron ) ) { // Creating cron action.
+			$action_type = 'cron';
+			$function    = 'as_schedule_cron_action';
+		}
+
+		$function_args = array_values( array_filter( $function_args ) );
+
+		try {
+			$actions_added = call_user_func_array( $function, $function_args );
```
vedanshujain: this is not the number of actions added, but instead the action ID. I suppose there will always be only one action created in this command anyway. And would be `0` if there was no action created, we should also print_error when the returning ID is 0.

classes/WP_CLI/Action/Cancel.php
```diff
@@ -0,0 +1,79 @@
...
+
+		if ( $all ) {
+			$function_name = 'as_unschedule_all_actions';
+			$multiple      = true;
+		}
+
+		try {
+			call_user_func( $function_name, $hook, $callback_args, $group );
+		} catch ( \Exception $e ) {
+			$this->print_error( $e, $multiple );
```
vedanshujain: note that it won't throw an exception necessarily even if cancelling failed, there could be an error caught by the try_catch in `as_unschedule_action`. It would be more reliable to either 
1. check the logs for failed action, or
2. verify that the action is indeed cancelled, or
3. modify the `as_unschedule_action` to return the result value so that we can display it reliably.

classes/WP_CLI/Action/Delete.php
```diff
@@ -0,0 +1,90 @@
+<?php
+
+/**
+ * WP-CLI command: action-scheduler action delete
+ */
+class ActionScheduler_WPCLI_Action_Delete_Command extends ActionScheduler_WPCLI_Command {
```
vedanshujain: do you think if we should support a bulk delete option as well, for example, deleting by hook name? folks may want to use this command to control number of action, in which case deleting by hook would be more useful than deleting by id.

classes/WP_CLI/Action/Delete.php
```diff
@@ -0,0 +1,90 @@
...
+	 * @param string[]              $args       Positional arguments.
+	 * @param array<string, string> $assoc_args Keyed arguments.
+	 */
+	public function __construct( array $args, array $assoc_args ) {
+		parent::__construct( $args, $assoc_args );
+
+		$this->action_ids             = array_map( 'absint', $args );
+		$this->action_counts['total'] = count( $this->action_ids );
+
+		add_action( 'action_scheduler_deleted_action', array( $this, 'action__deleted' ) );
```
vedanshujain: we should have a check on whether `WP_CLI` is set before hooking this action.

classes/WP_CLI/Action/Get.php
```diff
@@ -0,0 +1,59 @@
...
+		$store     = \ActionScheduler::store();
+		$logger    = \ActionScheduler::logger();
+		$action    = $store->fetch_action( $action_id );
+
+		$only_logs   = ! empty( $this->assoc_args['field'] ) && 'log_entries' === $this->assoc_args['field'];
+		$log_entries = array();
+
+		foreach ( $logger->get_logs( $action_id ) as $log_entry ) {
+			if ( $only_logs ) {
+				WP_CLI::line( sprintf( '%s %s', $log_entry->get_date()->format( static::DATE_FORMAT ), $log_entry->get_message() ) );
```
vedanshujain: nice, this seems very useful!

classes/WP_CLI/Action/List.php
```diff
@@ -0,0 +1,107 @@
+<?php
+
+/**
+ * WP-CLI command: action-scheduler action list
+ */
+class ActionScheduler_WPCLI_Action_List_Command extends ActionScheduler_WPCLI_Command {
+
+	const PARAMETERS = array(
```
vedanshujain: we should mention defaults, especially for pagination params somewhere.

vedanshujain (commented):
Great work! I posted some initial thoughts as I work my way through the PR, let me know what you think.

classes/WP_CLI/Action/Cancel.php
crstauf: Going to rework this command a bit: realized in its current state actions could not be unscheduled by group only.

classes/WP_CLI/Action/Delete.php
crstauf: Such a task should already be doable with something like this:

```shell
wp action-scheduler action delete $( wp action-scheduler list --hook=action_hook --format=ids )
```

classes/WP_CLI/Action/Delete.php
crstauf: Because of the way the `action` subcommands are being loaded, they only load in a `WP_CLI` environment, but I will add a check to the abstract class for redundancy.

classes/WP_CLI/Action/List.php
crstauf:
```
$ wp help action-scheduler action list
...
SYNOPSIS

  wp action-scheduler action list [--<field>=<value>] [--field=<field>] [--fields=<fields>] [--format=<format>]

  Display actions based on all arguments supported by
  [as_get_scheduled_actions()][1].

  ---
  [1] https://actionscheduler.org/api/#function-reference--as_get_scheduled_actions
...
```

See https://github.com/crstauf/action-scheduler/blob/76e55cfd59282ab2a6a0f2303ff4df0e4dc1469d/classes/WP_CLI/ActionScheduler_WPCLI_Action_Command.php#L209-L213.

This follows the convention set by [the `wp post list` command](https://developer.wordpress.org/cli/commands/post/list/):

```
$ wp help post list
...
SYNOPSIS

  wp post list [--<field>=<value>] [--field=<field>] [--fields=<fields>] [--format=<format>]

  Display posts based on all arguments supported by [WP_Query()][1].
  Only shows post types marked as post by default.

  ---
  [1] https://developer.wordpress.org/reference/classes/wp_query/
...
```

vedanshujain:
The code looks good, going to test around these flows in general:

- [ ] Create - async/single/cron/recurring/error
- [ ] Generate - testing for bulk count
- [x] Get - error in action.
- [x] Run
- [x] List - listing in bulk
- [x] Cancel - check for error flows too
- [x] Delete - check for deleting in bulk + error flows

classes/WP_CLI/Action/Delete.php
```diff
@@ -0,0 +1,90 @@
...
+			sprintf(
+				/* translators: %d: number of actions to be deleted */
+				_n( 'Deleting %d action', 'Deleting %d actions', $this->action_counts['total'], 'action-scheduler' ),
+				number_format_i18n( $this->action_counts['total'] )
+			),
+			$this->action_counts['total']
+		);
+
+		foreach ( $this->action_ids as $action_id ) {
+			$store->delete_action( $action_id );
```
vedanshujain: Seems like we not handling failures gracefully, passing one invalid ID crashes the whole command. I think we can wrap this in a try catch, and then display the actions that we were not able to delete:

```
> wp action-scheduler action delete 1561 163333
Deleting 2 actions  50 % [========================================================>                                                        ] 0:00 / 0:00Fatal error: Uncaught InvalidArgumentException: Unidentified action 163333: we were unable to delete this action. It may may have been deleted by another process. in /usr/src/public_html/wp-content/plugins/action-scheduler/classes/data-stores/ActionScheduler_DBStore.php:746
Stack trace:
#0 /usr/src/public_html/wp-content/plugins/action-scheduler/classes/WP_CLI/Action/Delete.php(53): ActionScheduler_DBStore->delete_action(163333)
#1 /usr/src/public_html/wp-content/plugins/action-scheduler/classes/WP_CLI/ActionScheduler_WPCLI_Action_Command.php(119): ActionScheduler_WPCLI_Action_Delete_Command->execute()
#2 [internal function]: ActionScheduler_WPCLI_Action_Command->delete(Array, Array)
#3 phar:///usr/local/bin/wp/vendor/wp-cli/wp-cli/php/WP_CLI/Dispatcher/CommandFactory.php(100): call_user_func(Array, Array, Array)
#4 [internal function]: WP_CLI\Dispatcher\CommandFactory::WP_CLI\Dispatcher\{closure}(Array, Array)
#5 phar:///usr/local/bin/wp/vendor/wp-cli/wp-cli/php/WP_CLI/Dispatcher/Subcommand.php(491): call_user_func(Objec in /usr/src/public_html/wp-content/plugins/action-scheduler/classes/data-stores/ActionScheduler_DBStore.php on line 746
Error: There has been a critical error on this website.Learn more about troubleshooting WordPress. There has been a critical error on this website.
```

classes/WP_CLI/Action/Get.php
```diff
@@ -0,0 +1,59 @@
...
+	/**
+	 * Execute command.
+	 *
+	 * @return void
+	 */
+	public function execute() {
+		$action_id = $this->args[0];
+		$store     = \ActionScheduler::store();
+		$logger    = \ActionScheduler::logger();
+		$action    = $store->fetch_action( $action_id );
```
vedanshujain: Same as delete, this throws a fatal for an action that does not exist. Let's handle that gracefully like `wp post` does.

classes/WP_CLI/Action/List.php
vedanshujain: ok, makes sense. although one change that we should make is change the order to `desc` by default. otherwise I am just seeing list of old actions and it's not immediately helpful.

classes/WP_CLI/Action/Create.php
```diff
@@ -0,0 +1,135 @@
...
+			'hook'          => $hook,
+			'callback_args' => $callback_args,
+			'group'         => $group,
+			'unique'        => $unique,
+			'priority'      => $priority,
+		);
+
+		// Generate schedule start if appropriate.
+		if ( ! in_array( $schedule_start, static::ASYNC_OPTS, true ) ) {
+			$schedule_start         = as_get_datetime_object( $schedule_start );
```
vedanshujain: `as_get_datetime_object` should be wrapped in a try--catch, mentioned that we were not able to parse the date (and possibly displaying an example or two)

classes/WP_CLI/Action/Create.php
```diff
@@ -0,0 +1,135 @@
+<?php
+
+use function \WP_CLI\Utils\get_flag_value;
+
+/**
+ * WP-CLI command: action-scheduler action create
+ */
+class ActionScheduler_WPCLI_Action_Create_Command extends ActionScheduler_WPCLI_Command {
```
vedanshujain: I am getting some issues using this command, seems like we are somehow setting the `function_args` to an int

```
> wp action-scheduler action create woocommerce_cleanup_draft_orders now

Fatal error: Uncaught TypeError: Argument 2 passed to ActionScheduler_Action::__construct() must be of the type array, int given, called in /usr/src/public_html/wp-content/plugins/action-scheduler/classes/ActionScheduler_ActionFactory.php on line 314 and defined in /usr/src/public_html/wp-content/plugins/action-scheduler/classes/actions/ActionScheduler_Action.php:26
Stack trace:
#0 /usr/src/public_html/wp-content/plugins/action-scheduler/classes/ActionScheduler_ActionFactory.php(314): ActionScheduler_Action->__construct('woocommerce_cle...', 10, Object(ActionScheduler_NullSchedule), '')
#1 /usr/src/public_html/wp-content/plugins/action-scheduler/functions.php(51): ActionScheduler_ActionFactory->create(Array)
#2 /usr/src/public_html/wp-content/plugins/action-scheduler/classes/WP_CLI/Action/Create.php(82): as_enqueue_async_action('woocommerce_cle...', 10)
```

classes/WP_CLI/Action/Generate.php
```diff
@@ -0,0 +1,124 @@
...
+			'hook'          => $hook,
+			'callback_args' => $callback_args,
+			'group'         => $group,
+		);
+
+		$action_type   = 'single';
+		$function_args = array_values( array_filter( $function_args ) );
+
+		try {
+			$actions_added = $this->generate( ...$function_args );
```
vedanshujain: let's ensure that we have minimum count of `functions_args` as 4

vedanshujain (commented):
Posting feedback for some initial testing I was able to get around to today. Seems like we need to sprinkle in some more defensive checks

classes/WP_CLI/Action/List.php
crstauf: > although one change that we should make is change the order to desc by default.

I will not be making this change: the behavior should match what is documented, and if a user wants to retrieve newer actions, they can use the `date` and `date_compare` parameters.

classes/WP_CLI/Action/Create.php
```diff
@@ -0,0 +1,137 @@
...
+		} else if ( ! empty( $interval ) ) { // Creating recurring action.
+			$action_type = 'recurring';
+			$function    = 'as_schedule_recurring_action';
+
+		} else if ( ! empty( $cron ) ) { // Creating cron action.
+			$action_type = 'cron';
+			$function    = 'as_schedule_cron_action';
+		}
+
+		$function_args = array_values( $function_args );
```
vedanshujain: this is converting a 2d array to 1d, causing issues when we pass this in `call_user_func_array`. So this fails:
` wp action-scheduler action create test_action now`

vedanshujain: Seems like its the same issue -> https://github.com/woocommerce/action-scheduler/pull/1071#discussion_r1666636495

crstauf: Hmm, did not fail for me. What was the output?

crstauf:
@barryhughes Stalled here.

barryhughes:
Sorry, @crstauf. Please bear with us, as we've had a few other unexpected/unplanned pieces of work demanding our attention, unrelated to this project.

classes/WP_CLI/Action/Create.php
vedanshujain: not sure, seems to be working after recent changed, although I don't see anything on first glance that might be causing it.

classes/WP_CLI/Action/Run.php
```diff
@@ -0,0 +1,191 @@
...
+			if ( empty( $count ) ) {
+				continue;
+			}
+
+			/*
+			 * translators:
+			 * %1$d: count of actions evaluated.
+			 * %2$s: type of action evaluated.
+			 */
+			$format = _n( '%1$d action %2$s.', '%1$d actions %2$s.', $count, 'action-scheduler' ),
```
vedanshujain: syntax error:

```suggestion
			$format = _n( '%1$d action %2$s.', '%1$d actions %2$s.', $count, 'action-scheduler' );
```

vedanshujain (commented):
Looking good, although seems like something's going on with the cron action:

I tried scheduling a cron action like so:

```
> wp  action-scheduler action create woocommerce_cleanup_draft_orders_test now --cron="1 4 1 3 0"
Success: Async action (67684) scheduled.
```

Above message indicate that action was successfully created, but looking in the DB seems like it's created with a null schedule instead of the cron schedule. See:

Screenshot 2024-07-29 at 12 18 06 PM

I expect it to be similar of following command:

```
>  wp  eval  "as_schedule_cron_action(1, '1 4 1 3 0', 'woocommerce_cleanup_draft_orders_test' );"
```

Which schedules the action like this:

Screenshot 2024-07-29 at 12 20 51 PM


Maybe the values are not being passed as expected?

vedanshujain:
Thanks for the latest changes, looking pretty good now. I am planning to do bit more testing, but hopefully we will be able to merge this before end of this week.

classes/ActionScheduler_AdminView.php
```diff
@@ -228,6 +228,10 @@ public function add_help_tabs() {
 					'<h2>' . sprintf( __( 'About Action Scheduler %s', 'action-scheduler' ), $as_version ) . '</h2>' .
 					'<p>' .
 						__( 'Action Scheduler is a scalable, traceable job queue for background processing large sets of actions. Action Scheduler works by triggering an action hook to run at some time in the future. Scheduled actions can also be scheduled to run on a recurring schedule.', 'action-scheduler' ) .
+					'</p>' .
+					'<h3>' . __( 'WP CLI', 'action-scheduler' ) . '</h3>' .
+					'<p>' .
+						__( 'WP CLI commands are available: execute <code>wp help action-scheduler</code> for a list of available commands.' ) .
```
barryhughes: We should escape strings whenever possible, avoid exposing translators to markup, and use text domains. Here's an alternative:

```suggestion
						sprintf(
							/* translators: %1$s contains the WP CLI command 'wp help action-scheduler' (not translatable). */
							esc_html__( 'WP CLI commands are available: execute %1$s for a list of available commands.', 'action-scheduler' ),
							'<code>wp help action-scheduler</code>'
						) .
```

classes/ActionScheduler_AdminView.php
```diff
@@ -228,6 +228,10 @@ public function add_help_tabs() {
 					'<h2>' . sprintf( __( 'About Action Scheduler %s', 'action-scheduler' ), $as_version ) . '</h2>' .
 					'<p>' .
 						__( 'Action Scheduler is a scalable, traceable job queue for background processing large sets of actions. Action Scheduler works by triggering an action hook to run at some time in the future. Scheduled actions can also be scheduled to run on a recurring schedule.', 'action-scheduler' ) .
+					'</p>' .
+					'<h3>' . __( 'WP CLI', 'action-scheduler' ) . '</h3>' .
```
barryhughes:
```suggestion
					'<h3>' . esc_html__( 'WP CLI', 'action-scheduler' ) . '</h3>' .
```

classes/WP_CLI/Action/Cancel.php
```diff
@@ -0,0 +1,122 @@
...
+ */
+class ActionScheduler_WPCLI_Action_Cancel_Command extends ActionScheduler_WPCLI_Command {
+
+	/**
+	 * Execute command.
+	 *
+	 * @uses as_unschedule_action()
+	 * @uses as_unschedule_all_actions()
+	 * @uses $this->print_error()
+	 * @uses $this->print_success()
```
barryhughes: We don't typically use these. To reduce noise and keep things readable, I'd recommend we remove them here and elsewhere in this changeset.

```suggestion
```

📝 See [Inline Doc Standards](https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/) for WordPress.

classes/WP_CLI/Action/Cancel.php
```diff
@@ -0,0 +1,122 @@
...
+
+		$this->cancel_single( $hook, $callback_args, $group );
+	}
+
+	/**
+	 * Cancel single action.
+	 *
+	 * @param string $hook
+	 * @param array $callback_args
+	 * @param string $group
```
barryhughes: Let's make sure we run the linter and get a clean bill of health on all these new files. For instance, we need param comments (you can run `composer run phpcs` to list these things).

classes/WP_CLI/Action/Cancel.php
```diff
@@ -0,0 +1,122 @@
...
+	 * @return void
+	 */
+	protected function cancel_single( $hook, $callback_args, $group ) {
+		if ( empty( $hook ) ) {
+			\WP_CLI::error( __( 'Please specify hook of action to cancel.', 'action-scheduler' ) );
+		}
+
+		try {
+			$result = call_user_func( 'as_unschedule_action', $hook, $callback_args, $group );
+		} catch ( \Exception $e ) {
```
barryhughes: Though they don't harm, we're qualifying our use of various types like this one unnecessarily. Here and elsewhere, I think if we can avoid unnecessary symbols, it becomes a little easier to read.

```suggestion
		} catch ( Exception $e ) {
```

classes/WP_CLI/Action/Cancel.php
```diff
@@ -0,0 +1,122 @@
...
+	 * @param \Exception $e The error object.
+	 * @param bool       $multiple Boolean if multiple actions.
+	 * @throws \WP_CLI\ExitException
+	 * @return void
+	 */
+	protected function print_error( \Exception $e, $multiple ) {
+		\WP_CLI::error(
+			sprintf(
+				/* translators: %1$s: singular or plural %2$s: refers to the exception error message. */
+				__( 'There was an error cancelling the scheduled %1$s: %2$s', 'action-scheduler' ),
```
barryhughes: Preferred approach here (singular/plural versions) is to use `_n()`.

classes/WP_CLI/Action/Run.php
```diff
@@ -0,0 +1,191 @@
...
+	}
+
+	/**
+	 * Action: action_scheduler_execution_ignored
+	 *
+	 * @param int $action_id Action ID.
+	 * @uses \WP_CLI::debug()
+	 * @return void
+	 */
+	public function action__ignored( $action_id ) {
```
barryhughes: Curious, why the double underscore here (and other places)? 

Might a name like `count_ignored_actions` or perhaps `on_action_ignored` be slightly clearer?

classes/WP_CLI/Action/List.php
```diff
@@ -0,0 +1,135 @@
...
+
+		$formatter = new \WP_CLI\Formatter( $this->assoc_args, $fields );
+
+		$query_args = array_filter(
+			$this->assoc_args,
+			static function ( $key ) {
+				return in_array( $key, static::PARAMETERS, true );
+			},
+			ARRAY_FILTER_USE_KEY
+		);
```
barryhughes: A bit shorter and possibly more efficient to do either:

```suggestion
		$query_args = wp_array_slice_assoc( $this->assoc_args, static::PARAMETERS );
```

Or:

```suggestion
		$query_args = array_intersect_key( $assoc_args, array_flip( static::PARAMETERS ) );
```

classes/WP_CLI/Action/List.php
```diff
@@ -0,0 +1,135 @@
...
+
+		if ( ! empty( $query_args['args'] ) ) {
+			$query_args['args'] = json_decode( $query_args['args'], true );
+		}
+
+		switch ( $formatter->format ) {
+
+			case 'ids':
+				$actions = as_get_scheduled_actions( $query_args, 'ids' );
```
barryhughes: Optional thing, we could consider moving this to right before we enter the `switch` (instead of repeating it at the start of each `case`).

classes/WP_CLI/Action/Next.php
```diff
@@ -0,0 +1,62 @@
...
+		$params['status'] = ActionScheduler_Store::STATUS_PENDING;
+		WP_CLI::debug( 'ActionScheduler()::store()->query_action( ' . var_export( $params, true ) . ' )' );
+
+		$action_id = ActionScheduler::store()->query_action( $params );
+		if ( $action_id ) {
+			echo $action_id;
+			return;
+		}
+
+		WP_CLI::warning( 'No matching next action.' );
```
barryhughes: Translate.

```suggestion
		WP_CLI::warning( __( 'No matching next action.', 'action-scheduler' ) );
```

classes/WP_CLI/Action/List.php
barryhughes: Thinking on it some more, I wonder if we should even have this check/filter.

- Should we subsequently add support for new query args, this means we need updates in two different places.
- If a new or custom data store supports additional query args of its own, they won't be supported through this interface.

classes/WP_CLI/ActionScheduler_WPCLI_System_Command.php
```diff
@@ -0,0 +1,213 @@
...
+/**
+ * System info WP-CLI commands for Action Scheduler.
+ */
+class ActionScheduler_WPCLI_System_Command {
+
+	/**
+	 * Construct.
+	 */
+	public function __construct() {
+		$this->store = \ActionScheduler::store();
```
barryhughes: Let's declare this property before using it.

classes/WP_CLI/ActionScheduler_WPCLI_System_Command.php
```diff
@@ -0,0 +1,213 @@
...
+				'newest' => $oldest_and_newest[ $status ]['newest'],
+			);
+		}
+
+		$formatter = new \WP_CLI\Formatter( $assoc_args, array( 'status', 'count', 'oldest', 'newest' ) );
+		$formatter->display_items( $rows );
+	}
+
+	/**
+	 * Get latest or all system versions.
```
barryhughes: I wonder if we can make the intent a little clearer?

```suggestion
	 * Display the active Action Scheduler version, or list all registered versions.
```

classes/WP_CLI/ActionScheduler_WPCLI_System_Command.php
```diff
@@ -0,0 +1,213 @@
...
+				);
+			}
+
+			uksort( $rows, 'version_compare' );
+
+			$formatter = new \WP_CLI\Formatter( $assoc_args, array( 'version', 'callback' ) );
+			$formatter->display_items( $rows );
+
+			return;
+		}
```
barryhughes: Currently the output is something like this:

```
+---------+-------------------------------------------+
| version | callback                                  |
+---------+-------------------------------------------+
| 3.6.0   | action_scheduler_initialize_3_dot_6_dot_0 |
| 3.8.1   | action_scheduler_initialize_3_dot_8_dot_1 |
+---------+-------------------------------------------+
```

Could it be useful to indicate which is active/loaded?

```
+---------+-------------------------------------------+--------+
| version | callback                                  | active |
+---------+-------------------------------------------+--------+
| 3.6.0   | action_scheduler_initialize_3_dot_6_dot_0 | no     |
| 3.8.1   | action_scheduler_initialize_3_dot_8_dot_1 | yes    |
+---------+-------------------------------------------+--------+
```

We generally hope it will be the latest version, but that is not always so (and my assumption is that this sub-command exists to aid debugging).

classes/WP_CLI/Action/Cancel.php
```diff
@@ -0,0 +1,122 @@
...
+	 * @param string $group
+	 * @return void
+	 */
+	protected function cancel_single( $hook, $callback_args, $group ) {
+		if ( empty( $hook ) ) {
+			\WP_CLI::error( __( 'Please specify hook of action to cancel.', 'action-scheduler' ) );
+		}
+
+		try {
+			$result = call_user_func( 'as_unschedule_action', $hook, $callback_args, $group );
```
barryhughes: Since we know the function we're calling, why don't we call it directly (why `call_user_func()`)?

classes/WP_CLI/ActionScheduler_WPCLI_Action_Command.php
```diff
@@ -0,0 +1,352 @@
...
+	 * : The ID of the action to get.
+	 * ---
+	 * default: 0
+	 * ---
+	 *
+	 * [--field=<field>]
+	 * : Instead of returning the whole action, returns the value of a single field.
+	 *
+	 * [--fields=<fields>]
+	 * : Limit the output to specific fields. Defaults to all fields.
```
barryhughes:
```suggestion
	 * : Limit the output to the specified fields (should be comma-separated). Defaults to all fields.
```

classes/WP_CLI/Action/Get.php
```diff
@@ -0,0 +1,72 @@
...
+
+		if ( is_a( $action, ActionScheduler_NullAction::class ) ) {
+			/* translators: %d is action ID. */
+			\WP_CLI::error( sprintf( esc_html__( 'Unable to retrieve action %d.', 'action-scheduler' ), $action_id ) );
+		}
+
+		$only_logs   = ! empty( $this->assoc_args['field'] ) && 'log_entries' === $this->assoc_args['field'];
```
barryhughes: Should this also check against `--fields` (ie, if `--fields=log_entries` was passed)?

classes/WP_CLI/Action/Next.php
```diff
@@ -0,0 +1,62 @@
...
+
+		$params = array(
+			'hook'    => $hook,
+			'orderby' => 'date',
+			'order'   => 'ASC',
+			'group'   => $group,
+		);
+
+		if ( is_array( $args ) ) {
+			$params['args'] = $args;
```
barryhughes: Should this be:

```suggestion
		if ( is_array( $this->assoc_args['args'] ) ) {
			$params['args'] = $this->assoc_args['args'];
```

classes/abstracts/ActionScheduler_WPCLI_Command.php
```diff
@@ -0,0 +1,98 @@
...
+	protected $assoc_args;
+
+	/**
+	 * Construct.
+	 *
+	 * @param string[]              $args       Positional arguments.
+	 * @param array<string, string> $assoc_args Keyed arguments.
+	 */
+	public function __construct( array $args, array $assoc_args ) {
+		if ( ! defined( 'WP_CLI' ) || ! constant( 'WP_CLI' ) ) {
```
barryhughes, line 32:
```suggestion
		if ( ! defined( 'WP_CLI' ) || ! WP_CLI ) ) {
```

Not sure we need the call to `constant`.

classes/abstracts/ActionScheduler_WPCLI_Command.php
```diff
@@ -0,0 +1,98 @@
...
+			/* translators: %s php class name */
+			throw new Exception( sprintf( __( 'The %s class can only be run within WP CLI.', 'action-scheduler' ), __CLASS__ ) );
```
barryhughes: `__CLASS__` will be the class we are in (the super-class), not the invoked class. Do we want:

```suggestion
			throw new Exception( sprintf( __( 'The %s class can only be run within WP CLI.', 'action-scheduler' ), get_class( $this ) ) );
```

classes/abstracts/ActionScheduler_WPCLI_Command.php
```diff
@@ -0,0 +1,98 @@
...
+	 * @return string
+	 */
+	protected function get_recurrence( $action ) {
+		$schedule = $action->get_schedule();
+		if ( $schedule->is_recurring() ) {
+			$recurrence = $schedule->get_recurrence();
+
+			if ( is_numeric( $recurrence ) ) {
+				/* translators: %s: time interval */
+				return sprintf( __( 'Every %s', 'action-scheduler' ), self::human_interval( $recurrence ) );
```
barryhughes: Can we check on `human_interval`, is this referenced correctly (is this intended to access the `ActionScheduler_ListTable` method)?

classes/abstracts/ActionScheduler_WPCLI_Command.php
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	/**
+	 * Transforms arguments with '__' from CSV into expected arrays.
+	 *
+	 * @see \WP_CLI\CommandWithDBObject::process_csv_arguments_to_arrays()
+	 * @link https://github.com/wp-cli/entity-command/blob/6e0e77a297eefa3329b94bec16c15cf7528d343f/src/WP_CLI/CommandWithDBObject.php
+	 * @return void
+	 */
+	protected function process_csv_arguments_to_arrays() {
```
barryhughes, line 75: When do we invoke this? I see it is called from `ActionScheduler_WPCLI_Action_List_Command::execute()`, to be clear, but can you give me an example of a command invocation that requires it?

barryhughes (commented):
It's a big change, thank you for working on it. I added some additional comments, but there's a lot to test here so it may take some more time.

Some high-level thoughts:

- In future, it may be better to break changes like this down into smaller chunks.
- We're adding lots of code here, but no unit tests have been added.
- We're also adding lots of functionality here, but there are no additions to `docs/wp-cli.md` or other doc files. This is something for us to think about.
- We're introducing a new naming convention for class-files. I'm not opposed to it, but am also envisioning a future where we want to introduce PSR-4 conventions, and that could leave us with three different naming conventions.

classes/WP_CLI/Action/Run.php
crstauf: It's my style, and I find it to be perfectly clear and not overly verbose, especially coupled with the phpdoc.

classes/WP_CLI/Action/Next.php
crstauf: Core WP CLI commands do not use translation functions:

- https://github.com/wp-cli/entity-command/blob/c270cc9a2367cb8f5845f26a6b5e203397c91392/src/Site_Command.php#L262
- https://github.com/wp-cli/entity-command/blob/c270cc9a2367cb8f5845f26a6b5e203397c91392/src/Post_Command.php#L186

classes/WP_CLI/ActionScheduler_WPCLI_System_Command.php
crstauf: Oh whoa. What would the situation be where the latest isn't used?

classes/WP_CLI/Action/Run.php
barryhughes: That's fine, it's not a criticism. However, it's a little atypical for *this* project and our team, though (and we need to strike a balance, and some consistency).

classes/WP_CLI/Action/Get.php
crstauf: Oh, that's interesting. Why have both `--field` and `--fields` flags? 🤔

classes/abstracts/ActionScheduler_WPCLI_Command.php
crstauf: Not needed, but more explicit.

classes/WP_CLI/Action/Cancel.php
crstauf: Going to leave this for now, for consistency (other AS classes use `\Exception`).

classes/WP_CLI/Action/Cancel.php
crstauf: There's no numbers available or needed here, which is how `_n()` works. Sure we could force `_n()`, but it's use would be silly in comparison.

classes/WP_CLI/Action/List.php
crstauf: The `default` case is different.

classes/WP_CLI/Action/Cancel.php
crstauf: I'm not sure. 🤷‍♂️ 😉

classes/abstracts/ActionScheduler_WPCLI_Command.php
crstauf: I suspect I initially copy and pasted it from `ActionScheduler_ListTable` for some reason, but it does not appear that the `get_recurrence()` function is being called.

classes/abstracts/ActionScheduler_WPCLI_Command.php
crstauf: I cannot give an example, I only noticed that [several of the core commands](https://github.com/search?q=repo%3Awp-cli%2Fentity-command%20process_csv_arguments_to_arrays&type=code) implement it.

crstauf:
> * We're adding lots of code here, but no unit tests have been added.

@barryhughes I don't see any tests for WP CLI commands. Can you please direct me to those so I can reference them? I've never written tests for WP CLI commands before.

classes/ActionScheduler_AdminView.php
```diff
@@ -228,6 +228,14 @@ public function add_help_tabs() {
...
+					'<p>' .
+						sprintf(
+							/* translators: %1$s is WP CLI command (not translatable) */
+							__( 'WP CLI commands are available: execute %1$s for a list of available commands.', 'action-scheduler' ),
```
barryhughes:
```suggestion
							esc_html__( 'WP CLI commands are available: execute %1$s for a list of available commands.', 'action-scheduler' ),
```

barryhughes:
> I don't see any tests for WP CLI commands.

Yep, doesn't look like we have coverage there (so far). We can stick a pin in this aspect while we think about how best to approach it.

crstauf:
@barryhughes @vedanshujain I am unable to reproduce the errors in the unit tests.

CleanShot 2024-08-09 at 21 06 46@2x

barryhughes:
Re-running them, in case there was some flakiness.

crstauf:
> Re-running them, in case there was some flakiness.

@barryhughes Thanks: looks like all checks passed. 👍

barryhughes:
One thing I'm a bit wary of here is introducing yet another convention in terms of how we name and structure class files. "Traditionally", class files in Action Scheduler seem to be nested in directories representing different sub-systems, and are always prefixed with `ActionScheduler_*`. The class name and the class file are basically 1:1 in these cases. Here's a good example of that:

`classes/data-stores/ActionScheduler_DBLogger.php` contains `ActionScheduler_DBLogger`

We also have started to use namespaces (broadly following PSR-4 conventions), with `classes/` acting as the base directory for the `Action_Scheduler` namespace: 

`classes/WP_CLI/Migration_Command.php` contains `Action_Scheduler\WP_CLI\Migration_Command`

This change introduces yet another convention that doesn't align with either of those:

`classes/WP_CLI/Action/Delete.php` contains `ActionScheduler_WPCLI_Action_Delete_Command`

I think it would be best to align with either of the existing styles, rather than introducing a third approach. Probably, namespacing as in the second example is the best way forward at this time.

crstauf:
@barryhughes I'll need to leave that to you, as I'm not interested in fighting with the autoloader for 10 hours.

classes/abstracts/ActionScheduler_WPCLI_Command.php
barryhughes: OK, but if in our case we aren't accepting CSV-formatted arguments, why add it?

classes/WP_CLI/Action/Cancel.php
barryhughes: Why would it be silly? In English, only the `%1$s` placeholder needs to change. In other languages, preceding words like `scheduled` may need to change, according to their grammar rules. This doesn't allow for that.

classes/abstracts/ActionScheduler_WPCLI_Command.php
crstauf: Core's documentation doesn't mention accepting CSV-formatted arguments either, so it's likely they've anticipated some situation that should be accounted for.

If you really want to get rid of it, please do so.

classes/WP_CLI/ActionScheduler_WPCLI_System_Command.php
barryhughes: If several versions of Action Scheduler are present and one is 3.2.0 or earlier then, depending on how it is initialized, it may take precedence over later versions [(related)](https://developer.woocommerce.com/2021/10/12/best-practices-for-deconflicting-different-versions-of-action-scheduler/). Not just hypothetical, we came across one such situation quite recently, impacting WooCommerce.

barryhughes: That said, the proposed change could be left until a future iteration.

classes/WP_CLI/Action/Cancel_Command.php
```diff
@@ -0,0 +1,120 @@
...
+	 * @param string $group The group the job is assigned to.
+	 * @return void
+	 */
+	protected function cancel_all( $hook, $callback_args, $group ) {
+		if ( empty( $hook ) && empty( $group ) ) {
+			\WP_CLI::error( __( 'Please specify hook and/or group of actions to cancel.', 'action-scheduler' ) );
+		}
+
+		try {
+			$result = call_user_func( 'as_unschedule_all_actions', $hook, $callback_args, $group );
```
barryhughes: I won't mark each one, but as in an earlier thread we should use direct invocation unless there is a reason for `call_user_func` (there doesn't seem to be, because we're passing a string literal and the same set of args each time).

classes/WP_CLI/ActionScheduler_WPCLI_System_Command.php
crstauf: Interesting. The `active` column was added.

classes/WP_CLI/Action/Get.php
barryhughes: Sticking to one may be cleaner, and would avoid doubling up on checks here. Instead of supporting [`--field` and `--fields`](https://github.com/woocommerce/action-scheduler/pull/1071/files#diff-ab2b213587b3f26cf90e53ed137c529583b32100193ae7b256a3c58dbb3d2232R184-R188), we could just rely on `--fields`.

classes/WP_CLI/Action/Cancel_Command.php
crstauf: This was the last one.

classes/WP_CLI/Action/Get.php
crstauf: Sorry, my question was meant for core. Because core implements both, I'm going to keep both.

classes/abstracts/ActionScheduler_WPCLI_Command.php
barryhughes: Asking out of interest, you mean it makes it more explicit that `WP_CLI` is a constant and not a `$string`? I'd have thought the primary use-case for `constant` is cases where the constant name is unknown (held in a variable) or is formed dynamically.

crstauf: That is likely the intent of `constant()`, but I've found that sometimes PHPStan gets mad at `WP_CLI`, and wrapping it in `constant()` fixes it. Always wondered if it's because there's both a constant and class with the same name.

classes/abstracts/ActionScheduler.php
```diff
@@ -297,9 +300,11 @@ protected static function is_class_migration( $class ) {
...
 	protected static function is_class_cli( $class ) {
 		static $cli_segments = array(
-			'QueueRunner' => true,
-			'Command'     => true,
-			'ProgressBar' => true,
+			'QueueRunner'                          => true,
+			'Command'                              => true,
+			'ProgressBar'                          => true,
+			'ActionScheduler_WPCLI_Action_Command' => true,
+			'ActionScheduler_WPCLI_System_Command' => true,
```
barryhughes: Do these need to be updated following the switch to namespacing?

classes/WP_CLI/ActionScheduler_WPCLI_System_Command.php
```diff
@@ -0,0 +1,220 @@
+<?php
+
+// phpcs:disable WordPress.Security.EscapeOutput.OutputNotEscaped -- Escaping output is not necessary in WP CLI.
+
+use function \WP_CLI\Utils\get_flag_value;
+
+/**
+ * System info WP-CLI commands for Action Scheduler.
+ */
+class ActionScheduler_WPCLI_System_Command {
```
barryhughes: Is it intentional to use this class naming convention here, whereas sibling class `Action_Command` is now namespaced?

classes/abstracts/ActionScheduler.php
crstauf: @barryhughes I absolutely hate this loading mechanism. May need your help to figure this out.

classes/WP_CLI/ActionScheduler_WPCLI_System_Command.php
crstauf: It was, but I'll change it.

classes/abstracts/ActionScheduler.php
crstauf: Okay, think I got it.

barryhughes: Thanks. Yep, it's a little tricky to navigate.

crstauf:
@barryhughes Stalled again. 😞

barryhughes:
Yep, I'm sorry. We've been focusing on other tasks and responsibilities that have taken priority. This is looking great, though, so bear with us a little longer. *Probably,* if any further changes are needed, they'll be minor/small pieces of clean-up, and so we can likely take care of those if required (and can similarly update the branch if it falls out of sync in the meantime).

barryhughes (approved):
Thanks again, @crstauf. Merging this, and it should therefore become generally available in the next release.

crstauf:
Yay! Thanks, @barryhughes!

PR #1077, opened 2024-08-01
`has_pending_actions_due()` performance improvements

### LINKED ISSUE #1036: `has_pending_actions_due()` performance improvements
`has_pending_actions_due()` should
1) pass `'claimed'` as `true` to the query to avoid triggering runners when there are no unclaimed tasks
2) pass `'per_page'` with value 1 since it's sufficient to get 1 result (instead of the default 5)
3) pass `'orderby'` with value `'none'` to avoid unnecessary overhead of sorting the actions table
4) pass 2nd parameter `'count'` as that's faster than SELECT and sufficient for this use case

crstauf:
Closes #1036.

At time of creation, does not include `'claimed' => true`: unclear if this is appropriate or not.


classes/abstracts/ActionScheduler_Store.php
```diff
@@ -449,10 +449,11 @@ public function get_status_labels() {
 	 */
 	public function has_pending_actions_due() {
 		$pending_actions = $this->query_actions( array(
-			'date'    => as_get_datetime_object(),
-			'status'  => self::STATUS_PENDING,
-			'orderby' => 'none',
-		) );
+			'per_page' => 1,
+			'date'     => as_get_datetime_object(),
+			'status'   => ActionScheduler_Store::STATUS_PENDING,
```
coreymckrill:
```suggestion
			'status'   => self::STATUS_PENDING,
```

Looks like this should be updated after [this other change](https://github.com/woocommerce/action-scheduler/commit/b69152038912f0a62de173d3d13587969174fcbc) was committed.

coreymckrill (commented):
Thanks for the PR @crstauf !

I'm trying to think of the best way to test if `has_pending_actions_due()` still works as expected with these changes. Any ideas?

> At time of creation, does not include 'claimed' => true: unclear if this is appropriate or not.

We discussed this a bit more and decided it's probably better to not include this right now, so this is fine.

crstauf:
> I'm trying to think of the best way to test if `has_pending_actions_due()` still works as expected with these changes. Any ideas?

@coreymckrill Kinda surprised that there are no unit tests for it, but looking at it, the changes would not impact the result, only the performance.

classes/abstracts/ActionScheduler_Store.php
```diff
@@ -448,11 +448,15 @@ public function get_status_labels() {
...
+			array(
+				'per_page' => 1,
+				'date'     => as_get_datetime_object(),
+				'status'   => self::STATUS_PENDING,
+				'orderby'  => 'none',
+			),
+			'count'
+		);
 
 		return ! empty( $pending_actions );
```
coreymckrill, line 479: Even though this still works, it might be worth updating since `query_actions` returns a numeric string now that it's set to `count`, instead of an array. Maybe something like

```
	return absint( $pending_actions ) > 0;
```

What do you think?

coreymckrill (commented):
@crstauf Fair point, adding unit tests for this in https://github.com/woocommerce/action-scheduler/pull/1092

classes/abstracts/ActionScheduler_Store.php
crstauf: I think it's developer preference. Both approaches amount to the same thing: `absint()` will return `0` on anything that `empty()` would return as `true`, so doesn't really matter.

coreymckrill (approved):
Unit tests added in #1092 are passing here, so calling this good 👍

PR #1078, opened 2024-08-01
Batch size should be adjustable during processing

### LINKED ISSUE #1037: Batch size should be adjustable during processing
https://github.com/woocommerce/action-scheduler/blob/trunk/classes/ActionScheduler_QueueRunner.php#L137

This line `$batch_size = apply_filters( 'action_scheduler_queue_runner_batch_size', 25 );` should be moved 1 line down to be inside the `do {` to allow adjusting the batch size during runtime.

Why?
This is useful to speed up the processing of actions with multiple concurrent batches. 
e.g. we know on average we can process 100 actions in a batch, which is why we have set the batch size to 100. 
Additionally, we have 5 concurrent batches allowed.
On average there are 250 pending actions at every given time, which means there are 2 continous running processes (2*100) + 1 one one/off (to process the 50 surplus)
However, when the batch size can be dynamically adjusted in the loop, we can use all 5 workers (with 50 each initially bc 250 / 5, which then goes down to ~25-30 since the average number of pending actions goes down since we process more), which means actions are processed faster and the number of pending actions goes down too.

Now you could ask: why not just reduce the batch size to 50 in the first place: when the number of actions in the queue increases during high load periods (e.g. lots of orders during the evening), a higher batch size is necessary to allow for faster processing.

crstauf:
Closes #1037.

### DIFF
--- a/classes/ActionScheduler_QueueRunner.php
+++ b/classes/ActionScheduler_QueueRunner.php
@@ -134,8 +134,8 @@ class ActionScheduler_QueueRunner extends ActionScheduler_Abstract_QueueRunner {
 
 		$this->processed_actions_count = 0;
 		if ( false === $this->has_maximum_concurrent_batches() ) {
-			$batch_size = apply_filters( 'action_scheduler_queue_runner_batch_size', 25 );
 			do {
+				$batch_size                     = apply_filters( 'action_scheduler_queue_runner_batch_size', 25 );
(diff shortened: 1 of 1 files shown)


vedanshujain (approved):
LGTM!

Tested by overwriting to 1 and working as expected. I don't see any concerns for existing usage by this change. 

Screenshot 2024-08-21 at 5 37 51 PM

PR #1083, opened 2024-08-05
Email notification for flooded past-due actions

### LINKED ISSUE #1080: [Enhancement]: Add option to enable notification email when ActionScheduler past-due queue hits a certain threshold
### Describe the solution you'd like

When the ActionScheduler had a backed up past-due queue, there is a notification at the top of wp-admin. That is great, but you have to log in and see it. What would be great is a notification email to the admin once it hits a certain theshold, perhaps one that is user-defined. And a user-defined frequency of hourly/daily/weekly. 

### Describe alternatives you've considered

This can be added via custom code, but it would be a great product improvement!

### Additional context

This came from a client using Woo Subscriptions, because a flood of traffic and a slow webhook endpoint caused the ActionScheduler to back up with over 32K actions over the course of 24 hours.

crstauf:
Closes #1080.

I suspect there will be a lot of opinions and compatibility issues here, so providing only as a stepping stone for the Action Scheduler team, in an attempt to help them get it out quickly (we'll see).


Konamiman (changes requested):
Hi @crstauf, thanks for your contribution. I have a couple of comments:

* The methods that are removed in `classes/ActionScheduler_AdminView.php` shouldn't actually be removed because they are public and the removal would break backwards compatibility. Could they be marked as deprecated, in the same way as they are in `deprecated/ActionScheduler_AdminView_Deprecated.php` below?
* While receiving the notifications via email instead of as an admin notice seems indeed like it will be useful for most people, some might prefer to continue seeing these in the admin area. So maybe a setting could be provided to choose whether to use the new email notifications, the old warning in the admin area, or both.
* Related to the above, a setting to provide an email address for the notifications (defaulting to the default admin email) would be useful.

crstauf:
> Could they be marked as deprecated, in the same way as they are in `deprecated/ActionScheduler_AdminView_Deprecated.php` below?

@Konamiman I do not understand. I have defined the functions that were deprecated into the class that handles deprecated functions. Please advise.

> some might prefer to continue seeing these in the admin area

This PR does not prevent the notice from appearing in the admin view. It seems you have not thoroughly reviewed or tested this PR?

Will add some appropriate filters.

Konamiman:
> I have defined the functions that were deprecated into the class that handles deprecated functions.

Correct. I hadn't noticed that `ActionScheduler_AdminView` is actually extending `ActionScheduler_AdminView_Deprecated`.

> It seems you have not thoroughly reviewed or tested this PR?

Indeed I haven't tested the PR yet, I was waiting to your answer to my comment before doing that.

> Will add some appropriate filters.

Perfect, thank you. Please ping me once you do that and then I'll do all the testing.

crstauf:
@barryhughes Stalled.

classes/ActionScheduler_PastDueMonitor.php
```diff
@@ -0,0 +1,363 @@
...
+		}
+	}
+
+	/**
+	 * Check if notification method enabled.
+	 *
+	 * @param string $method
+	 * @return bool
+	 */
+	protected function notify( $method ) {
```
Konamiman, line 136: This method name is a bit misleading, it seems to suggest that a notification will be sent. What about something like `notify_methods_has` instead?

classes/ActionScheduler_PastDueMonitor.php
```diff
@@ -0,0 +1,363 @@
...
+			);
+		}
+
+		// Allow site devs to specify notification preference.
+		$methods = apply_filters( 'action_scheduler_pastdue_monitor_notify', $methods );
+
+		// Support for scalar and bool values (ex: `__return_false`).
+		if ( is_scalar( $methods ) || is_bool( $methods ) ) {
+			return boolval( $methods ) ? $all : array();
+		}
```
Konamiman, line 181: If the idea is that the filter yielding `__return_false` would cause an empty array to be returned, this won't work, since `boolval('__return_false')` evaluates to true. Adding the following before this block will solve that, and as a bonus it will add support for filters providing a callable that returns an array:

```php
if ( is_callable( $methods ) ) {
  $methods = $methods();
}
```

classes/ActionScheduler_PastDueMonitor.php
```diff
@@ -0,0 +1,363 @@
...
+
+		return $methods;
+	}
+
+	/**
+	 * Check if threshold for past-due actions is flooded.
+	 *
+	 * @return bool
+	 */
+	protected function flooded() {
```
Konamiman, line 199: Suggested rename: `threshold_is_flooded`

classes/ActionScheduler_PastDueMonitor.php
```diff
@@ -0,0 +1,363 @@
...
+		// Facilitate third-parties to evaluate and print notices.
+		do_action( 'action_scheduler_pastdue_actions_extra_notices', $this->query_args );
+	}
+
+	/**
+	 * Message for admin notice and email notice.
+	 *
+	 * @return string
+	 */
+	protected function message() {
```
Konamiman, line 350: Suggested rename: `message_contents`

classes/ActionScheduler_PastDueMonitor.php
```diff
@@ -0,0 +1,363 @@
...
+		if ( has_filter( 'action_scheduler_check_pastdue_actions' ) ) {
+			$methods['admin'] = apply_filters_deprecated(
+				'action_scheduler_check_pastdue_actions',
+				array( $methods['admin'] ),
+				'', // todo: add version number
+				'action_scheduler_pastdue_monitor_notify'
+			);
+		}
+
+		// Allow site devs to specify notification preference.
```
Konamiman, line 165:
```suggestion
		// Allow site devs to specify the active notification methods.
		// Default: array( 'admin' => true, 'email' => true )
		// The filter can return:
		// - An array.
		// - A boolean, true = default array, false = empty array.
		// - A callback that evaluates to one of the above.
```

Konamiman (changes requested):
Hi @crstauf, sorry for the delay in this second review. I have requested a few changes.

classes/ActionScheduler_PastDueMonitor.php
crstauf: This code is to support the filter:

```php
add_filter( 'action_scheduler_pastdue_monitor_notify', '__return_false' );
```

crstauf:
Added documentation, did not make name change requests.

classes/ActionScheduler_PastDueMonitor.php
Konamiman: Yes, I was confused, a snippet like `add_filter( 'action_scheduler_pastdue_monitor_notify', '__return_false' );` will cause `$methods` to become `false` and thus your code would work.

However I've discussed this with my team and we think that adding support for booleans adds unnecessary complexity. Supporting arrays should be enough, client code willing to suppress all the notifications could do just `add_filter( 'action_scheduler_pastdue_monitor_notify', function($methods) { return array(); } );` and you would be able to remove the entire "Support for scalar and bool values" block. Would you mind implementing this change please?

crstauf: You truly think three lines of code is complex?

Konamiman: While this change in particular is not terribly complex, everything adds up, and the simpler we can keep the code without removing essential functionality, the better.

crstauf: Oh! Well that's great, because this code is essential functionality: makes it easy for devs to turn off the notifications, regardless of any notifications added later.

kalessil:
Hey @crstauf :wave: I just got a review request for this PR, and given the last update's date, I'm wondering how we should proceed. Do you plan to move forward with it?

crstauf:
@kalessil Lol I was right!

I'm swamped with client work right now, so the AS team will need to resolve the conflicts if there's interest in merging this into the plugin.

It's unfortunate it took 10 months for it to receive attention...

kalessil:
> @kalessil Lol I was right!
> It's unfortunate it took 10 months for it to receive attention...

Sorry for that. It's been a challenging environment for the industry lately, which has affected our capacity allocation as well.

> I'm swamped with client work right now, so the AS team will need to resolve the conflicts if there's interest in merging this into the plugin.

It would be me, but I'm in the same boat as you - in my case capacity is allocated to release automation and performance.
Let's abandon this PR until one of us can revisit it with quicker review and update loops.

PR #1084, opened 2024-08-08
Missing parameter on apply_filters( 'pre_as_enqueue_async_action')

### LINKED ISSUE #1082: Missing parameter on apply_filters( 'pre_as_enqueue_async_action')
The hook `apply_filters( 'pre_as_enqueue_async_action')` is missing the `$unique` parameter, which is needed on `ActionScheduler::factory()->create()`.

The `$unique` value is available in the same scope, as it is received as a parameter in the `as_enqueue_async_action()` function.

Location in the code:
https://github.com/woocommerce/action-scheduler/blob/cc041aa3cba238f1ad883e0ada833f0795ba3072/functions.php#L39

crstauf:
Makes it possible for `pre_as_enqueue_async_action` callbacks to see if the action about to be created is considered unique.

Fixes #1082.

### Testing instructions

Add the following to a mu-plugin file, such as `wp-content/mu-plugins/test-pr-1084.php`:

```php
<?php

add_filter( 'pre_as_enqueue_async_action', function ( $null, $hook, $args, $group, $priority, $unique ) {
	var_dump( $null, $hook, $args, $group, $priority, $unique );
}, 10, 6 );
```

Now create a new async action via WP CLI:

```sh
wp eval "as_enqueue_async_action( 'foo', [ 'bar' => 'baz' ], 'foobar', true, 100 );"
```

You should see output similar to the following (whitespace added for readability):

```
./wp-content/mu-plugins/test-pr-1084.php.php:4:
NULL

./wp-content/mu-plugins/test-pr-1084.php.php:4:
string(3) "foo"

./wp-content/mu-plugins/test-pr-1084.php.php:4:
array(1) {
  'bar' =>
  string(3) "baz"
}

./wp-content/mu-plugins/test-pr-1084.php.php:4:
string(6) "foobar"

./wp-content/mu-plugins/test-pr-1084.php.php:4:
int(100)

./wp-content/mu-plugins/test-pr-1084.php.php:4:
bool(true)
```

In particular, we expect to see `(bool) true` for that final entry.

### Changelog

> Fix - Allows `pre_as_enqueue_async_action` callbacks to see if the action about to be scheduled should be unique.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -35,8 +35,9 @@ function as_enqueue_async_action( $hook, $args = array(), $group = '', $unique =
 	 * @param array    $args       Action arguments.
 	 * @param string   $group      Action group.
 	 * @param int      $priority   Action priority.
+	 * @param bool     $unique     Unique action.
 	 */
-	$pre = apply_filters( 'pre_as_enqueue_async_action', null, $hook, $args, $group, $priority );
+	$pre = apply_filters( 'pre_as_enqueue_async_action', null, $hook, $args, $group, $priority, $unique );
 	if ( null !== $pre ) {
 		return is_int( $pre ) ? $pre : 0;
 	}


crstauf:
@barryhughes All the failed checks are with Composer. Can you please fix and/or re-run?

barryhughes:
Re-running again. Not sure why we're having such a spate of fails, but at least those I sampled are clearly unrelated to the change at hand:

```
Error: file_put_contents(/home/runner/work/action-scheduler/action-scheduler/vendor/bin/phpunit): failed to open stream: No such file or directory
```

crstauf:
> Re-running again. Not sure why we're having such a spate of fails, but at least those I sampled are clearly unrelated to the change at hand

Thanks, @barryhughes.

barryhughes (approved):
Thanks, LGTM!

- Added testing instructions and verified the new behavior.
- No back-compat concerns, since we don't consume this filter ourselves (we've previously had cases where third party code independently triggers the same hook, with the original number of parameters, which can potentially lead to problems).
- CI test fails are sporadic and pretty clearly unrelated.

Issue #971, opened 2023-08-01
[Task] Codebase should pass linting checks


eddr:
Improve coding standards across Action Scheduler. We should pass our own linting tests, and ideally those should intersect with the WPCS checks implemented by the WooCommerce Marketplace.

- This accommodates a scenario where a plugin wishes to be listed in the marketplace, but also functions in a standalone capacity (ie, it cannot depend on the version of Action Scheduler bundled with WooCommerce). See discussion in the comments for more context.
- Irrespective of the above, this should result in improved code hygiene for Action Scheduler and less noise when linting locally.
- This is probably best handled incrementally; I advise against anyone submitting a monolithic fixes-everything style PR.

Additionally, as described [in issues like this one](https://github.com/woocommerce/action-scheduler/issues/1068), we should aim to get Action Scheduler's code to a point where it meets the expectations of the W.org plugin review team.

---

Original title: 

> Action Scheduler shows phpcs errors using wpcs : can't submit new extension to WC marketplace

Original description:

> Hi and thanks for excellent and useful plugin!
> 
> I'm facing issues when trying to submit a new extension to the WC marketplace. 
> The plugin uses Action Scheduler library and it causes the automatic tests to not pass WC own phpcs checks
>
> Any clue what can be done?
>
> Thanks

barryhughes:
Internal discussion: p1690905319416299-slack-C3L7CFBRV

barryhughes:
@eddr since you are creating a WooCommerce extension, you can presumably depend on WooCommerce being present (or can test for it). WooCommerce includes its own copy of Action Scheduler, and it should therefore be possible to use that from your extension (or else you can use `WC()->queue()` which effectively wraps it).

I acknowledge this comes with some trade-offs (ie, you may not be able to take advantage of the very latest functionality)—but it is probably the most practical path forward at present.

eddr:
Hi! Thank you very much!

I am aware that the AS included in WC, yes
The biggest problem for me is that the extension in mind should work even if WC if not active..

barryhughes:
Gotcha ... in that case, I don't have an easy workaround I can share. I've updated the issue title and description, though, to create a task for cleaning things up within the Action Scheduler codebase (on our end, it is not likely something we can prioritize straight away, however).

eddr:
Thanks

I really think it's mostly a problem with the WC extension submitting system

remcotolsma:
[...]

PR #1085, opened 2024-08-10
PHPCS fixes: `Squiz`

crstauf:
Eliminates smells from the `Squiz` standard.

### DIFF
--- a/classes/abstracts/ActionScheduler.php
+++ b/classes/abstracts/ActionScheduler.php
@@ -50,7 +50,7 @@ abstract class ActionScheduler {
 	public static function plugin_path( $path ) {
 		$base = dirname(self::$plugin_file);
 		if ( $path ) {
-			return trailingslashit($base).$path;
+			return trailingslashit($base) . $path;
 		} else {
 			return untrailingslashit($base);
 		}
--- a/classes/abstracts/ActionScheduler.php
+++ b/classes/abstracts/ActionScheduler.php
@@ -309,11 +309,11 @@ abstract class ActionScheduler {
 	}
 
 	final public function __clone() {
-		trigger_error("Singleton. No cloning allowed!", E_USER_ERROR);
+		trigger_error('Singleton. No cloning allowed!', E_USER_ERROR);
 	}
 
 	final public function __wakeup() {
-		trigger_error("Singleton. No serialization allowed!", E_USER_ERROR);
+		trigger_error('Singleton. No serialization allowed!', E_USER_ERROR);
 	}
 
 	final private function __construct() {}
--- a/classes/abstracts/ActionScheduler.php
+++ b/classes/abstracts/ActionScheduler.php
@@ -151,7 +151,7 @@ abstract class ActionScheduler {
 		// Ensure initialization on plugin activation.
 		if ( ! did_action( 'init' ) ) {
 			// phpcs:ignore Squiz.PHP.CommentedOutCode
-			add_action( 'init', array( $admin_view, 'init' ), 0, 0 ); // run before $store::init()
+			add_action( 'init', array( $admin_view, 'init' ), 0, 0 ); // run before $store::init().
 			add_action( 'init', array( $store, 'init' ), 1, 0 );
 			add_action( 'init', array( $logger, 'init' ), 1, 0 );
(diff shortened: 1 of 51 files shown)


crstauf:
@barryhughes More Composer failures.

classes/abstracts/ActionScheduler.php
```diff
@@ -149,7 +173,8 @@ public static function init( $plugin_file ) {
 
 		// Ensure initialization on plugin activation.
 		if ( ! did_action( 'init' ) ) {
-			add_action( 'init', array( $admin_view, 'init' ), 0, 0 ); // run before $store::init()
+			// phpcs:ignore Squiz.PHP.CommentedOutCode
```
barryhughes, line 176: Is this line required (which commented out code does it refer to)?

```suggestion
```

crstauf: `// run before $store::init().`

classes/data-stores/ActionScheduler_wpPostStore_PostStatusRegistrar.php
```diff
@@ -5,6 +5,10 @@
  * @codeCoverageIgnore
  */
 class ActionScheduler_wpPostStore_PostStatusRegistrar {
+
+	/**
+	 * Registrar.
```
barryhughes, line 10: Bit of a nit, doesn't matter too much, but why *registrar* vs *register* (the latter matches the method name)?

classes/schema/ActionScheduler_StoreSchema.php
```diff
@@ -33,10 +36,16 @@ public function init() {
...
+	/**
+	 * Get table definition.
+	 *
+	 * @param string $table Table name.
+	 */
 	protected function get_table_definition( $table ) {
 		global $wpdb;
 		$table_name       = $wpdb->$table;
 		$charset_collate  = $wpdb->get_charset_collate();
+		// phpcs:ignore Squiz.PHP.CommentedOutCode
```
barryhughes, line 48: Similar to an earlier question: is this required/what does it refer to?

```suggestion
```

crstauf: `// @see wp_get_db_schema()`

classes/data-stores/ActionScheduler_wpPostStore_PostStatusRegistrar.php
crstauf: Matches the class name.

classes/abstracts/ActionScheduler.php
barryhughes: Interesting. It isn't triggered for me (ie, I don't see an error for that rule, if the `phpcs:ignore` is removed).

barryhughes: (Though I would see it if I commented out the full line...) are you invoking phpcs via Composer (are you using the project's PHPCS)?

classes/schema/ActionScheduler_StoreSchema.php
barryhughes: I guess `// @see wp_get_db_schema()` on the next line, yet this rule violation is not reported when I run PHPCS (assuming that ignore statement isn't present).

barryhughes (approved):
Thanks! I have a couple of questions, but they are non-blocking (though I'm curious about the differences we are possibly seeing, in relation to commented out code).

We seem to be having some issues with CI right now, so bear with me while I look at that.

Otherwise, do you have a preference in terms of when this is merged (is it fine from your perspective if we merge this next, relative to your other PHPCS-related PR)?

crstauf:
> Otherwise, do you have a preference in terms of when this is merged (is it fine from your perspective if we merge this next, relative to your other PHPCS-related PR)?

@barryhughes I think merging it before the PHPCS fixes makes sense.

crstauf:
> Otherwise, do you have a preference in terms of when this is merged (is it fine from your perspective if we merge this next, relative to your other PHPCS-related PR)?

@barryhughes Yes, fine to do Squiz before WP/WC.

barryhughes:
- Merged via https://github.com/woocommerce/action-scheduler/pull/1087
- Same set of commits.

PR #1086, opened 2024-08-13
Set option value to `no` rather than deleting

crstauf:
> I’m tuning my page load performance. I’ve noticed that on every page there is an individual call to `get_option` for `as_has_wp_comment_logs` – rather than this option being loaded as part of `wp_load_alloptions`.
> 
> I think the problem is that, by default, there is no entry in the database options table for `as_has_wp_comment_logs` so it cannot be autoloaded. If there was an entry in the options table for `as_has_wp_comment_logs` with value “no” and `autoload=yes`, I think that would improve page load performance for most people using your plugin.

https://wordpress.org/support/topic/as_has_wp_comment_logs-loading/


classes/ActionScheduler_WPCommentCleaner.php
```diff
@@ -66,13 +66,17 @@ public static function has_logs() {
...
 		if ( (bool) get_comments( array( 'type' => ActionScheduler_wpCommentLogger::TYPE, 'number' => 1, 'fields' => 'ids' ) ) ) {
-			update_option( self::$has_logs_option_key, 'yes' );
+			$has_logs = 'yes';
 
 			if ( ! as_next_scheduled_action( self::$cleanup_hook ) ) {
 				as_schedule_single_action( gmdate( 'U' ) + ( 6 * MONTH_IN_SECONDS ), self::$cleanup_hook );
 			}
 		}
+
+		update_option( self::$has_logs_option_key, $has_logs );
```
coreymckrill:
```suggestion
		update_option( self::$has_logs_option_key, $has_logs, true );
```

Since this option gets queried on every page load, I'm thinking the autoload should be explicitly set to true. [Currently, WP heuristics would presumably determine this to be true anyway](https://make.wordpress.org/core/2024/06/18/options-api-disabling-autoload-for-large-options/), but that could change in the future.

classes/ActionScheduler_WPCommentCleaner.php
```diff
@@ -81,7 +85,7 @@ public static function maybe_schedule_cleanup() {
 	public static function delete_all_action_comments() {
 		global $wpdb;
 		$wpdb->delete( $wpdb->comments, array( 'comment_type' => ActionScheduler_wpCommentLogger::TYPE, 'comment_agent' => ActionScheduler_wpCommentLogger::AGENT ) );
-		delete_option( self::$has_logs_option_key );
+		update_option( self::$has_logs_option_key, 'no' );
```
coreymckrill:
```suggestion
		update_option( self::$has_logs_option_key, 'no', true );
```

Same as above comment.

crstauf:
@barryhughes Recently #1106, a small PR, was merged into `trunk`. This too is a small PR. What can I do to get preferential attention? Apparently @coreymckrill has disappeared, and his PR to add unit tests fails because of #1094.

coreymckrill (approved):

PR #1088, opened 2024-08-15
Update required PHP and platform PHP.

barryhughes:
- Our `composer.json` and the currently committed `composer.lock` are out of sync.
- We can't update various dependencies, because we (were) setting the platform PHP version to 5.6.
- We now expect and require PHP 7.0 as a minimum, so this should be reflected in these files.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -7,7 +7,7 @@
   "prefer-stable": true,
   "minimum-stability": "dev",
   "require": {
-    "php": ">=5.6"
+    "php": ">=7.0"
   },
   "require-dev": {
     "phpunit/phpunit": "^7.5",
@@ -20,7 +20,7 @@
       "dealerdirect/phpcodesniffer-composer-installer": true
     },
     "platform": {
-      "php": "5.6"
+      "php": "7.1"
     }
   },
   "archive": {
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "48f51bf1cc7f0010853b7c39db39a2db",
+    "content-hash": "81fd3951ccfed0a222e6f4b949998df1",
     "packages": [],
     "packages-dev": [
         {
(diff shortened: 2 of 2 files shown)


barryhughes:
Part of the motivation here is to resolve a class of CI errors that are blocking various tests.

coreymckrill:
Given that WordPress itself [requires](https://wordpress.org/about/requirements/) a minimum of PHP 7.2, could we bump these numbers further up?

coreymckrill:
Er, our support policy here is WP - 2, right? WordPress supported PHP 7.0 until 6.6, so I guess if we want to peg to WP's minimum requirements, we can't bump past 7.0 quite yet.

composer.json
```diff
@@ -20,7 +20,7 @@
       "dealerdirect/phpcodesniffer-composer-installer": true
     },
     "platform": {
-      "php": "5.6"
+      "php": "7.1"
```
coreymckrill, line 23: Why would this not be the same value as `require`?

barryhughes: To satisfy the constraints of our current PHPUnit requirement (PHPUnit 7.5.20 asks for PHP 7.1+). In CI this isn't an issue, because if we are in a part of the test matrix specifying PHP 7.0 then we detect that and install PHPUnit 5.7 instead.

barryhughes: [(Reference)](https://github.com/woocommerce/action-scheduler/blob/3.8.1/.github/workflows/pr-unit-tests.yml#L57-L80)

barryhughes:
> Er, our support policy here is WP - 2, right?

Yep, L-2, so we generally need to match WP 6.4 in terms of minimum supported PHP.

composer.json
barryhughes: You can possibly get a better sense of what's happening if you checkout the current trunk branch, remove the Composer lock file, and run `composer i`.

coreymckrill: Ah, I see!

```
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Root composer.json requires woocommerce/woocommerce-sniffs 0.1.0 -> satisfiable by woocommerce/woocommerce-sniffs[0.1.0].
    - woocommerce/woocommerce-sniffs 0.1.0 requires php >=7.0 -> your php version (5.6; overridden via config.platform, actual: 8.3.9) does not satisfy that requirement.
  Problem 2
    - phpunit/phpunit[7.5.0, ..., 7.5.20] require php ^7.1 -> your php version (5.6; overridden via config.platform, actual: 8.3.9) does not satisfy that requirement.
    - Root composer.json requires phpunit/phpunit ^7.5 -> satisfiable by phpunit/phpunit[7.5.0, ..., 7.5.20].
```

coreymckrill (approved):

PR #1089, opened 2024-08-16
PHPCS fixes: `PSR2`, `PSR12`, `PEAR`

crstauf:
Eliminates most smells from the `PSR2`, `PSR12`, and `PEAR` standards. Some rules excluded.


crstauf:
@barryhughes Stalled.

PR #1090, opened 2024-08-16
PHPCS fixes: `PHPCompatibility`, `Generic`

crstauf:
Eliminates most smells from the `PHPCompatibility` and `Generic` standards.

### DIFF
--- a/classes/ActionScheduler_ListTable.php
+++ b/classes/ActionScheduler_ListTable.php
@@ -2,6 +2,7 @@
 
 /**
  * Implements the admin view of the actions.
+ *
  * @codeCoverageIgnore
  */
 class ActionScheduler_ListTable extends ActionScheduler_Abstract_ListTable {
@@ -50,8 +51,7 @@ class ActionScheduler_ListTable extends ActionScheduler_Abstract_ListTable {
 
 	/**
 	 * Bulk actions. The key of the array is the method name of the implementation:
-	 *
-	 *     bulk_<key>(array $ids, string $sql_in).
+	 * bulk_<key>(array $ids, string $sql_in).
 	 *
 	 * See the comments in the parent class for further details
 	 *
--- a/classes/ActionScheduler_ListTable.php
+++ b/classes/ActionScheduler_ListTable.php
@@ -362,7 +362,7 @@ class ActionScheduler_ListTable extends ActionScheduler_Abstract_ListTable {
 				if ( ! in_array( $wpdb->prefix . $table_name, $found_tables ) ) {
 					$this->admin_notices[] = array(
 						'class'   => 'error',
-						'message' => __( 'It appears one or more database tables were missing. Attempting to re-create the missing table(s).' , 'action-scheduler' ),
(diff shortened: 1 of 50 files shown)


crstauf:
@barryhughes Composer failures again.

classes/ActionScheduler_ListTable.php
```diff
@@ -50,8 +51,7 @@ class ActionScheduler_ListTable extends ActionScheduler_Abstract_ListTable {
 
 	/**
 	 * Bulk actions. The key of the array is the method name of the implementation:
-	 *
-	 *     bulk_<key>(array $ids, string $sql_in).
+	 * bulk_<key>(array $ids, string $sql_in).
```
barryhughes, line 54: This seems less readable. I think we would prefer:

```php
	/**
	 * Map of bulk actions.
	 *
	 * The key of the array is the method name of the implementation:
	 *
	 *     bulk_<key>(array $ids, string $sql_in)

File: classes/WP_CLI/ActionScheduler_WPCLI_Clean_Command.php [hunk 2]
```diff
@@ -81,7 +81,7 @@ public function clean( $args, $assoc_args ) {
 	 *
 	 * @param int $batches_processed Number of batches processed.
 	 */
-	protected function print_total_batches( int $batches_processed ) {
+	protected function print_total_batches( $batches_processed ) {
```
barryhughes, line 84: We don't want to remove these. We'll [update the PHPCS config](https://github.com/woocommerce/action-scheduler/pull/1091) instead.

classes/WP_CLI/ActionScheduler_WPCLI_Scheduler_command.php
```diff
@@ -134,7 +134,7 @@ public function run( $args, $assoc_args ) {
 	 *
 	 * @return array
 	 */
-	private function parse_comma_separated_string( $string ): array {
+	private function parse_comma_separated_string( $string ) {
```
barryhughes, line 137: Similar to an earlier comment, let's not remove these.

barryhughes (changes requested):
Thanks! Requested a couple of small changes.

classes/WP_CLI/ActionScheduler_WPCLI_Clean_Command.php
crstauf: @barryhughes Well that's unfortunate, to have wasted that time. Will kill this PR and branch and start again.

barryhughes: Why kill the branch? There are, as far as I could see, only two instances of type hints being removed.

barryhughes: Do bear in mind that these overlapping PRs with large changesets also come at a cost to us.

crstauf: Because it's likely there are other rules for PHP 7.

crstauf:
New PR: https://github.com/woocommerce/action-scheduler/pull/1093.

PR #1092, opened 2024-08-24
Add unit tests for the has_pending_actions_due method

coreymckrill:
In #1077 we're making some performance improvements to the `has_pending_actions_due` method, but there are no unit tests to protect against regressions, so this is simply adding some. Because the tests are in the abstract class, they will get run for each data store type.


naman03malhotra (approved):
Looks, good to me, thanks for adding missing tests. Some CI tests are failing though.

crstauf:
More progress: keep it going! :beers:

PR #1095, opened 2024-08-26
PHPCS fixes: `PHPCompatibility`, `Generic`

crstauf:
Re-open #1093.
See #971.

### DIFF
--- a/classes/ActionScheduler_ActionClaim.php
+++ b/classes/ActionScheduler_ActionClaim.php
@@ -16,7 +16,7 @@ class ActionScheduler_ActionClaim {
 	 * @param int[]  $action_ids Action IDs.
 	 */
 	public function __construct( $id, array $action_ids ) {
-		$this->id = $id;
+		$this->id         = $id;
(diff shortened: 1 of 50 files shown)


crstauf:
@barryhughes Stalled.

crstauf:
@barryhughes What's the reason for the hold up here? This brings the files into compliance with your own standards. Why the delay?

barryhughes:
@crstauf I've requested a reviewer. 

Partly, we're just stretched pretty thin (not your problem, but it's a current reality). Additionally, these are very large changesets, each of which touches pretty much the entire codebase—and there are several of them—which makes review and testing a pretty time-consuming affair. Working class-by-class or system-by-system, rather than by groups of PHPCS rules, would probably have made this easier and faster.

We're also keen to close https://github.com/woocommerce/action-scheduler/issues/971, but it also is not a high priority. To be clear, that's not me saying it is unimportant, however it is also true this does not solve a functional problem or active bug.

PR #1097, opened 2024-09-04
Update minimum versions and test matrix.

barryhughes:
Some repo maintenance.

🧪 **Test matrix revised** → The versions we should now test against are 7.0 and 7.4, plus 8.0 and 8.3. 

📝 **Readme and plugin header** → Our policy for supporting WordPress is [L-2](https://developer.woocommerce.com/2023/10/24/action-scheduler-to-adopt-l-2-dependency-version-policy/), which means WP 6.4 is our current minimum. We've also had real world exposure and testing (via WooCommerce) under WP 6.6, so I've bumped *'tested up to'.*

### DIFF
--- a/action-scheduler.php
+++ b/action-scheduler.php
@@ -7,8 +7,8 @@
  * Author URI: https://automattic.com/
  * Version: 3.8.1
  * License: GPLv3
- * Requires at least: 6.2
- * Tested up to: 6.5
+ * Requires at least: 6.4
+ * Tested up to: 6.6
  * Requires PHP: 5.6
  *
  * Copyright 2019 Automattic, Inc.  (https://automattic.com/contact/)
--- a/action-scheduler.php
+++ b/action-scheduler.php
@@ -9,7 +9,7 @@
  * License: GPLv3
  * Requires at least: 6.4
  * Tested up to: 6.6
- * Requires PHP: 5.6
+ * Requires PHP: 7.0
  *
  * Copyright 2019 Automattic, Inc.  (https://automattic.com/contact/)
  *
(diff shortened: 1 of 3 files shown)


coreymckrill:
> The only remaining PHP 7.x version we need to test is 7.4.

If we still support WP 6.4 right now, don't we still need to support its minimum PHP version (7.0)?

barryhughes:
> If we still support WP 6.4 right now, don't we still need to support its minimum PHP version (7.0)?

Yes, you're right. Corrected the description, and committed https://github.com/woocommerce/action-scheduler/pull/1097/commits/e69fd0393198ea7acf73bdfbe67a1a730f0abe4a.

action-scheduler.php
```diff
@@ -7,9 +7,9 @@
  * Author URI: https://automattic.com/
  * Version: 3.8.1
  * License: GPLv3
- * Requires at least: 6.2
- * Tested up to: 6.5
- * Requires PHP: 5.6
+ * Requires at least: 6.4
+ * Tested up to: 6.6
+ * Requires PHP: 7.0
```
naman03malhotra, line 12: That's a fair leap in php version, though I suppose PHP > 7 would be widely used.

image
https://stitcher.io/blog/php-version-stats-january-2024

naman03malhotra (approved):
Looks good, thanks for this PR.

action-scheduler.php
barryhughes: Yes, though worth noting that the change in PHP requirements is not new as of this PR—it's just that when we [originally bumped those requirements](https://github.com/woocommerce/action-scheduler/pull/1056) we missed this particular spot (it is however already reflected in the [plugin's `readme.txt`](https://github.com/woocommerce/action-scheduler/blob/3.8.1/readme.txt#L8)).

Also, because of our [L-2 policy](https://developer.woocommerce.com/2023/10/24/action-scheduler-to-adopt-l-2-dependency-version-policy/), our minimum required version of WordPress should now be 6.4 (which itself [does not support PHP 5.6](https://make.wordpress.org/core/handbook/references/php-compatibility-and-wordpress-versions/)).

PR #1098, opened 2024-09-04
Dev/phpunit setup

barryhughes:


barryhughes:
Progress so far:

- Switched to PhpUnit 9 for all runs (now viable, because PHP 7.4 and WP 6.4 are our baseline ... so, for now, we [don't need to switch between different versions](https://make.wordpress.org/core/handbook/references/phpunit-compatibility-and-wordpress-versions/)).
- Resolved an unreported exception stemming from a [use of `assertContains()`](https://github.com/woocommerce/action-scheduler/pull/1098/commits/602cfdff9af627918a52a9318558fc982c483b7c) that was valid in older versions of PhpUnit, but is no longer valid under PhpUnit 9.
- In CI: it seems there is a problem with the selected version of PhpUnit (that includes a class named `Match`, which is now a reserved name). I don't see this locally, though, so either CI caching is causing an issue or something else is out of sync.
- Locally: two failures (see below) still to resolve:

```
ActionScheduler_HybridStore_Test::test_actions_are_migrated_on_query
Failed asserting that an array contains '1001'
```

☝🏼 This surfaces twice (because the test suite runs twice). It's unclear why it surfaces only under PHP 8.0+, and I should dig into the actual test to ensure it is indeed a reasonable assertion to make.

barryhughes:
Correction to the above, looks like we need PhpUnit 6 for PHP 7.0, and we need to exclude WP 6.6+ from that set of tests [[*]](https://make.wordpress.org/core/handbook/references/phpunit-compatibility-and-wordpress-versions/).

barryhughes:
Closing/deleting, we'll create a fresh PR to handle these updates soon (and can refer back to this to see what works and doesn't work).

PR #773, opened 2021-10-15
Allow filtering of ORDER BY on claim_actions

glagonikas:
Conditionally resolves #530 

Using `ORDER BY` makes the query excessively slow and leads to deadlocks. Without ORDER BY, the query will still pick the oldest actions (in most cases) because it'll order them based on the action_id which is sequential. For most cases, dropping the ORDER BY will have no negative impact, but will greatly improve performance and reduce deadlocks.

Stores that do not mind the order that AS actions run, can reduce/prevent deadlocks by removing the `ORDER BY` using this filter.

This is a less invasive and backwards compatible approach compared to #748 

### Changelog

> Dev - Adds new filter `action_scheduler_claim_actions_order_by` to allow tuning of the claim query (props @glagonikas)

PR #1106, opened 2024-09-28
Added $claim_id and $hooks as params to action_scheduler_claim_actions_order_by filter

gosixl:
Allows toggling order by statement on the basis of the specific action. Allows skipping where not needed and benefiting from performance

See #773 from @glagonikas & @barryhughes

### DIFF
--- a/classes/data-stores/ActionScheduler_DBStore.php
+++ b/classes/data-stores/ActionScheduler_DBStore.php
@@ -932,7 +932,7 @@ AND `group_id` = %d
 		 *
 		 * @param string $order_by_sql
 		 */
-		$order    = apply_filters( 'action_scheduler_claim_actions_order_by', 'ORDER BY priority ASC, attempts ASC, scheduled_date_gmt ASC, action_id ASC' );
+		$order    = apply_filters( 'action_scheduler_claim_actions_order_by', 'ORDER BY priority ASC, attempts ASC, scheduled_date_gmt ASC, action_id ASC', $claim_id, $hooks );
 		$params[] = $limit;
 
 		$sql           = $wpdb->prepare( "{$update} {$where} {$order} LIMIT %d", $params ); // phpcs:ignore WordPress.DB.PreparedSQL.InterpolatedNotPrepared, WordPress.DB.PreparedSQLPlaceholders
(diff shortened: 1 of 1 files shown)


classes/data-stores/ActionScheduler_DBStore.php
```diff
@@ -932,7 +932,7 @@ protected function claim_actions( $claim_id, $limit, \DateTime $before_date = nu
 		 *
 		 * @param string $order_by_sql
 		 */
-		$order    = apply_filters( 'action_scheduler_claim_actions_order_by', 'ORDER BY priority ASC, attempts ASC, scheduled_date_gmt ASC, action_id ASC' );
+		$order    = apply_filters( 'action_scheduler_claim_actions_order_by', 'ORDER BY priority ASC, attempts ASC, scheduled_date_gmt ASC, action_id ASC', $claim_id, $hooks );
```
barryhughes, line 938: I wonder if it would be cleaner (in terms of further possible future modifications) to pass additional information via an array? For instance:

```php
$properties = array(
	'claim_id' => $claim_id,
	'hooks'    => $hooks,
);

$order = apply_filters( 
	'action_scheduler_claim_actions_order_by', 
	'ORDER BY priority ASC, attempts ASC, scheduled_date_gmt ASC, action_id ASC', 
	$properties
);
```

We should also make sure the docblock reflects the available parameters, whether or not we adopt that suggestion.

gosixl:
Thanks @barryhughes just updated the docblocks for my first edit - hopefully correctly

classes/data-stores/ActionScheduler_DBStore.php
gosixl: I'm easy - would you like me to add that?

barryhughes: On reflection, let's stick with your current approach. It's probably worth adding an *additional* [`@since`](https://docs.phpdoc.org/guide/references/phpdoc/tags/since.html) tag, though. Something like:

```php
 * @since 3.8.3 Made $claim_id and $hooks available.
```

barryhughes (approved):
Thank you for submitting this!

barryhughes:
...Once CI passes, we can go ahead and merge (or else we can tweak if they indicate some problem) 👍🏼

gosixl:
No prob - thank you!

PR #1109, opened 2024-10-03
PHPCS: `deprecated/ActionScheduler_AdminView_Deprecated.php`

crstauf:


crstauf:
@barryhughes Can you please re-trigger the checks?

barryhughes:
Failing again: we (probably) need to revise the way Composer dependencies are cached. They seem pretty clearly unrelated to the change you made, though, so I think we can ignore/fall back on running tests locally.

barryhughes (approved):

crstauf:
@barryhughes FYI I'm not running local tests, as Local is a real pain with unit testing. I'm depending on GitHub checks to inform on things breaking, and addressing that way. Right now, I'm just getting the PHPCS fixes implemented, then will swing around behind and check for failures.

barryhughes:
> I'm depending on GitHub checks to inform on things breaking, and addressing that way.

Understood. The suite is not in an ideal state at present.

PR #1115, opened 2024-10-03
PHPCS: `classes/ActionScheduler_AdminView.php`

crstauf:


classes/ActionScheduler_AdminView.php
```diff
@@ -35,7 +50,7 @@ public static function instance() {
 	 * @codeCoverageIgnore
 	 */
 	public function init() {
-		if ( is_admin() && ( ! defined( 'DOING_AJAX' ) || false == DOING_AJAX ) ) {
+		if ( is_admin() && ( ! defined( 'DOING_AJAX' ) || false === DOING_AJAX ) ) {
```
barryhughes:
```suggestion
		if ( is_admin() && ( ! defined( 'DOING_AJAX' ) || ! DOING_AJAX ) ) {
```

I think this might be the safer way to solve for `WordPress.PHP.StrictComparisons.LooseComparison` in this particular scenario. For instance, it doesn't seem impossible that someone somewhere will have code like this:

```php
define( 'DOING_AJAX', 0 );
```

classes/ActionScheduler_AdminView.php
```diff
@@ -190,23 +203,35 @@ protected function check_pastdue_actions() {
...
-			// translators: 1) is the number of affected actions, 2) is a link to an admin screen.
-			_n(
-				'<strong>Action Scheduler:</strong> %1$d <a href="%2$s">past-due action</a> found; something may be wrong. <a href="https://actionscheduler.org/faq/#my-site-has-past-due-actions-what-can-i-do" target="_blank">Read documentation &raquo;</a>',
-				'<strong>Action Scheduler:</strong> %1$d <a href="%2$s">past-due actions</a> found; something may be wrong. <a href="https://actionscheduler.org/faq/#my-site-has-past-due-actions-what-can-i-do" target="_blank">Read documentation &raquo;</a>',
-				$num_pastdue_actions,
-				'action-scheduler'
+			wp_kses(
+				// translators: 1) is the number of affected actions, 2) is a link to an admin screen.
+				_n(
+					'<strong>Action Scheduler:</strong> %1$d <a href="%2$s">past-due action</a> found; something may be wrong. <a href="https://actionscheduler.org/faq/#my-site-has-past-due-actions-what-can-i-do" target="_blank">Read documentation &raquo;</a>',
```
barryhughes, line 221: (For my own future ref) recognizing that ideally we wouldn't have HTML in translatable strings, but it seems fine to leave it in place here (to minimize unnecessary disruption).

barryhughes (approved):

PR #1119, opened 2024-10-03
PHPCS: `classes/ActionScheduler_DateTime.php`

crstauf:


classes/ActionScheduler_DateTime.php
```diff
@@ -15,7 +15,7 @@ class ActionScheduler_DateTime extends DateTime {
 	 *
 	 * @var int
 	 */
-	protected $utcOffset = 0;
+	protected $utc_offset = 0;
```
barryhughes: I don't expect this will be a problem in practice, but changing the names of non-private properties or methods *could* be a problem for third party extensions.

crstauf: @barryhughes I can create a PR to exclude that sniff, or ignore existing variables. Please advise.

barryhughes: I think disabling the sniff (for this class only) makes sense. Let's not exclude for the entire codebase.

barryhughes (approved):

PR #1132, opened 2024-10-03
PHPCS: `classes/WP_CLI/ActionScheduler_WPCLI_Scheduler_command.php`

crstauf:


classes/WP_CLI/ActionScheduler_WPCLI_Scheduler_command.php
```diff
@@ -175,11 +171,7 @@ protected function print_total_batches( $batches_completed ) {
 	/**
 	 * Convert an exception into a WP CLI error.
 	 *
-	 * @author Jeremy Pry
-	 *
 	 * @param Exception $e The error object.
-	 *
-	 * @throws \WP_CLI\ExitException
```
barryhughes, line 182: Why remove?

crstauf: PHPCS requires a comment with the statement, and as I did not see an exception being thrown, I could not determine the comment to write, so removed.

barryhughes: For completeness, I think it's better to keep it, as the exception being raised is still a possible outcome of calling this method. Maybe:

```php
	 * @throws \WP_CLI\ExitException Under some conditions WP CLI may throw an exception.
```

crstauf: Where's the documentation for that? I looked for it first time around, and today, and never found any documentation to suggest it may throw an exception.

barryhughes: I'm not sure if it is documented, but it's in the source for [WP_CLI::error()](https://github.com/wp-cli/wp-cli/blob/v2.11.0/php/class-wp-cli.php#L905-L923), and so static analysis tools also see and warn of it (unless a `@throws` tag is present).

classes/WP_CLI/ActionScheduler_WPCLI_Scheduler_command.php
```diff
@@ -175,11 +171,9 @@ protected function print_total_batches( $batches_completed ) {
 	/**
 	 * Convert an exception into a WP CLI error.
 	 *
-	 * @author Jeremy Pry
-	 *
 	 * @param Exception $e The error object.
 	 *
-	 * @throws \WP_CLI\ExitException
+	 * @throws \WP_ClI\ExitException Under some conditions WP CLI may throw an exception.
```
barryhughes: Applying a further adjustment (minor).

```suggestion
	 * @throws \WP_CLI\ExitException Under some conditions WP CLI may throw an exception.
```

barryhughes (approved):

PR #1134, opened 2024-10-04
PHPCS: `classes/WP_CLI/ProgressBar.php`

crstauf:


classes/WP_CLI/ProgressBar.php
```diff
@@ -64,7 +84,7 @@ public function tick() {
 		$this->progress_bar->tick();
 		$this->total_ticks++;
 
-		do_action( 'action_scheduler/progress_tick', $this->total_ticks );
+		do_action_deprecated( 'action_scheduler/progress_tick', $this->total_ticks ); // phpcs:ignore WordPress.NamingConvetions.ValidHookName.UseUnderscores
```
barryhughes: Is the switch to `do_action_deprecated` intentional and related to the larger [WP CLI-related PR](https://github.com/woocommerce/action-scheduler/pull/1071), can we/should we offer an explanation/supply a `$message` if so?

crstauf: I think I must've meant to add a hook that adheres to the PHPCS standards. Will add that now.

classes/WP_CLI/ProgressBar.php
```diff
@@ -64,7 +84,9 @@ public function tick() {
 		$this->progress_bar->tick();
 		$this->total_ticks++;
 
-		do_action( 'action_scheduler/progress_tick', $this->total_ticks );
+		// TODO: add version.
+		do_action_deprecated( 'action_scheduler/progress_tick', $this->total_ticks, '', 'action_scheduler_progress_tick' ); // phpcs:ignore WordPress.NamingConvetions.ValidHookName.UseUnderscores
+		do_action( 'action_scheduler_progress_tick', $this->total_ticks );
```
barryhughes: Suggestion:

```suggestion
		do_action( 'action_scheduler_progress_tick', $this->total_ticks ); // phpcs:ignore WordPress.NamingConventions.ValidHookName.UseUnderscores
```

Thinking on this/notes:

- Most importantly, we are not giving other `action_scheduler/*` hooks the same treatment and, at this point, I also don't think we should:
    - This naming convention is a bit of an anomaly, but it's not unheard of (some popular plugins out there do something similar) and, if developers out there are using it, I don't think there is any real benefit to causing disruption/noise for them or their users, just to satisfy a suppressable linter rule.
    - If we do decide to address this, we should do it consistently across the codebase.
- Also, in the event we do keep this approach, we would no longer need to `phpcs:ignore` rule (which has a typo btw, but in any case it doesn't apply to deprecated hooks).

crstauf: @barryhughes I'm not sure what I am to do here. Please clarify.

barryhughes: Sorry. I'm recommending the approach in the suggested change snippet: keep the `phpcs:ignore` (but fixing the rule name), and removing the `do_action_deprecated()` call.

```suggestion
		do_action( 'action_scheduler_progress_tick', $this->total_ticks ); // phpcs:ignore WordPress.NamingConventions.ValidHookName.UseUnderscores
```

classes/WP_CLI/ProgressBar.php
```diff
@@ -64,7 +84,7 @@ public function tick() {
 		$this->progress_bar->tick();
 		$this->total_ticks++;
 
-		do_action( 'action_scheduler/progress_tick', $this->total_ticks );
+		do_action( 'action_scheduler_progress_tick', $this->total_ticks );
```
barryhughes: I apologize, there was a mistake in my earlier recommendation (which is probably why it was so confusing).

```suggestion
		do_action( 'action_scheduler/progress_tick', $this->total_ticks ); // phpcs:ignore WordPress.NamingConventions.ValidHookName.UseUnderscores
```

I'll apply the above: we cannot (or, should not) change the hook name, so simply ignoring the rule here makes the most sense.

barryhughes (approved):

classes/WP_CLI/ProgressBar.php
crstauf: I implemented your suggestion in [the previous comment](https://github.com/woocommerce/action-scheduler/pull/1134#discussion_r1815791089)! 😆

PR #1163, opened 2024-10-04
PHPCS: `classes/schedules/ActionScheduler_CanceledSchedule.php`

crstauf:


classes/schedules/ActionScheduler_CanceledSchedule.php
```diff
@@ -33,6 +35,8 @@ public function get_next( DateTime $after ) {
 	}
 
 	/**
+	 * Action is not recurreing.
```
barryhughes:
```suggestion
	 * Action is not recurring.
```

barryhughes (approved):

PR #1189, opened 2024-10-04
PHPCS: `tests/phpunit/logging/ActionScheduler_wpCommentLogger_Test.php`

crstauf:


crstauf:
@barryhughes Check failure is because of Composer: please re-run.

tests/phpunit/logging/ActionScheduler_wpCommentLogger_Test.php
```diff
@@ -1,10 +1,13 @@
 <?php
+// phpcs:disable WordPress.PHP.StrictInArray.MissingTrueStrict
```
barryhughes: On first glance, the `in_array` calls look like they could be made strict, which I think would be preferred over disabling this check for the entire class (or did you find any problems with that)?

crstauf: Making them strict caused the tests to fail. The tests seems to be checking that numeric strings work, so need to ignore that sniff in this case.

You can see that by reviewing the checks on commit f97745b.

crstauf:
@barryhughes PHPUnit setup failure...?

tests/phpunit/logging/ActionScheduler_wpCommentLogger_Test.php
barryhughes: Thanks, then I think it still makes more sense to apply the exceptions line-by-line (or, enabling and disabling for *blocks* of lines like [here](https://github.com/woocommerce/action-scheduler/blob/3.8.2/tests/phpunit/logging/ActionScheduler_wpCommentLogger_Test.php#L109-L111), where we have three `in_array()` calls in a row), rather than excluding the entire file.

This is principally because if at some future point we add additional tests or assertions in this file, we will want to flag any non-strict `in_array()` checks.

barryhughes (approved):

PR #1193, opened 2024-10-04
PHPCS: `tests/phpunit/migration/LogMigrator_Test.php`

crstauf:


tests/phpunit/migration/LogMigrator_Test.php
```diff
@@ -7,27 +7,27 @@
...
-		$logs = [];
-		for ( $i = 0 ; $i < 3 ; $i++ ) {
-			for ( $j = 0 ; $j < 5 ; $j++ ) {
-				$logs[ $i ][ $j ] = md5(rand());
-				if ( $i == 1 ) {
+		$logs = array();
+		for ( $i = 0; $i < 3; $i++ ) {
+			for ( $j = 0; $j < 5; $j++ ) {
+				$logs[ $i ][ $j ] = md5( wp_rand() );
+				if ( 1 === absint( $i ) ) {
```
barryhughes: `absint()` seems unnecessary, we could just have a strict comparison (we can see from line 27 it will be an integer belonging to a pre-defined range).

```suggestion
				if ( 1 === $i ) {
```

barryhughes (approved):

PR #1205, opened 2024-10-17
Make nullable parameters explicitly nullable

barryhughes:
For improved compatibility with PHP 8.4, and to reduce the amount of error noise we see when running under that version, this change fixes a number of method signatures where parameters are implicitly nullable. For example:

```php
public function __construct( ActionScheduler_Store $store = null, $batch_size = 20 )
```

`$store` must be of type `ActionScheduler_Store`, but if not specified by the caller it is set to `null`. This is fine, but as of PHP 8.4 we are encouraged to be explicit about it, so we need:

```php
public function __construct( ?ActionScheduler_Store $store = null, $batch_size = 20 ) {
```

The problem with the above is that `?Type` is only valid [from PHP 7.1 onwards](https://www.php.net/manual/en/migration71.new-features.php), whereas we still support PHP 7.0. We also cannot yet use type unions. So, we have a decision to make:

- Bump the minimum supported version of PHP (feels reasonable if it's a small adjustment, from 7.0 to 7.1).
- Dispense with this change and put up with the warning/error noise.

### DIFF
--- a/classes/abstracts/ActionScheduler_Store.php
+++ b/classes/abstracts/ActionScheduler_Store.php
@@ -37,7 +37,7 @@ abstract class ActionScheduler_Store extends ActionScheduler_Store_Deprecated {
 	 *
 	 * @return int The action ID
 	 */
-	abstract public function save_action( ActionScheduler_Action $action, DateTime $scheduled_date = null );
+	abstract public function save_action( ActionScheduler_Action $action, ?DateTime $scheduled_date = null );
 
 	/**
 	 * Get action.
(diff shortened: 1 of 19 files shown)


barryhughes:
💡 Please note that (assuming we are ultimately happy with this change) we should not merge ahead of https://github.com/woocommerce/action-scheduler/pull/1212, which increases the minimum required PHP version to 7.1. Even so, it felt like there was no particular reason to delay review and testing.

(The test fails you will see all relate to PHP 7.0 ... that version will be removed from the test matrix via the other PR I mentioned.)

classes/abstracts/ActionScheduler_Store.php
```diff
@@ -313,11 +313,11 @@ protected function get_scheduled_date_string( ActionScheduler_Action $action, Da
 	/**
 	 * Get the time MySQL formatted date/time string for an action's (next) scheduled date.
 	 *
-	 * @param ActionScheduler_Action $action Action.
-	 * @param null|DateTime          $scheduled_date Action's scheduled date (optional).
+	 * @param ActionScheduler_Action|null $action Action.
```
coreymckrill, line 316: It looks like this param isn't actually nullable?

barryhughes:
```suggestion
	 * @param ActionScheduler_Action      $action Action.
```

barryhughes: Thanks for spotting that ... corrected via the above suggestion: https://github.com/woocommerce/action-scheduler/pull/1205/commits/9c65d1cd4220ecbe2d7ae84d6679f2b6cf050098

coreymckrill (approved):

barryhughes:
Thanks for the review. Rebased following the merge of https://github.com/woocommerce/action-scheduler/pull/1212, and will merge this if/once all our checks pass.

PR #1212, opened 2024-10-30
Increase minimum required version of PHP to 7.1 (up from 7.0).

### LINKED ISSUE #1211: Bump PHP min version to 7.1
We plan to bump the minimum required PHP version to 7.1 (previously was 7.0). This will make it easier to solve a class of problems that occur when running PHP 8.4 (expected to be released sometime in November 2024).

- The upcoming version number will then become 3.9.0 (instead of 3.8.3, as planned before now).
- Projects that support earlier versions of PHP (and that cannot make a matching adjustment) should continue to use the 3.8.2 release.

barryhughes:
We are increasing the minimum required version of PHP. Functionally, nothing has changed (but we will take advantage of this in some upcoming PRs, to address problems that occur when using PHP 8.4).

Closes #1211.

### 🥼 Testing

1. As noted above, functionally, nothing has changed. As a minimum, we still expect the test suite to pass without errors.
2. A high level review or 'smoketest' of the plugin is probably useful:
    1. Activate the plugin (as a top-level, standalone plugin).
    2. Ensure the **Tools ‣ Scheduled Actions** admin screen is still present and functions as normal.
3. Feel free to do any other exploratory testing you feel may be useful.

### :memo: Changelog

> Bump minimum PHP version to 7.1.

### DIFF
--- a/.github/workflows/pr-unit-tests.yml
+++ b/.github/workflows/pr-unit-tests.yml
@@ -10,16 +10,16 @@ jobs:
       fail-fast: false
       matrix:
         # We test against the earliest and latest PHP versions for each major supported version.
-        php: [ '7.0', '7.4', '8.0', '8.3' ]
+        php: [ '7.1', '7.4', '8.0', '8.3' ]
         wp: [ '6.4', '6.5', '6.6', 'latest', 'nightly' ]
         multisite: [ '0', '1' ]
         exclude:
           # WordPress 6.6+ requires PHP 7.2+
-          - php: 7.0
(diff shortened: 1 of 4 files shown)


barryhughes:
Once/if we are happy with this change, we can go ahead and update the base branch for https://github.com/woocommerce/action-scheduler/pull/1205, then hopefully get that worked reviewed and merged. This will also be useful in getting projects like WooCommerce ready for PHP 8.4.

albarin (approved):

PR #1218, opened 2024-12-02
Add `source` WP CLI command

crstauf:
WP CLI command that lists all the sources and versions of Action Scheduler available on the WP site.

Supersedes and closes crstauf/action-scheduler#3.

```shell
$ wp action-scheduler source       
wp-content/plugins/action-scheduler% 

$ wp action-scheduler source --fullpath
/Volumes/crstauf/actionscheduler/app/public/wp-content/plugins/action-scheduler%  

$ wp action-scheduler source --all
+-------------------------------------------------------------------------------+---------+--------+
| source                                                                        | version | active |
+-------------------------------------------------------------------------------+---------+--------+
| wp-content/plugins/action-scheduler/action-scheduler.php                      | 3.9.0   | yes    |
| wp-content/plugins/woocommerce/packages/action-scheduler/action-scheduler.php | 3.8.2   | no     |
+-------------------------------------------------------------------------------+---------+--------+

$ wp action-scheduler source --all --fullpath
+---------------------------------------------------------------------------------------------------------------------------+---------+--------+
| source                                                                                                                    | version | active |
+---------------------------------------------------------------------------------------------------------------------------+---------+--------+
| /Volumes/crstauf/actionscheduler/app/public/wp-content/plugins/action-scheduler/action-scheduler.php                      | 3.9.0   | yes    |
| /Volumes/crstauf/actionscheduler/app/public/wp-content/plugins/woocommerce/packages/action-scheduler/action-scheduler.php | 3.8.2   | no     |
+---------------------------------------------------------------------------------------------------------------------------+---------+--------+
```

### Changelog

> Add - New `wp action-scheduler source` command to help determine how Action Scheduler is being loaded.


barryhughes:
Thanks! I think the only concern I have here is about the behavior of `source --all`:

```
% wp action-scheduler source --all

+----------------------------------------------------------+---------+--------+
| source                                                   | version | active |
+----------------------------------------------------------+---------+--------+
| wp-content/plugins/action-scheduler/action-scheduler.php | 3.9.0   | yes    |
| wp-content/plugins/foo/.../action-scheduler.php          | 3.8.2   | no     |
| wp-content/plugins/bar/.../action-scheduler.php          | 3.7.3   | no     |
+----------------------------------------------------------+---------+--------+
```

It's behaving as expected, in the sense it is listing all *registered* copies of Action Scheduler (as described if you run `wp help action-scheduler source`), but I wonder if users will grasp the nuances of this, particularly in the sense that given two or more copies with the same version number, only the first will be registered.

In my hypothetical example above, I could for instance have 10 other plugins, each containing Action Scheduler 3.9.0, but we wouldn't know it from this report.

Not a major blocker, because I think the new `source` sub-command brings a lot of value outside of this scenario. Still, I think we may want to make this clearer *or* perhaps remove support for `--all`.

What do you think?

crstauf:
Yeah, I've been mulling over how to solve that for some time now. I think `source --all` can be useful as-is though, as an easy way to see which plugins are behind on the versions. For instance, if WooCommerce has an older version than another plugin, that may prove useful in debugging an issue. Maybe moving it to the `version --all` command would be better?

My only solution to this is to add something like below to `/action-scheduler.php` to track all sources:

```php
add_action( 'plugins_loaded', static function () {
    ActionScheduler::register_source( __FILE__, '3.9.0' );
} );
```

I think that would work, but I'm not sold on it: this is really only a debugging feature, so doing this in production environments seems weird. A dev could add debugging code themselves to figure it out anyway.

barryhughes:
For sure, it's a bit of an oddity. I agree it's still useful, I just want to make sure it doesn't inadvertently cause confusion for our users, for other developers, or for our future selves. What do you think of a change in which we add some text immediately above the table? For instance:

```
% wp action-scheduler source --all

Please note there can only be one unique registered instance of Action Scheduler per
version number, so this list may not include all the currently present copies of 
Action Scheduler.

+----------------------------------------------------------+---------+--------+
| source                                                   | version | active |
+----------------------------------------------------------+---------+--------+
| wp-content/plugins/action-scheduler/action-scheduler.php | 3.9.0   | yes    |
| wp-content/plugins/foo/.../action-scheduler.php          | 3.8.2   | no     |
| wp-content/plugins/bar/.../action-scheduler.php          | 3.7.3   | no     |
+----------------------------------------------------------+---------+--------+
```

crstauf:
Sure. I'll get that implemented.

crstauf:
@barryhughes Added notice.

classes/WP_CLI/System_Command.php
```diff
@@ -180,6 +184,10 @@ public function source( array $args, array $assoc_args ) {
 
 		ksort( $rows );
 
+		if ( ! $skipnote ) {
+			\WP_CLI::line( PHP_EOL . 'Please note there can only be one unique registered instance of Action Scheduler per ' . PHP_EOL . 'version number, so this list may not include all the currently present copies of ' . PHP_EOL . 'Action Scheduler.' . PHP_EOL );
```
barryhughes: Wondering now if it is or is not a good practice to manually "word-wrap" like this 🤔

barryhughes: Following this train-of-thought (I was curious if WP CLI might either have some notes on best practices, or else have a helper that manages this for us) I noticed that perhaps we should [now prefer `::log()` instead of `::line()`](https://make.wordpress.org/cli/handbook/references/internal-api/wp-cli-line/#notes):

> Message is written to STDOUT. WP_CLI::log() is typically recommended;
WP_CLI::line() is included for historical compat.

Thinking about it, using `::log()` may be more elegant because it pretty much [removes the need for us to implement `--skip-notice`](https://make.wordpress.org/cli/handbook/references/internal-api/wp-cli-log/#notes) (which otherwise was a nice touch)

> Message is written to STDOUT, or discarded when --quiet flag is supplied.

barryhughes (approved):
Left some notes regarding use of `WP_CLI::line()`, but I'm not sure it is essential we address them 'today'. Leaving this open so you can review, then please either update or else leave a reply if you prefer us to move ahead without. Thanks!

barryhughes (approved):
Great, thank you!

PR #1220, opened 2024-12-06
Make some other nullable parameters explicitly nullable

jorgeatorres:
Here we continue the work started in #1205, where most function parameters that are implicitly nullable were explicitly marked as nullable. That is, function definitions such as 

```
public function __construct( ActionScheduler_Store $store = null, $batch_size = 20 )
```

were turned into

```
public function __construct( ?ActionScheduler_Store $store = null, $batch_size = 20 ) {
```

(Notice the type change in the `$store` argument).

We missed a couple instances of these nullable args in the prior PR and so that's what this one adds. The aim is to reduce noise when running A-S (or a plugin that bundles A-S) under PHP 8.4.

### DIFF
--- a/lib/cron-expression/CronExpression.php
+++ b/lib/cron-expression/CronExpression.php
@@ -54,7 +54,7 @@ class CronExpression
      *
      * @return CronExpression
      */
-    public static function factory($expression, CronExpression_FieldFactory $fieldFactory = null)
+    public static function factory($expression, ?CronExpression_FieldFactory $fieldFactory = null)
     {
         $mappings = array(
             '@yearly' => '0 0 1 1 *',
--- a/classes/ActionScheduler_ActionFactory.php
+++ b/classes/ActionScheduler_ActionFactory.php
@@ -8,17 +8,17 @@ class ActionScheduler_ActionFactory {
 	/**
 	 * Return stored actions for given params.
 	 *
-	 * @param string                   $status The action's status in the data store.
-	 * @param string                   $hook The hook to trigger when this action runs.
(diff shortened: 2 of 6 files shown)


lib/cron-expression/CronExpression.php
coreymckrill, line 1: Given that this is a 3rd party library, I wondered if we should be directly editing the files in it. It looks like [it is abandoned](https://packagist.org/packages/mtdowling/cron-expression), which might be why we've got it under version control rather than as a composer dependency. At some point, the package [was forked](https://github.com/dragonmantank/cron-expression), and the new maintainer is still active. The latest version does add the [necessary nullable type](https://github.com/dragonmantank/cron-expression/blob/f37e405332cd1cca9d287f5044ae2a72c2f53239/src/Cron/CronExpression.php#L150), but the readme also says there are breaking changes since the version that we're using.

So in short, probably fine to edit this file, but perhaps we should look into updating this to a composer dependency using the latest version, both for PHP compatibility reasons, and also potentially security reasons.

coreymckrill (approved):
👍 Did a quick smoke test and everything looks fine.

lib/cron-expression/CronExpression.php
coreymckrill: Opened an issue for this: https://github.com/woocommerce/action-scheduler/issues/1221

jorgeatorres: Thank you @coreymckrill! 💯 Same thought crossed my mind when I was modifying that file but forgot to mention it in the description.

Issue #772, opened 2021-10-07
Add parent plugin details to `Help` pulldown


barryhughes:
Following some (mostly internal) discussion around https://github.com/woocommerce/action-scheduler/issues/730 it was surfaced that it could be useful to add information about the plugin (or theme) responsible for loading Action Scheduler to the **Tools → Scheduled Actions** screen's **Help** tab.

Mock:

<img width="957" alt="modified-help-tab" src="https://user-images.githubusercontent.com/3594411/136425001-b5e80f8d-1fb9-4de4-a8ce-5d42c530b6f2.png">

This would be useful to assist with troubleshooting when:

* Multiple plugins/theme all include Action Scheduler.
* When an end-user does not know which of their components is including Action Scheduler, but needs assistance with it.

crstauf:
@barryhughes see #271.

barryhughes:
Recently submitted https://github.com/woocommerce/action-scheduler/pull/1218 reminded me of this issue. If it can successfully determine the active version (and, based on the description, it can) we could also use that as a foundation for *this* issue.

crstauf:
Simple enough: all I did was use `dirname( __DIR__, $level )` to return the path to the root of Action Scheduler library.

crstauf:
@barryhughes Are you interested in implementing the name of the theme or plugin from which Action Scheduler is being loaded (as in OP screenshot), or is the source path sufficient (what was merged)?

barryhughes:
I think this is *sufficient,* but I also think if there is a reliable way to supplement that with the actual plugin name and version (assuming it is coming from a plugin), that would be a really nice addition.

crstauf:
Understood. I had started working on that, so will open that up in a different branch and PR.

barryhughes:
Excellent, thank you.

PR #1222, opened 2024-12-09
Display name of source of Action Scheduler library

crstauf:
See #772.

---

as-help-pulldown

Information about the active version of Action Scheduler can now be found under the **Source** heading, via the **Help** pulldown found in the **Tools ‣ Scheduled Actions** screen.

### Changelog

> Add - Added additional information about the active copy of Action Scheduler to the Help pulldown.

### DIFF
--- a/classes/ActionScheduler_Versions.php
+++ b/classes/ActionScheduler_Versions.php
@@ -110,11 +110,21 @@ class ActionScheduler_Versions {
 	}
 
 	/**
-	 * Get directory of active source.
+	 * Returns information about the plugin or theme which contains the current active version
+	 * of Action Scheduler.
 	 *
-	 * @return string
+	 * If this cannot be determined, or if Action Scheduler is being loaded via some other
+	 * method, then it will return an empty array. Otherwise, if populated, the array will
+	 * look like the following:
+	 *
+	 *     [
+	 *         'type' => 'plugin', # or 'theme'
+	 *         'name' => 'Name',
+	 *     ]
+	 *
+	 * @return array
 	 */
-	public function active_source() {
+	public function active_source(): array {
 		$file         = __FILE__;
 		$dir          = __DIR__;
 		$plugins      = get_plugins();
@@ -157,6 +167,7 @@ class ActionScheduler_Versions {
 
 			return array(
 				'type' => 'theme',
+				// phpcs:ignore WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase
 				'name' => $theme->Name,
 			);
 		}
@@ -164,7 +175,12 @@ class ActionScheduler_Versions {
 		return array();
 	}
 
-	public function active_source_path() {
+	/**
+	 * Returns the directory path for the currently active installation of Action Scheduler.
+	 *
+	 * @return string
+	 */
+	public function active_source_path(): string {
 		return trailingslashit( dirname( __DIR__ ) );
 	}
 }
--- a/classes/ActionScheduler_Versions.php
+++ b/classes/ActionScheduler_Versions.php
@@ -115,6 +115,58 @@ class ActionScheduler_Versions {
 	 * @return string
 	 */
 	public function active_source() {
+		$file   = __FILE__;
+		$dir    = __DIR__;
+		$path   = $this->active_source_path();
+		$themes = (array) search_theme_directories();
+
+		foreach ( $themes as $slug => $data ) {
+			$needle = trailingslashit( $data['theme_root'] ) . $slug . '/';
+
+			if ( 0 !== strpos( $file, $needle ) ) {
+				continue;
+			}
+
+			$theme = wp_get_theme( $slug );
+
+			if ( ! is_object( $theme ) || ! is_a( $theme, \WP_Theme::class ) ) {
+				continue;
+			}
+
+			return array(
+				'type' => 'theme',
+				'name' => $theme->Name,
+			);
+		}
+
+		$plugins      = get_plugins();
+		$plugin_files = array_keys( $plugins );
+		$plugin_paths = array();
+
+		foreach ( $plugin_files as $plugin_file ) {
+			$plugin_path = trailingslashit( WP_PLUGIN_DIR ) . dirname( $plugin_file );
+			$plugin_file = trailingslashit( WP_PLUGIN_DIR ) . $plugin_file;
+
+			if ( 0 !== strpos( dirname( $dir ), $plugin_path ) ) {
+				continue;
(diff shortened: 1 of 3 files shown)


crstauf:
@barryhughes Okay, I think this is ready.

classes/ActionScheduler_Versions.php
```diff
@@ -115,6 +115,58 @@ public static function initialize_latest_version() {
 	 * @return string
 	 */
 	public function active_source() {
+		$file   = __FILE__;
+		$dir    = __DIR__;
```
barryhughes: Instead of assigning the value of these constants to variables (especially since we don't subsequently mutate those variables), we could just use the constants directly later on.

classes/ActionScheduler_Versions.php
```diff
@@ -115,6 +115,58 @@ public static function initialize_latest_version() {
...
+		$path   = $this->active_source_path();
```
barryhughes: I may be mistaken, but I don't believe `$path` is used, so it can probably be removed.

classes/ActionScheduler_Versions.php
```diff
@@ -115,6 +115,58 @@ public static function initialize_latest_version() {
...
+		$themes = (array) search_theme_directories();
+
+		foreach ( $themes as $slug => $data ) {
```
barryhughes: I strongly suspect themes are generally the least common consumers of Action Scheduler. Could it make sense to evaluate this possibility *after* we've looped through the plugins?

classes/ActionScheduler_Versions.php
```diff
@@ -115,6 +115,58 @@ public static function initialize_latest_version() {
...
+
+			$theme = wp_get_theme( $slug );
+
+			if ( ! is_object( $theme ) || ! is_a( $theme, \WP_Theme::class ) ) {
+				continue;
+			}
+
+			return array(
+				'type' => 'theme',
+				'name' => $theme->Name,
```
barryhughes:
```suggestion
				'name' => $theme->name,
```

Seems like this should be lowercase.

classes/ActionScheduler_Versions.php
```diff
@@ -115,6 +115,58 @@ public static function initialize_latest_version() {
...
+			);
+		}
+
+		$plugins      = get_plugins();
+		$plugin_files = array_keys( $plugins );
+		$plugin_paths = array();
```
barryhughes: I think we can remove this one? It doesn't seem to be used.

```suggestion
```

classes/ActionScheduler_Versions.php
```diff
@@ -115,6 +115,58 @@ public static function initialize_latest_version() {
...
+			return array(
+				'type' => 'plugin',
+				'name' => $plugin_data['Name'],
+			);
+		}
+
+		return array();
+	}
+
+	public function active_source_path() {
```
barryhughes: I may be misunderstanding, but I think we'll need to use this new method name in the code underpinning the `wp action-scheduler source` command (which calls `active_source()` but expects it to return a string, whereas with this PR it now returns an array).

barryhughes (changes requested):
Thanks! Works pretty nicely, though I left a few notes and suggestions.

classes/ActionScheduler_Versions.php
crstauf: I have these because it makes it easier to test with. But you're right, they can easily be removed.

classes/ActionScheduler_Versions.php
crstauf: I decided to put themes first, as it seemed the least costly to check. Many sites only have a handful of themes, while plugins (especially WooCommerce sites) can have dozens. If you have a sample WooCommerce site you test with, would be interesting to see data indicating the performance of each section.

classes/ActionScheduler_Versions.php
crstauf: You'd think so, but no:
https://developer.wordpress.org/reference/classes/wp_theme/#properties.

classes/ActionScheduler_Versions.php
barryhughes: > As it seemed the least costly to check.

Probably this is true ... but it also seems far less likely to yield a positive result. 

Suppose the cost of looping over all the themes is typically 5 units, and the cost of looping through all the plugins is typically 20 units. If in 99.% of cases Action Scheduler will A) never be found in the theme loop but B) will be found before we reach the end of the plugin loop, allowing us to bail out early, I think it makes sense to prioritize the plugin loop.

barryhughes: My figures and units are obviously both imaginary. I can't recall a bug ticket or support case where we were working with a theme, though, whereas a great many have surfaced in relation to plugins.

classes/ActionScheduler_Versions.php
barryhughes: Both work, in fact, but yep that seems like the correct one to use (initially it wasn't working for me, which is why I flagged it ... must have been a temporary blip).

crstauf:
@barryhughes Should we support AS being loaded in an mu-plugin?

classes/ActionScheduler_Versions.php
barryhughes: https://github.com/woocommerce/action-scheduler/pull/1222/commits/f5c12a75c06138b33ff1b53ccb5de2efa9e15201

barryhughes:
> Should we support AS being loaded in an mu-plugin?

I think that would make this more complete, though my instinct is it's less important. What is the current behavior in this case? An alternative is doing nothing special, so long as we can still surface the source path.

crstauf:
> What is the current behavior in this case?

If active AS isn't found in theme or plugin, it displays the path.

barryhughes:
> If active AS isn't found in theme or plugin, it displays the path.

That feels reasonable to me. That said, I don't think we'd object if you wanted to implement something that goes beyond this, to handle the specific case of mu-plugins (and/or we could also start with what we already have, and follow-up with that at some later point if still desired).

crstauf:
It would seem to me that installation via mu-plugin is even less likely than installation via a theme. For now, let's leave it searching themes and plugins, and if you all notice more support requests where it's installed via mu-plugins, that can be addressed then.

barryhughes:
📝 Test fails can be disregarded (it's a CI issue), so we'll run locally if we can't resolve through some other means.

barryhughes:
Tests are passing locally, though I'm not running the full test matrix.

barryhughes (approved):

PR #1229, opened 2025-01-12
New 2 handly functions

Arifursdev:
This pull request introduces two new functions to the `functions.php` file to enhance the functionality of the action scheduler. These functions allow for running all queued actions or the next queued action for a given hook, arguments, and group.

New functionality for action scheduler:

* Added `as_run_queued_actions` function to run all queued actions for a given hook, arguments, and group. This function initializes the action scheduler, queries the actions, and processes each action while handling exceptions.
* Added `as_run_next_queued_action` function to run the next queued action for a given hook, arguments, and group. Similar to `as_run_queued_actions`, this function initializes the action scheduler, queries the next action, and processes it while handling exceptions.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -16,7 +16,7 @@
  *
  * @return int The action ID. Zero if there was an error scheduling the action.
  */
-function as_enqueue_async_action( $hook, $args = array(), $group = '', $unique = false, $priority = 10 ) {
+function as_enqueue_async_acspetion( $hook, $args = array(), $group = '', $unique = false, $priority = 10 ) {
 	if ( ! ActionScheduler::is_initialized( __FUNCTION__ ) ) {
 		return 0;
 	}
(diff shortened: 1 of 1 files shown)


functions.php
```diff
@@ -493,3 +493,91 @@ function as_get_datetime_object( $date_string = null, $timezone = 'UTC' ) {
...
+        'orderby' => 'date',
+        'order'   => 'ASC',
+        'group'   => $group,
+    );
+
+    if (is_array($args)) {
+        $params['args'] = $args;
+    }
+
+    $actions = ActionScheduler::store()->query_actions($params);
```
barryhughes, line 521: A concern here is that we may pick up claimed actions (once an action is claimed, its status initially remains `pending`) that are on the verge of being processed by a conventional queue runner, leading to them being processed twice.

functions.php
```diff
@@ -493,3 +493,91 @@ function as_get_datetime_object( $date_string = null, $timezone = 'UTC' ) {
...
+        'orderby' => 'date',
+        'order'   => 'ASC',
+        'group'   => $group,
+    );
+
+    if (is_array($args)) {
+        $params['args'] = $args;
+    }
+
+    $action_id = ActionScheduler::store()->query_action($params);
```
barryhughes, line 565: Same concern as before (possible race condition here).

functions.php
```diff
@@ -493,3 +493,91 @@ function as_get_datetime_object( $date_string = null, $timezone = 'UTC' ) {
...
+}
+
+/**
+ * Run the next queued action for a given hook, arguments, and group.
+ *
+ * @param string $hook The hook that the action is associated with.
+ * @param array  $args Optional. An array of arguments to filter the action. Default is an empty array.
+ * @param string $group Optional. The group the action is assigned to. Default is an empty string.
+ */
+function as_run_next_queued_action($hook, $args = array(), $group = '') {
```
barryhughes, line 548: For both functions, what are your thoughts on a return type? That might be a boolean for this function (because it is only interested in running a single action) or possibly an int for the other function (a count of how many actions were processed)?

functions.php
```diff
@@ -493,3 +493,91 @@ function as_get_datetime_object( $date_string = null, $timezone = 'UTC' ) {
...
 }
+
+/**
+ * Run all queued actions for a given hook, arguments, and group.
+ *
+ * @param string $hook The hook that the actions are associated with.
+ * @param array  $args Optional. An array of arguments to filter the actions. Default is an empty array.
+ * @param string $group Optional. The group the actions are assigned to. Default is an empty string.
+ */
+function as_run_queued_actions($hook, $args = array(), $group = '') {
```
barryhughes, line 504: Worth running the new functions through the linter (`composer run phpcs`).

barryhughes (changes requested):
@Arifursdev sorry for the long delay. Can you confirm if you are still interested in driving this forward? If you are, please review the comments I left.

- My biggest concern with this approach is the potential for a race condition leading to the same action being processed twice (because we skip the claim process/there is no awareness of existing claims).
- New tests would be merited (see `tests/phpunit/procedural_api/`).
- We'd also want to update [the API ref at actionscheduler.org](https://actionscheduler.org/api/) (see `docs/api.md`).
- I definitely see some utility in these functions, but it would be great to hear about your specific use case, to help us make the right calls before committing this to the public API surface.

If you are no longer interested, no worries—we'll probably close the PR in due course.

barryhughes:
Closing this branch: if you (or anyone!) wishes to revisit this, please feel free. The best avenue is probably to create a new issue, and we can discuss requirements etc there before moving to implementation.

Again, our apologies that this sat for so long without a response.

PR #1219, opened 2024-12-06
Add active source to `Help` drawer

crstauf:
Closes #772.
See #1218.

### Changelog

> Add - Additional information about the active instance of Action Scheduler is now available in the Help pull-down drawer.

PR #1237, opened 2025-01-23
Make version/source information available via new class

### LINKED ISSUE #1236: Fatal errors possible when 3.9.1 runs alongside 3.9.0 or earlier
Some [recent changes](https://github.com/woocommerce/action-scheduler/compare/3.9.0...3.9.1) touched the [`ActionScheduler_Versions`](https://github.com/woocommerce/action-scheduler/blob/3.9.1/classes/ActionScheduler_Versions.php) class, as part of an effort to make it easier to determine which plugin is loading Action Scheduler (see https://github.com/woocommerce/action-scheduler/pull/1218 and https://github.com/woocommerce/action-scheduler/pull/1219).

The problem is that, even if 3.9.1 is the active version of Action Scheduler, the `ActionScheduler_Versions` class itself may come from an older version and therefore may not contain the new methods introduced in those PRs.

- This problem can be demonstrated if you have a plugin named `aaa` (named as such so that it loads earlier than most other plugins) containing Action Scheduler 3.8.2, and also Action Scheduler 3.9.1.
- Activate both and try visiting the **Tools ‣ Scheduled Actions** screen: the page will be broken. Via your logs, or possibly on screen if `display_errors` is enabled, you should be able to see that a fatal error has taken place *("call to undefined method ActionScheduler_Versions::active_source()")*.
- The same problem also surfaces via a limited number of WP CLI paths, for example: `wp action-scheduler source`.
- *Core* functionality, by which I mean the ability for plugins to add actions and have them processed asynchronously, **is not impacted**.

### Steps to replicate

On a clean WordPress site, add two plugins. One should be Action Scheduler 3.9.1, the other should be a custom plugin named `aaa/aaa.php`:

```php
<?php
/**
 * Plugin name: Test plugin
 */

require __DIR__ . '/vendor/woocommerce/action-scheduler/action-scheduler.php';
```

It should contain a `composer.json` file as follows:

```json
{
    "require": {
        "woocommerce/action-scheduler": "3.8.2"
    }
}
```

Go ahead and run `composer install` for that plugin, then activate both of them. Next, navigate to **Tools ‣ Scheduled Actions** and observe the problem.

barryhughes:
It was though that it could be helpful to add additional information to the *help* pull down (in the **Tools ‣ Scheduled Actions** screen) and via WP CLI about the current version and registered sources of Action Scheduler. This work mostly happened via:

- https://github.com/woocommerce/action-scheduler/pull/1218
- https://github.com/woocommerce/action-scheduler/pull/1219

<div align="center">'Help' pulldown, showing information about the Action Scheduler version</div>

As noted in the [linked issue](https://github.com/woocommerce/action-scheduler/issues/1236), however, this causes a problem because the methods newly added to `ActionScheduler_Versions` will not always be available (if an older version of Action Scheduler is the first to be initialized, then *its* version of that class is the one that is defined). In such a case, we experience fatal errors:

- When the **Tools ‣ Scheduled Actions** screen is visited.
- When a user executes `wp action-scheduler sources`.

This PR attempts to correct the problem by:

- Moving the new methods to a new class. We can be confident this will be available and will belong to the active version of Action Scheduler.
- Deprecates the corresponding methods in `ActionScheduler_Versions` (with one exception, where we add a warning).

However, I also understand it may be preferable to simply revert the changes (...though, the longer 3.9.1 is live, the more problematic it is to strip out public methods).

Closes https://github.com/woocommerce/action-scheduler/issues/1236.

---

### Testing instructions

1. Create a plugin as follows, and run `composer install`. I recommend initially creating it within a folder named `aaa-test`. The main plugin file can simply be named `test.php`.

```php
<?php
/**
 * Plugin name: Test plugin
 */

require __DIR__ . '/vendor/woocommerce/action-scheduler/action-scheduler.php';
```

```json
{
    "require": {
        "woocommerce/action-scheduler": "3.8.2"
    }
}
```

2. Go ahead and activate it, then install and activate Action Scheduler as a standalone plugin. To avoid confusion, deactivate any other plugins besides these two.
3. Visit **Tools ‣ Scheduled Actions**:
    1. Without this branch, it will be broken (fatal error).
    2. With this branch, there should be no problem.
4. Run `wp action-scheduler source --all`:
    1. Without this branch, it will be broken (fatal error).
    2. With this branch, it should inform you that _detailed information about registered sources is not currently available._
5. Rename `aaa-test` to `zzz-test` and reactivate it.
6. Repeat the above tests:
    1. With and without this branch, there should be no fatal errors (that's because the *newer* versions class has been loaded).
    2. There is a slight difference in the case of the WP CLI command. This time you should get a list detailing both plugins:

```
% wp action-scheduler source --all

Please note there can only be one unique registered instance of Action Scheduler per 
version number, so this list may not include all the currently present copies of 
Action Scheduler.

+------------------------------------------------------------------------------------+---------+--------+
| source                                                                             | version | active |
+------------------------------------------------------------------------------------+---------+--------+
| wp-content/plugins/action-scheduler/action-scheduler.php                           | 3.9.0   | no     |
| wp-content/plugins/zztest/vendor/woocommerce/action-scheduler/action-scheduler.php | 3.8.2   | no     |
+------------------------------------------------------------------------------------+---------+--------+
```

### DIFF
--- a/classes/ActionScheduler_SystemInformation.php
+++ b/classes/ActionScheduler_SystemInformation.php
@@ -80,6 +80,10 @@ class ActionScheduler_SystemInformation {
 	/**
 	 * Get registered sources.
 	 *
+	 * It is not always possible to obtain this information. For instance, if earlier versions (<=3.9.0) of
+	 * Action Scheduler register themselves first, then the necessary data about registered sources will
+	 * not be available.
+	 *
 	 * @return array<string, string>
 	 */
 	public static function get_sources() {
--- a/classes/ActionScheduler_SystemInformation.php
+++ b/classes/ActionScheduler_SystemInformation.php
@@ -1,0 +1,89 @@
+<?php
+
+/**
+ * Provides information about active and registered instances of Action Scheduler.
+ */
+class ActionScheduler_SystemInformation {
+	/**
+	 * Returns information about the plugin or theme which contains the current active version
+	 * of Action Scheduler.
+	 *
+	 * If this cannot be determined, or if Action Scheduler is being loaded via some other
+	 * method, then it will return an empty array. Otherwise, if populated, the array will
+	 * look like the following:
+	 *
+	 *     [
+	 *         'type' => 'plugin', # or 'theme'
+	 *         'name' => 'Name',
+	 *     ]
+	 *
+	 * @return array
+	 */
+	public static function active_source(): array {
+		$plugins      = get_plugins();
+		$plugin_files = array_keys( $plugins );
+
+		foreach ( $plugin_files as $plugin_file ) {
+			$plugin_path = trailingslashit( WP_PLUGIN_DIR ) . dirname( $plugin_file );
+			$plugin_file = trailingslashit( WP_PLUGIN_DIR ) . $plugin_file;
+
+			if ( 0 !== strpos( dirname( __DIR__ ), $plugin_path ) ) {
+				continue;
+			}
+
+			$plugin_data = get_plugin_data( $plugin_file );
+
+			if ( ! is_array( $plugin_data ) || empty( $plugin_data['Name'] ) ) {
+				continue;
+			}
+
+			return array(
+				'type' => 'plugin',
+				'name' => $plugin_data['Name'],
+			);
+		}
+
+		$themes = (array) search_theme_directories();
+
+		foreach ( $themes as $slug => $data ) {
+			$needle = trailingslashit( $data['theme_root'] ) . $slug . '/';
+
+			if ( 0 !== strpos( __FILE__, $needle ) ) {
+				continue;
+			}
+
+			$theme = wp_get_theme( $slug );
+
+			if ( ! is_object( $theme ) || ! is_a( $theme, \WP_Theme::class ) ) {
+				continue;
+			}
+
+			return array(
+				'type' => 'theme',
+				// phpcs:ignore WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase
+				'name' => $theme->Name,
+			);
+		}
+
+		return array();
+	}
+
+	/**
+	 * Returns the directory path for the currently active installation of Action Scheduler.
+	 *
+	 * @return string
+	 */
+	public static function active_source_path(): string {
+		return trailingslashit( dirname( __DIR__ ) );
+	}
+
+	/**
+	 * Get registered sources.
+	 *
+	 * @return array<string, string>
+	 */
+	public static function get_sources() {
+		$versions = ActionScheduler_Versions::instance();
+		return method_exists( $versions, 'get_sources' ) ? $versions->get_sources() : array();
+	}
+}
--- a/classes/ActionScheduler_Versions.php
+++ b/classes/ActionScheduler_Versions.php
@@ -133,7 +133,7 @@ class ActionScheduler_Versions {
 	 * @return array
 	 */
 	public function active_source(): array {
-		_deprecated_function( __METHOD__, '3.9.1', 'ActionScheduler_SystemInformation::active_source()' );
+		_deprecated_function( __METHOD__, '3.9.2', 'ActionScheduler_SystemInformation::active_source()' );
 		return ActionScheduler_SystemInformation::active_source();
 	}
 
--- a/classes/ActionScheduler_Versions.php
+++ b/classes/ActionScheduler_Versions.php
@@ -145,7 +145,7 @@ class ActionScheduler_Versions {
 	 * @return string
 	 */
 	public function active_source_path(): string {
-		_deprecated_function( __METHOD__, '3.9.1', 'ActionScheduler_SystemInformation::active_source_path()' );
+		_deprecated_function( __METHOD__, '3.9.2', 'ActionScheduler_SystemInformation::active_source_path()' );
 		return ActionScheduler_SystemInformation::active_source_path();
 	}
 }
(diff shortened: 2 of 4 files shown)


barryhughes:
Sharing this initially as a draft, since we may prefer a different course of action (like reverting the two PRs introducing the relevant changes to the version class).

classes/ActionScheduler_SystemInformation.php
```diff
@@ -0,0 +1,93 @@
+<?php
+
+/**
+ * Provides information about active and registered instances of Action Scheduler.
+ */
+class ActionScheduler_SystemInformation {
```
barryhughes, line 6: The code in this new class is essentially lifted direct from `ActionScheduler_Versions` (though they have become static methods in the process).

barryhughes:
> [!NOTE]
> I've requested a review, but this is still a draft and shouldn't be merged until we know if we want to use this approach (ie, fixing the problematic code) or if instead want to revert the relevant sections (for which we have [these](https://github.com/woocommerce/action-scheduler/pull/1239) [other PRs](https://github.com/woocommerce/action-scheduler/pull/1238)).

classes/ActionScheduler_Versions.php
```diff
@@ -122,65 +128,24 @@ public static function initialize_latest_version() {
...
-
-		return array();
+		_deprecated_function( __METHOD__, '3.9.1', 'ActionScheduler_SystemInformation::active_source()' );
+		return ActionScheduler_SystemInformation::active_source();
 	}
 
 	/**
 	 * Returns the directory path for the currently active installation of Action Scheduler.
 	 *
+	 * @deprecated 3.9.2 Use ActionScheduler_SystemInformation::active_source_path().
```
dwainm, line 143: Is this supposed to be 3.9.**2**?

dwainm (approved):
The test instructions work perfectly.  I think it's a good solution to avoid cross calling completely by using a separate class. I agree with you on the sooner 3.9.2 goes out the better.

classes/ActionScheduler_Versions.php
barryhughes: Yes: but thank you for flagging as, unless I'm having a moment's confusion, some of the other references to 3.9.1 should also be updated.

classes/ActionScheduler_Versions.php
```diff
@@ -122,65 +128,24 @@ public static function initialize_latest_version() {
 	 *         'name' => 'Name',
 	 *     ]
 	 *
+	 * @deprecated 3.9.1 Use ActionScheduler_SystemInformation::active_source().
```
barryhughes:
```suggestion
	 * @deprecated 3.9.2 Use ActionScheduler_SystemInformation::active_source().
```

Building on the discussion in [this thread](https://github.com/woocommerce/action-scheduler/pull/1237/files#r1935667047), this version number should be bumped to 3.9.2.

classes/ActionScheduler_Versions.php
```diff
@@ -122,65 +128,24 @@ public static function initialize_latest_version() {
...
-
-			return array(
-				'type' => 'theme',
-				// phpcs:ignore WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase
-				'name' => $theme->Name,
-			);
-		}
-
-		return array();
+		_deprecated_function( __METHOD__, '3.9.1', 'ActionScheduler_SystemInformation::active_source()' );
```
barryhughes:
```suggestion
		_deprecated_function( __METHOD__, '3.9.2', 'ActionScheduler_SystemInformation::active_source()' );
```

Another one that builds on the discussion in [this thread](https://github.com/woocommerce/action-scheduler/pull/1237/files#r1935667047): this version number should be bumped to 3.9.2.

classes/ActionScheduler_Versions.php
```diff
@@ -122,65 +128,24 @@ public static function initialize_latest_version() {
...
 	/**
 	 * Returns the directory path for the currently active installation of Action Scheduler.
 	 *
+	 * @deprecated 3.9.2 Use ActionScheduler_SystemInformation::active_source_path().
+	 *
 	 * @return string
 	 */
 	public function active_source_path(): string {
-		return trailingslashit( dirname( __DIR__ ) );
+		_deprecated_function( __METHOD__, '3.9.1', 'ActionScheduler_SystemInformation::active_source_path()' );
```
barryhughes:
```suggestion
		_deprecated_function( __METHOD__, '3.9.2', 'ActionScheduler_SystemInformation::active_source_path()' );
```

Again, per [this thread](https://github.com/woocommerce/action-scheduler/pull/1237/files#r1935667047), this version number should be bumped to 3.9.2.

classes/ActionScheduler_Versions.php
barryhughes: @dwainm. Made some updates. Summarizing:

- The problem was introduced in 3.9.1 (current stable release).
- Assuming for the moment this set of changes ship in 3.9.2, then that is the version number to use for the deprecations.
- And in case of any doubt, [this specific version number](https://github.com/woocommerce/action-scheduler/blob/412f067449f98e15048af21d3052205c6fb4a0b4/classes/ActionScheduler_Versions.php#L62) (which is not a deprecation) remains at 3.9.1.

I hope that makes sense, but let me know if this feels incorrect to you.

dwainm: Yes, it makes sense for this to be targeting 3.9.2 as 3.9.1 is out already. We just want to catch any updates early so folks don't even think about 1 and jump straight to two.

barryhughes:
Thank you for the review. Moving this out of draft state (just to make it clear that, from my perspective, it's good-to-go).

PR #1246, opened 2025-03-13
Assume an action is valid until proven otherwise

### LINKED ISSUE #1245: Potential endless loop when processing actions in a claim
In some specific sets of circumstances, the queue runner can attempt to process a group of claimed actions and experience an error early enough in the process that the actions don't get marked as failed, so it continues to try and process them over and over again endlessly.

An example to reproduce the issue:

Add the following code to a file in your test site's mu-plugins directory:

```php
<?php

add_filter( 'action_scheduler_allow_async_request_runner', '__return_false' );
add_action( 'init', fn() => remove_action( ActionScheduler_QueueRunner::WP_CRON_HOOK, [ ActionScheduler_QueueRunner::instance(), 'run' ] ) );
add_action( 'dummy_action', '__return_true' );
add_action( 'action_scheduler_before_execute', 'test_before_execute_callback' );

function test_before_execute_callback() {
	nonexistant_function();
}

function test_generate_dummy_actions() {
	$counter = 0;
	while ( $counter < 10 ) {
		as_schedule_single_action(
			time(),
			'dummy_action',
			array( 1, 2, 3 )
		);
		$counter++;
	}

	printf(
		'Generated %d dummy actions.',
		$counter
	);
}
```

(The first two lines simply disable the async queue runner so we can test using the wp-cli queue runner instead.)

Now generate some fake actions:

```
wp eval "test_generate_dummy_actions();"
```

Then try to process those actions:

```
wp action-scheduler run
```

The call to `nonexistant_function` causes a fatal error, which gets caught by the error handler in `process_action` and converted into an exception. This exits the `try` block before `$valid_action` gets set to `true`, so when it gets sent to the `handle_action_error` method, it looks like an invalid action, so it doesn't try to mark the action as `failed`. Since the action is still `pending`, then, it will just get processed over and over again. (You might have to do a CTRL + C to get out of the loop in the cli.)

coreymckrill:
This way, if an error is thrown before the validation happens, the error handler will still attempt to set the action's status as failed instead of just skipping over it.

Fixes #1245

## To test

Follow the reproduction steps in #1245. With this PR branch checked out, instead of an endless loop, you should get some error messages in the cli buffer, and the actions should only get processed once.

### DIFF
--- a/classes/abstracts/ActionScheduler_Abstract_QueueRunner.php
+++ b/classes/abstracts/ActionScheduler_Abstract_QueueRunner.php
@@ -86,15 +86,16 @@ abstract class ActionScheduler_Abstract_QueueRunner extends ActionScheduler_Abst
 		 */
 		try {
 			try {
-				$valid_action = false;
+				$valid_action = true;
+
 				do_action( 'action_scheduler_before_execute', $action_id, $context );
 
 				if ( ActionScheduler_Store::STATUS_PENDING !== $this->store->get_status( $action_id ) ) {
+					$valid_action = false;
 					do_action( 'action_scheduler_execution_ignored', $action_id, $context );
 					return;
 				}
 
-				$valid_action = true;
 				do_action( 'action_scheduler_begin_execute', $action_id, $context );
 
 				$action = $this->store->fetch_action( $action_id );


jorgeatorres (approved):
LGTM. Followed the instructions in #1245 and can confirm the endless loop no longer occurs. The actions are correctly marked as failed with the error indicating the true reason (call to undefined function).

Issue #1104, opened 2024-09-23
Slow claim actions update query causing lock wait timeouts


prettyboymp:
## Overview

When there are a large number of pending actions, the query built by ActionScheduler_DBStore::claim_actions() to claim actions can become slow and lock the entire table. While this lock is active, other requests to update actions from Queue_Runner are stalled or may fail due to transaction lock timeouts.

This causes actions to fail due to errors in transitioning their status. In high-usage environments, the queue of pending actions grows as successful queries are delayed while waiting for locks to clear.

## Environment

* WooCommerce and WooCommerce Subscriptions active
* Webhooks are set up for order and subscription creation/update events, running as scheduled actions
* Approximately 26,000 active subscriptions, each renewing weekly, leading to a regular accumulation of 26,000+ pending actions

A script was run to update a decent sized portion of the subscriptions.  This triggered the queueing of a few thousand more pending actions for the webhooks to be processed, resulting in a total of ~31,000 pending actions.

## Issue Detail

Upon reaching ~31000 pending actions and ~2,500,000+ actions total, we started encountering the following fatal error: 

```
Uncaught RuntimeException: Unable to claim actions. Database error: Lock wait timeout exceeded; try restarting transaction.
```

There were other `Lock wait timeout` errors being generated by the database for other queries to update and delete actions, however, those didn't trigger PHP errors because the handling code doesn't check for errors, only the modified count.  

The query causing the long running locks was:
```
UPDATE wp_actionscheduler_actions 
SET claim_id=2074272, last_attempt_gmt='2024-09-18 17:23:58', last_attempt_local='2024-09-18 19:23:58' 
WHERE claim_id = 0 
AND scheduled_date_gmt <= '2024-09-18 17:23:58' 
AND status='pending' 
ORDER BY priority ASC, attempts ASC, scheduled_date_gmt ASC, action_id ASC
LIMIT 10
```
Explain:
```
+------+-------------+----------------------------+-------------+-----------------------------------------------------------------------+-----------------+---------+------+-------+---------------------------------------------------------------+
| id   | select_type | table                      | type        | possible_keys                                                         | key             | key_len | ref  | rows  | Extra                                                         |
+------+-------------+----------------------------+-------------+-----------------------------------------------------------------------+-----------------+---------+------+-------+---------------------------------------------------------------+
[...]

PR #1248, opened 2025-03-17
Only release claims on pending actions.

### LINKED ISSUE #1105: Poor selectivity of claim_id based indexes of actions table when querying unclaimed actions.
The current process of running a batch of actions is followed by the release of the claim on actions.  The release of the claim after running a batch does the following:

1. Updates the claims with the given claim_id from that claim_id to `0`.
2. Delete the claim from the claims table.

Because we're resetting the claim_id back to `0` for all actions, including completed ones that don't need to be reclaimed, it reduces the selectivity of claim_id based indexes when searching for unclaimed actions.  As an example, we have a site with 2.52M actions.  Of those, 2.49M are completed actions and 25K are pending.  This makes it so that almost all of the 2.52M records have a claim_id of `0` increasing the cost to use the `claim_id_status_scheduled_date_gmt` index, especially when trying to claim  pending actions.

## Suggestion

Instead of deleting a claim and resetting all the completed actions back to a claim_id of `0`, only reset non-completed actions back to a claim_id of 0 when releasing a claim.  Removing old claims can then be part of the job to clean old completed actions.  This will increase the cardinality of claim_id column in the actions table, making indexes using it much more useful.

Related #1104

prettyboymp:
Fixes: #1105, improves #1104

The current query for claiming actions can get really slow for a sites that process a ton of actions.  Part of the cause to this is that all actions are reset back to claim_id `0` after a batch is processed.  This increases the amount of time needed to query the pending actions that need to be claimed because the initial segment of the index used to claim actions is heavily seeded with `0` values causing the optimizer to frequently choose a less performant index.

This improves that by no longer resetting the claim_id of completed or running actions back to `0` when a claim is released.  Since only `pending` actions are still processed, leaving these with their old shouldn't cause any side-effects, and if anything improves debugging potential issues since the original claim that processed any actions will still be present.

Testing Instructions:

1. Clear out any existing completed or failed actions that are already set to `claim_id=0` or take note of the current count to verify the results of the changes later:
  ```sql
-- Remove Existing:
DELETE from wp_actionscheduler_actions where claim_id = 0 and status != 'pending';
 -- Get the count:
SELECT count(*) from wp_actionscheduler_actions where claim_id = 0 and status != 'pending';
  ```
2. Generate actions to be processed.  You can use the code below in a `mu-plugins` file and a wp-cli command `wp generate_actions` to create a bunch of pseudo actions to be processed.

```php
/**
 * Actions used for testing.
 */
function schedule_random_action() {

	$foo         = rand( 0, 10000 );
	$priority    = rand( 0, 255 );
	$time_offset = rand( 1, 24 ) * HOUR_IN_SECONDS;

	return as_schedule_single_action( time() + $time_offset, 'as_performance_testing_action', array( 'foo' => $foo ), '', false, $priority );
}

// The action has to have a callback to avoid being marked as failed.
add_action( 'as_performance_testing_action', function() {
  // do nothing.
});


if ( class_exists( 'WP_CLI' ) ) {
	/**
	 * Generate 50K actions for testing.  These are no-op callbacks so that we're testing the load of scheduling, not the actions themselves.
	 *
	 * Example:
	 *  wp generate_actions
	 */
	WP_CLI::add_command(
		'generate_actions',
		function () {
			if ( ! did_action( 'action_scheduler_init' ) ) {
				WP_CLI::error( 'ActionScheduler is not initialized!!!' );
			}

			global $wpdb;
			WP_CLI::line( ' current actions: ' . $wpdb->get_var( "SELECT COUNT(*) FROM wp_actionscheduler_actions" ) );

			$num_to_create = 50000;

			for ( $i = 0; $i < $num_to_create; $i ++ ) {
				if ( ! schedule_random_action() ) {
					WP_CLI::warning( 'action not created' );
				}
			}
			WP_CLI::success( "Created $num_to_create actions" );
		}
	);
}
```
3. Open the admin and allow the queue to process over time.
4. Verify that no new non-pending actions have a claim_id = 0:  `SELECT count(*) from wp_actionscheduler_actions where claim_id = 0 and status != 'pending'`.

Note that these changes will not effect actions that have already been completed or failed as those really no longer matter to the system and are only retained for a period of time for admin review.  They will eventually get cleaned up via a job.

### DIFF
--- a/classes/data-stores/ActionScheduler_DBStore.php
+++ b/classes/data-stores/ActionScheduler_DBStore.php
@@ -1094,7 +1094,7 @@ AND `group_id` = %d
 	}
 
 	/**
-	 * Release actions from a claim and delete the claim.
+	 * Release pending actions from a claim and delete the claim.
 	 *
 	 * @param ActionScheduler_ActionClaim $claim Claim object.
 	 * @throws \RuntimeException When unable to release actions from claim.
@@ -1107,14 +1107,21 @@ AND `group_id` = %d
 		 */
 		global $wpdb;
 
+		if ( 0 === intval( $claim->get_id() ) ) {
+			// Verify that the claim_id is valid before attempting to release it.
+			return;
+		}
+
 		/**
 		 * Deadlock warning: This function modifies actions to release them from claims that have been processed. Earlier, we used to it in a atomic query, i.e. we would update all actions belonging to a particular claim_id with claim_id = 0.
 		 * While this was functionally correct, it would cause deadlock, since this update query will hold a lock on the claim_id_.. index on the action table.
 		 * This allowed the possibility of a race condition, where the claimer query is also running at the same time, then the claimer query will also try to acquire a lock on the claim_id_.. index, and in this case if claim release query has already progressed to the point of acquiring the lock, but have not updated yet, it would cause a deadlock.
 		 *
 		 * We resolve this by getting all the actions_id that we want to release claim from in a separate query, and then releasing the claim on each of them. This way, our lock is acquired on the action_id index instead of the claim_id index. Note that the lock on claim_id will still be acquired, but it will only when we actually make the update, rather than when we select the actions.
+		 *
+		 * We only release pending actions in order for them to be claimed by another process.
 		 */
-		$action_ids = $wpdb->get_col( $wpdb->prepare( "SELECT action_id FROM {$wpdb->actionscheduler_actions} WHERE claim_id = %d", $claim->get_id() ) );
+		$action_ids = $wpdb->get_col( $wpdb->prepare( "SELECT action_id FROM {$wpdb->actionscheduler_actions} WHERE claim_id = %d AND status = %s", $claim->get_id(), self::STATUS_PENDING ) );
 
 		$row_updates = 0;
 		if ( count( $action_ids ) > 0 ) {
--- a/classes/ActionScheduler_wcSystemStatus.php
+++ b/classes/ActionScheduler_wcSystemStatus.php
@@ -76,7 +76,6 @@ class ActionScheduler_wcSystemStatus {
 
 		$action = $this->store->query_actions(
 			array(
-				'claimed'  => false,
 				'status'   => $status,
 				'per_page' => 1,
 				'order'    => $order,
--- a/classes/WP_CLI/System_Command.php
+++ b/classes/WP_CLI/System_Command.php
@@ -262,7 +262,6 @@ class System_Command {
 		$order = 'oldest' === $date_type ? 'ASC' : 'DESC';
 
 		$args = array(
-			'claimed'  => false,
 			'status'   => $status,
 			'per_page' => 1,
 			'order'    => $order,
--- a/classes/data-stores/ActionScheduler_HybridStore.php
+++ b/classes/data-stores/ActionScheduler_HybridStore.php
@@ -432,7 +432,7 @@ class ActionScheduler_HybridStore extends Store {
 	}
 
 	/**
-	 * Release a claim in the table data store.
+	 * Release a claim in the table data store on any pending actions.
 	 *
 	 * @param ActionScheduler_ActionClaim $claim Claim object.
 	 */
--- a/classes/data-stores/ActionScheduler_wpPostStore.php
+++ b/classes/data-stores/ActionScheduler_wpPostStore.php
@@ -791,24 +791,41 @@ class ActionScheduler_wpPostStore extends ActionScheduler_Store {
 	}
 
 	/**
-	 * Release claim.
+	 * Release pending actions from a claim.
 	 *
 	 * @param ActionScheduler_ActionClaim $claim Claim object to release.
 	 * @return void
 	 * @throws RuntimeException When the claim is not unlocked.
 	 */
 	public function release_claim( ActionScheduler_ActionClaim $claim ) {
-		$action_ids = $this->find_actions_by_claim_id( $claim->get_id() );
-		if ( empty( $action_ids ) ) {
-			return; // nothing to do.
-		}
-		$action_id_string = implode( ',', array_map( 'intval', $action_ids ) );
 		/**
 		 * Global wpdb object.
 		 *
 		 * @var wpdb $wpdb
 		 */
 		global $wpdb;
+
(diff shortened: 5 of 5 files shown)


prettyboymp:
What I'm not sure is the potential backward compatibility conflicts here.  I don't know why existing extensions would need completed actions to get reset back to claim_id 0, but we can't be sure that is the case.

joshuatf (commented):
Actions on my end still appear to be reset to `0`.  Here's what I did:

1. Ran `wp generate_actions` which created a lot of `as_performance_testing_action ` actions
2. Manually run the actions by visiting Tools->Actoin Schedule
3. Run the above SQL query and see actions are set to `0` after completion.

Screenshot 2025-03-31 at 4 03 55 PM

Are there any other steps I need to take or that I did wrong?

prettyboymp:
> Actions on my end still appear to be reset to `0`. Here's what I did:
> 
> 1. Ran `wp generate_actions` which created a lot of `as_performance_testing_action ` actions
> 2. Manually run the actions by visiting Tools->Actoin Schedule
> 3. Run the above SQL query and see actions are set to `0` after completion.
> 
> Screenshot 2025-03-31 at 4 03 55 PM
> Are there any other steps I need to take or that I did wrong?

@joshuatf , sorry, I made an assumption that the site wouldn't have any existing actions that were already failed or completed that were reset to claim_id 0.  I've updated the instructions. Can you confirm whether those were new actions or not?

classes/data-stores/ActionScheduler_DBStore.php
```diff
@@ -1107,14 +1107,21 @@ public function release_claim( ActionScheduler_ActionClaim $claim ) {
...
 		 * This allowed the possibility of a race condition, where the claimer query is also running at the same time, then the claimer query will also try to acquire a lock on the claim_id_.. index, and in this case if claim release query has already progressed to the point of acquiring the lock, but have not updated yet, it would cause a deadlock.
 		 *
 		 * We resolve this by getting all the actions_id that we want to release claim from in a separate query, and then releasing the claim on each of them. This way, our lock is acquired on the action_id index instead of the claim_id index. Note that the lock on claim_id will still be acquired, but it will only when we actually make the update, rather than when we select the actions.
+		 *
+		 * We only release pending actions in order for them to be claimed by another process.
 		 */
-		$action_ids = $wpdb->get_col( $wpdb->prepare( "SELECT action_id FROM {$wpdb->actionscheduler_actions} WHERE claim_id = %d", $claim->get_id() ) );
+		$action_ids = $wpdb->get_col( $wpdb->prepare( "SELECT action_id FROM {$wpdb->actionscheduler_actions} WHERE claim_id = %d AND status = %s", $claim->get_id(), self::STATUS_PENDING ) );
```
joshuatf, line 1124: I can confirm that this array is empty and the below query to set the `claim_id = 0` does not get run.  However, my claims in the DB are still getting set to `0`.

Is there anywhere else this might be happening or is it possible for multiple instances of AS to be running at once?

joshuatf (commented):
Hey @prettyboymp, sorry for the late review.  This is happening on new actions, though it doesn't seem related to the area you touched.

The code you've added otherwise LGTM; I'm just not entirely sure where claims are being released in my case.  

It might be good to get someone else more familiar with AS to take a glance at this and give final approval, especially around any potential backwards compatibility issues. cc @jorgeatorres

classes/data-stores/ActionScheduler_DBStore.php
jorgeatorres: Note that actions processed via Tools > Scheduled Actions are not claimed/released. Is this how you're running the actions @joshuatf?

For the runner to properly stake a claim and then release the actions, you'd have to either let those actions run naturally or use a runner such as the CLI one (i.e. run `wp action-scheduler run`).

jorgeatorres (commented):
This looks good to me @prettyboymp and tests well. I like the performance improvements this can bring and also that by retaining the claim ID we've gained a bit on the auditing side, which is always useful.

I'm not _too_ concerned about the backwards compatibility of this change in the sense that 3rd parties _need_ completed actions to have `claim_id = 0`. In theory they shouldn't necessarily be querying the database directly.

That said, we do have a `claimed` field that can be used to query actions via `as_get_scheduled_actions()` (implemented as `query_actions()` in the data stores). We would be changing the behavior here as `claimed = true` wouldn't have returned completed actions prior to this change. Maybe we can "force" the query function to ignore non-pending actions when the `claimed = true` arg is provided.

FWIW, this arg is used in a few places, and it'd be good to check that we're not breaking anything in those places. In particular, the [queue cleaner seems fine](https://github.com/woocommerce/action-scheduler/blob/2720a770e5bd809ebf4bd50d5db477ceb2ea1787/classes/ActionScheduler_QueueCleaner.php#L185), but other integrations such as the one with WC's SSR [might need some adjustments](https://github.com/woocommerce/action-scheduler/blob/12f09c1b581f113957f42a7db414ed5cf38fb7c7/classes/ActionScheduler_wcSystemStatus.php#L77-L84).

What do you think?

prettyboymp:
@jorgeatorres, I updated the status reports to no longer include the `claimed` parameter since it doesn't really make sense to have them applied when searching for the oldest/newest of each action status.

> That said, we do have a claimed field that can be used to query actions via as_get_scheduled_actions() (implemented as query_actions() in the data stores). We would be changing the behavior here as claimed = true wouldn't have returned completed actions prior to this change. Maybe we can "force" the query function to ignore non-pending actions when the claimed = true arg is provided.

I’m a bit hesitant to introduce forced manipulation of the `status` query argument for special cases where `claimed` is set, mainly because I’m not aware of any use cases that would be impacted by completed actions remaining in a claimed state—aside from scenarios explicitly trying to retrieve currently claimed actions. Even in those cases, unless the `status` parameter is also specified, the query could return non-pending actions while a claim is still active, since the claim is only released after all actions are processed.

I also reviewed all plugins on WooCommerce.com and didn’t find any instances where `claimed` was used without a specific `status` parameter, apart from what we’ve already accounted for within Action Scheduler itself.

jorgeatorres (approved):
Thanks for the changes @prettyboymp! This looks good to me.

I'm not merging yet in case @joshuatf wants to give it another look before.

PR #1259, opened 2025-03-28
Implement SKIP LOCKED during action claiming

### LINKED ISSUE #1104: Slow claim actions update query causing lock wait timeouts

WPprodigy:
### Avoiding SQL deadlocks during action claiming

When running multiple action scheduler queues concurrently, many will start to fatal error due to SQL deadlocks. I have found that this always stems from the action claims query. This query ends up scanning many rows, and being an UPDATE, it ends up creating locks on the table/rows. This results in concurrent claims queries conflicting with each other and deadlocking. It also causes deadlocks during individual action status updates from other queues that are already processing actions. This leaves actions in unknown states and results in lots needing to be failed later on due to timeouts during the cleanup periods.

This PR prevents the deadlocks altogether. The magic here comes from the `SKIP LOCKED` SQL feature: https://dev.mysql.com/blog-archive/mysql-8-0-1-using-skip-locked-and-nowait-to-handle-hot-rows/. It's a perfect fit for the claims query, because we don't really care if it needs to skip over a few rows. All we want is for it to claim actions without conflict.

The `SKIP LOCKED` has to come from a SELECT query. So the new query here uses a SELECT sub-query (w/ `FOR UPDATE` to claim a lock) and then JOINs the results to accomplish first selecting the items and then updating them all within a single transaction.

I've done a lot of testing with this and ran into no deadlocks or failing actions. I've run hundreds of thousands of actions at 250 queue concurrency without any signs of trouble.

```
BEFORE (causes deadlocks):

UPDATE wp_actionscheduler_actions SET claim_id=4322, last_attempt_gmt='2025-01-15 03:24:52', last_attempt_local='2025-01-14 21:24:53' WHERE claim_id = 0 AND scheduled_date_gmt <= '2025-01-15 03:23:18' AND status='pending' ORDER BY priority ASC, attempts ASC, scheduled_date_gmt ASC, action_id ASC LIMIT 20;

AFTER (no deadlocks):

UPDATE wp_actionscheduler_actions t1 JOIN ( SELECT action_id from wp_actionscheduler_actions WHERE claim_id = 0 AND scheduled_date_gmt <= '2025-01-15 03:24:52' AND status='pending' ORDER BY priority ASC, attempts ASC, scheduled_date_gmt ASC, action_id ASC LIMIT 20 FOR UPDATE SKIP LOCKED ) t2 ON t1.action_id = t2.action_id SET claim_id=4322, last_attempt_gmt='2025-01-15 03:24:52', last_attempt_local='2025-01-14 21:24:53';
```

### Considerations

I'm not necessarily suggesting we merge this PR as is (leaving as a draft for now), but I wanted to show the solution here and what I've found. We'll need to test and discover database compatibility before shipping this as the new default behavior.

`SKIP LOCKED` shipped with MySQL 8.0.1 (2017-04-10) and MariaDB in 10.6.0 (2021-4-26), and works only for InnoDB tables (ignored otherwise). I've only done the testing in MySQL only, and I'm not sure if it's safely ignored on the older versions or if it would cause issues.

I do think this would definitely be a worthwhile change to get into the plugin, as I've seen it be a source of a lot of issues. Some possible rollout options:

A) Put it into a new method (`claim_actions_v2()`), and use the new method when a feature flag is present via some constant or filter.
B) Check the database version, and automatically use the v2 method when it's modern enough.

### DIFF
--- a/classes/ActionScheduler_DataController.php
+++ b/classes/ActionScheduler_DataController.php
@@ -162,10 +162,19 @@ class ActionScheduler_DataController {
 			return;
 		}
 
-		$wp_object_cache->group_ops      = array();
-		$wp_object_cache->stats          = array();
-		$wp_object_cache->memcache_debug = array();
-		$wp_object_cache->cache          = array();
+		// Not all drop-ins support these props, however, there may be existing installations that rely on these being cleared.
+		if ( property_exists( $wp_object_cache, 'group_ops' ) ) {
+			$wp_object_cache->group_ops = array();
+		}
+		if ( property_exists( $wp_object_cache, 'stats' ) ) {
+			$wp_object_cache->stats = array();
+		}
+		if ( property_exists( $wp_object_cache, 'memcache_debug' ) ) {
+			$wp_object_cache->memcache_debug = array();
+		}
+		if ( property_exists( $wp_object_cache, 'cache' ) ) {
+			$wp_object_cache->cache = array();
+		}
 
 		if ( is_callable( array( $wp_object_cache, '__remoteset' ) ) ) {
 			call_user_func( array( $wp_object_cache, '__remoteset' ) ); // important!
--- a/classes/data-stores/ActionScheduler_DBStore.php
+++ b/classes/data-stores/ActionScheduler_DBStore.php
@@ -932,15 +932,12 @@ AND `group_id` = %d
 		 */
 		global $wpdb;
 
+		// Start by building a SELECT query for which actions we want to claim.
+		$select = "SELECT action_id from {$wpdb->actionscheduler_actions}";
+		$select_params = [];
+
 		$now  = as_get_datetime_object();
 		$date = is_null( $before_date ) ? $now : clone $before_date;
-		// can't use $wpdb->update() because of the <= condition.
-		$update = "UPDATE {$wpdb->actionscheduler_actions} SET claim_id=%d, last_attempt_gmt=%s, last_attempt_local=%s";
-		$params = array(
-			$claim_id,
-			$now->format( 'Y-m-d H:i:s' ),
-			current_time( 'mysql' ),
-		);
 
 		// Set claim filters.
 		if ( ! empty( $hooks ) ) {
@@ -954,14 +951,14 @@ AND `group_id` = %d
 			$group = $this->get_claim_filter( 'group' );
 		}
 
-		$where    = 'WHERE claim_id = 0 AND scheduled_date_gmt <= %s AND status=%s';
-		$params[] = $date->format( 'Y-m-d H:i:s' );
-		$params[] = self::STATUS_PENDING;
+		$where           = 'WHERE claim_id = 0 AND scheduled_date_gmt <= %s AND status=%s';
+		$select_params[] = $date->format( 'Y-m-d H:i:s' );
+		$select_params[] = self::STATUS_PENDING;
 
 		if ( ! empty( $hooks ) ) {
-			$placeholders = array_fill( 0, count( $hooks ), '%s' );
-			$where       .= ' AND hook IN (' . join( ', ', $placeholders ) . ')';
-			$params       = array_merge( $params, array_values( $hooks ) );
+			$placeholders  = array_fill( 0, count( $hooks ), '%s' );
+			$where        .= ' AND hook IN (' . join( ', ', $placeholders ) . ')';
+			$select_params = array_merge( $select_params, array_values( $hooks ) );
 		}
 
 		$group_operator = 'IN';
@@ -1003,11 +1000,21 @@ AND `group_id` = %d
 		 * @param string $claim_id Claim Id.
 		 * @param array  $hooks Hooks to filter for.
 		 */
-		$order    = apply_filters( 'action_scheduler_claim_actions_order_by', 'ORDER BY priority ASC, attempts ASC, scheduled_date_gmt ASC, action_id ASC', $claim_id, $hooks );
-		$params[] = $limit;
+		$order           = apply_filters( 'action_scheduler_claim_actions_order_by', 'ORDER BY priority ASC, attempts ASC, scheduled_date_gmt ASC, action_id ASC', $claim_id, $hooks );
+		$select_params[] = $limit;
 
-		$sql           = $wpdb->prepare( "{$update} {$where} {$order} LIMIT %d", $params ); // phpcs:ignore WordPress.DB.PreparedSQL.InterpolatedNotPrepared, WordPress.DB.PreparedSQLPlaceholders
-		$rows_affected = $wpdb->query( $sql ); // phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared, WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching
+		// Selecting the action_ids that we plan to claim, while skipping any locked rows to avoid deadlocking.
+		$select_sql = $wpdb->prepare( "{$select} {$where} {$order} LIMIT %d FOR UPDATE SKIP LOCKED", $select_params );
+
+		// Now place it into an UPDATE statement by joining the result sets, allowing for the SKIP LOCKED behavior to take effect.
+		$update_sql    = "UPDATE {$wpdb->actionscheduler_actions} t1 JOIN ( $select_sql ) t2 ON t1.action_id = t2.action_id SET claim_id=%d, last_attempt_gmt=%s, last_attempt_local=%s";
+		$update_params = array(
+			$claim_id,
+			$now->format( 'Y-m-d H:i:s' ),
+			current_time( 'mysql' ),
+		);
+
+		$rows_affected = $wpdb->query( $wpdb->prepare( $update_sql, $update_params ) );
 		if ( false === $rows_affected ) {
 			$error = empty( $wpdb->last_error )
 				? _x( 'unknown', 'database error', 'action-scheduler' )
--- a/classes/data-stores/ActionScheduler_DBStore.php
+++ b/classes/data-stores/ActionScheduler_DBStore.php
@@ -1029,7 +1029,7 @@ AND `group_id` = %d
 	}
 
 	/**
-	 * Determines whether the database supports using SKIP LOCKED.
+	 * Determines whether the database supports using SKIP LOCKED. This logic mimicks the $wpdb::has_cap() logic.
 	 *
 	 * SKIP_LOCKED support was added to MariaDB in 10.6.0 and to MySQL in 8.0.1
 	 *
--- a/classes/data-stores/ActionScheduler_DBStore.php
+++ b/classes/data-stores/ActionScheduler_DBStore.php
@@ -1039,7 +1039,7 @@ AND `group_id` = %d
 		global $wpdb;
 		$db_version     = $wpdb->db_version();
 		$db_server_info = $wpdb->db_server_info();
-		$is_mariadb     = ( false !== str_contains( $db_server_info, 'MariaDB' ) );
+		$is_mariadb     = ( false !== strpos( $db_server_info, 'MariaDB' ) );
 
 		if ( $is_mariadb &&
 		     '5.5.5' === $db_version &&
--- a/classes/data-stores/ActionScheduler_DBStore.php
+++ b/classes/data-stores/ActionScheduler_DBStore.php
@@ -931,11 +931,6 @@ AND `group_id` = %d
 		 * @var \wpdb $wpdb
 		 */
 		global $wpdb;
-
-		// Start by building a SELECT query for which actions we want to claim.
-		$select = "SELECT action_id from {$wpdb->actionscheduler_actions}";
-		$select_params = [];
-
 		$now  = as_get_datetime_object();
 		$date = is_null( $before_date ) ? $now : clone $before_date;
 
@@ -951,14 +946,16 @@ AND `group_id` = %d
 			$group = $this->get_claim_filter( 'group' );
 		}
 
-		$where           = 'WHERE claim_id = 0 AND scheduled_date_gmt <= %s AND status=%s';
-		$select_params[] = $date->format( 'Y-m-d H:i:s' );
-		$select_params[] = self::STATUS_PENDING;
+		$where        = 'WHERE claim_id = 0 AND scheduled_date_gmt <= %s AND status=%s';
+		$where_params = array(
+			$date->format( 'Y-m-d H:i:s' ),
+			self::STATUS_PENDING,
+		);
 
 		if ( ! empty( $hooks ) ) {
-			$placeholders  = array_fill( 0, count( $hooks ), '%s' );
+			$placeholders = array_fill( 0, count( $hooks ), '%s' );
 			$where        .= ' AND hook IN (' . join( ', ', $placeholders ) . ')';
-			$select_params = array_merge( $select_params, array_values( $hooks ) );
+			$where_params = array_merge( $where_params, array_values( $hooks ) );
 		}
 
 		$group_operator = 'IN';
@@ -993,18 +990,18 @@ AND `group_id` = %d
 		/**
 		 * Sets the order-by clause used in the action claim query.
 		 *
-		 * @since 3.4.0
-		 * @since 3.8.3 Made $claim_id and $hooks available.
-		 *
 		 * @param string $order_by_sql
 		 * @param string $claim_id Claim Id.
-		 * @param array  $hooks Hooks to filter for.
+		 * @param array  $hooks    Hooks to filter for.
+		 *
+		 * @since 3.8.3 Made $claim_id and $hooks available.
+		 * @since 3.4.0
 		 */
-		$order           = apply_filters( 'action_scheduler_claim_actions_order_by', 'ORDER BY priority ASC, attempts ASC, scheduled_date_gmt ASC, action_id ASC', $claim_id, $hooks );
-		$select_params[] = $limit;
+		$order       = apply_filters( 'action_scheduler_claim_actions_order_by', 'ORDER BY priority ASC, attempts ASC, scheduled_date_gmt ASC, action_id ASC', $claim_id, $hooks );
+		$skip_locked = $this->db_supports_skip_locked() ? ' SKIP LOCKED' : '';
 
 		// Selecting the action_ids that we plan to claim, while skipping any locked rows to avoid deadlocking.
-		$select_sql = $wpdb->prepare( "{$select} {$where} {$order} LIMIT %d FOR UPDATE SKIP LOCKED", $select_params );
+		$select_sql = $wpdb->prepare( "SELECT action_id from {$wpdb->actionscheduler_actions} {$where} {$order} LIMIT %d FOR UPDATE{$skip_locked}", array_merge( $where_params, array( $limit ) ) );
 
 		// Now place it into an UPDATE statement by joining the result sets, allowing for the SKIP LOCKED behavior to take effect.
 		$update_sql    = "UPDATE {$wpdb->actionscheduler_actions} t1 JOIN ( $select_sql ) t2 ON t1.action_id = t2.action_id SET claim_id=%d, last_attempt_gmt=%s, last_attempt_local=%s";
@@ -1019,7 +1016,6 @@ AND `group_id` = %d
 			$error = empty( $wpdb->last_error )
 				? _x( 'unknown', 'database error', 'action-scheduler' )
 				: $wpdb->last_error;
-
 			throw new \RuntimeException(
 				sprintf(
 					/* translators: %s database error. */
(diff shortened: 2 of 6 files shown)


classes/ActionScheduler_DataController.php
```diff
@@ -162,10 +162,19 @@ public static function free_memory() {
...
+		}
+		if ( property_exists( $wp_object_cache, 'stats' ) ) {
+			$wp_object_cache->stats = array();
+		}
+		if ( property_exists( $wp_object_cache, 'memcache_debug' ) ) {
+			$wp_object_cache->memcache_debug = array();
+		}
+		if ( property_exists( $wp_object_cache, 'cache' ) ) {
+			$wp_object_cache->cache = array();
+		}
```
prettyboymp, line 177: These changes aren't directly related to this PR, but the CLI commands being spawned when testing with https://github.com/woocommerce/as-pressure-cooker were causing a bunch of errors/warnings.

classes/schema/ActionScheduler_StoreSchema.php
```diff
@@ -80,7 +80,9 @@ protected function get_table_definition( $table ) {
 				        KEY args (args($max_index_length)),
 				        KEY group_id (group_id),
 				        KEY last_attempt_gmt (last_attempt_gmt),
-				        KEY `claim_id_status_scheduled_date_gmt` (`claim_id`, `status`, `scheduled_date_gmt`)
+				        KEY `claim_id_status_priority_scheduled_date_gmt` (`claim_id`,`status`,`priority`,`scheduled_date_gmt`),
+				        KEY `status_last_attempt_gmt` (`status`,`last_attempt_gmt`),
+				        KEY `status_claim_id` (`status`,`claim_id`)
```
prettyboymp, line 85: Only `claim_id_status_priority_scheduled_date_gmt` is specific to the claim process specific to this PR, however, the other two queries kept coming up as slow queries when testing at a high load.

- claim_id_status_priority_scheduled_date_gmt - improved index for staking claims on actions.
- status_last_attempt_gmt - used for QueueCleaner::clean_actions()
- status_claim_id - used calculate distinct claim_ids to determine current claims being processed.

tests/phpunit/procedural_api/procedural_api_Test.php
```diff
@@ -449,7 +449,7 @@ public function test_as_recover_from_incorrect_schema() {
 		$this->assertContains( 'Caught exception while enqueuing action "hook_18": Error saving action', $logged_errors );
 		$this->assertContains( 'Caught exception while enqueuing action "hook_19": Error saving action', $logged_errors );
 		$this->assertContains( 'Caught exception while enqueuing action "hook_20": Error saving action', $logged_errors );
-		$this->assertContains( "Unknown column 'priority' in 'field list'", $logged_errors );
+		$this->assertContains( "Unknown column 'priority' in ", $logged_errors );
```
prettyboymp, line 452: The format of this error is slightly different across supported DBs.

prettyboymp:
I've done some testing against this and am happy with the results.  So I've made some additional changes to deal with compatibility:

- Claim process always use the `UPDATE .... JOIN (SELECT ... FOR UPDATE ... WHERE.... ORDER BY...)` format (without `SKIP LOCKED` if not supported) rather than the previous `UPDATE... WHERE... ORDER BY...`.   

  This alone should get rid of most deadlocks since the latter is non-deterministic because the db doesn't lock the rows until attempting to update them when another process may have already locked a subset of the rows.  By moving to a `SELECT .. FOR UPDATE` the locks are set during the select process and claimed in a known order.  This can still result in long running lock waits for claim queries, but they should at least not result in deadlocks.

- Use the `SKIP LOCKED` modifier on the the sub-select only on DB versions that support it.  

  This should further improve performance for DB versions that support it by preventing the claim process from stacking caused by multiple concurrent processes attempting to wait on locks to clear by just continuing with the next records.

## Testing Instructions

1. Setup a site running Action Scheduler - it can be installed as a separate plugin.  If you're running other plugins with ActionScheduler, you'll need to either disable those or apply the following patch to make sure this version loads:

```
diff --git a/action-scheduler.php b/action-scheduler.php
index ef12ca2..8d04084 100644
--- a/action-scheduler.php
+++ b/action-scheduler.php
@@ -5,8 +5,8 @@
  * Author: Automattic
- * Version: 3.9.2
- * License: GPLv3
+ * Version: 3.9.3
+ * License: GPLv
  * Requires at least: 6.5
  * Tested up to: 6.7
  * Requires PHP: 7.1
@@ -29,29 +29,29 @@
  * @package ActionScheduler
  */

-if ( ! function_exists( 'action_scheduler_register_3_dot_9_dot_2' ) && function_exists( 'add_action' ) ) { // WRCS: DEFINED_VERSION.
+if ( ! function_exists( 'action_scheduler_register_3_dot_9_dot_3' ) && function_exists( 'add_action' ) ) { // WRCS: DEFINED_VERSION.

        if ( ! class_exists( 'ActionScheduler_Versions', false ) ) {
                require_once __DIR__ . '/classes/ActionScheduler_Versions.php';
                add_action( 'plugins_loaded', array( 'ActionScheduler_Versions', 'initialize_latest_version' ), 1, 0 );
        }

-       add_action( 'plugins_loaded', 'action_scheduler_register_3_dot_9_dot_2', 0, 0 ); // WRCS: DEFINED_VERSION.
+       add_action( 'plugins_loaded', 'action_scheduler_register_3_dot_9_dot_3', 0, 0 ); // WRCS: DEFINED_VERSION.

        // phpcs:disable Generic.Functions.OpeningFunctionBraceKernighanRitchie.ContentAfterBrace
        /**
         * Registers this version of Action Scheduler.
         */
-       function action_scheduler_register_3_dot_9_dot_2() { // WRCS: DEFINED_VERSION.
+       function action_scheduler_register_3_dot_9_dot_3() { // WRCS: DEFINED_VERSION.
-               $versions->register( '3.9.2', 'action_scheduler_initialize_3_dot_9_dot_2' ); // WRCS: DEFINED_VERSION.
+               $versions->register( '3.9.3', 'action_scheduler_initialize_3_dot_9_dot_3' ); // WRCS: DEFINED_VERSION.
        }

        // phpcs:disable Generic.Functions.OpeningFunctionBraceKernighanRitchie.ContentAfterBrace
        /**
         * Initializes this version of Action Scheduler.
         */
-       function action_scheduler_initialize_3_dot_9_dot_2() { // WRCS: DEFINED_VERSION.
+       function action_scheduler_initialize_3_dot_9_dot_3() { // WRCS: DEFINED_VERSION.
                // A final safety check is required even here, because historic versions of Action Scheduler
                // followed a different pattern (in some unusual cases, we could reach this point and the
                // ActionScheduler class is already defined—so we need to guard against that).
@@ -63,7 +63,7 @@ if ( ! function_exists( 'action_scheduler_register_3_dot_9_dot_2' ) && function_

        if ( did_action( 'plugins_loaded' ) && ! doing_action( 'plugins_loaded' ) && ! class_exists( 'ActionScheduler', false ) ) {
-               action_scheduler_initialize_3_dot_9_dot_2(); // WRCS: DEFINED_VERSION.
+               action_scheduler_initialize_3_dot_9_dot_3(); // WRCS: DEFINED_VERSION.
                do_action( 'action_scheduler_pre_theme_init' );
                ActionScheduler_Versions::initialize_latest_version();
        }
```
2. Install [AS Pressure Cooker](https://github.com/woocommerce/as-pressure-cooker) as a plugin.  This PR has some tweaks/fixes for PHP8 and improved details on the report: https://github.com/woocommerce/as-pressure-cooker/pull/6
3. Watch your site's error_log and slow query log (if available).
4. Run the AS Pressure Cooker to test the changes:
  ```
  wp action-scheduler pressure-cooker --starting_queue_runners=1 --finishing_queue_runners=30 --initial_past_due_actions=60000
  ```

Notes when testing:
* The pressure cooker will start with `--starting_queue_runners` and end with `--finishing_queue_runners`, allowing you to see where errors start occurring, if any.
* You may be limited in how many runners you can test with based on the memory and connection limits of your environment.  So it may be a good idea to start with a lower number to see where your environment breaks down and then run a full test just below that limit.
* `--initial_past_due_actions` determines how many actions are created prior to starting the test.  More will be generated during the test, however, that generation may not be able to keep up with the processing.  So it is a good idea to set this high enough to make sure the processing rate isn't brought down by there not being enough actions to be process.

## Testing Results

I ran the following tests locally with the below diff applied to better simulate how action scheduler runs normally through requests and the limitation of an assumed 30s request timeout and forcing the clean up queries to be run.

```
diff --git a/classes/WP_CLI/ActionScheduler_WPCLI_QueueRunner.php b/classes/WP_CLI/ActionScheduler_WPCLI_QueueRunner.php
index 08cb0cd..150b470 100644
--- a/classes/WP_CLI/ActionScheduler_WPCLI_QueueRunner.php
+++ b/classes/WP_CLI/ActionScheduler_WPCLI_QueueRunner.php
@@ -60,6 +60,9 @@ class ActionScheduler_WPCLI_QueueRunner extends ActionScheduler_Abstract_QueueRu
 	 */
 	public function setup( $batch_size, $hooks = array(), $group = '', $force = false ) {
+		if( $this->get_execution_time() >= 28 ) {
+			return 0;
+		}
 		$this->run_cleanup();
 		$this->add_hooks();
 
@@ -120,6 +123,10 @@ class ActionScheduler_WPCLI_QueueRunner extends ActionScheduler_Abstract_QueueRu
 
 			$this->process_action( $action_id, $context );
 			$this->progress_bar->tick();
+
+			if( $this->get_execution_time() >= 28 ) {
+				break;
+			}
 		}
 
 		$completed = $this->progress_bar->current();
```

### This PR - MySQL 8.0.4 - InnoDB Engine

```
wp action-scheduler pressure-cooker --starting_queue_runners=30 --finishing_queue_runners=30 --initial_past_due_actions=60000

Deleting existing actions and claims.
Generating test actions  100% [===================================================================================================] 1:34 / 1:35
Created 60000 past due actions.

Elapsed:              120s (100%)
Generated actions:    64400
Processed actions:    62180
Processing Rate:      518.2/s
Active queue runners: 28/30
Replacement rate      6/28
Database errors:      none detected

Elapsed:              120s (100%)
Generated actions:    64343
Processed actions:    52066
Processing Rate:      433.9/s
Active queue runners: 30/30
Replacement rate      0/30
Database errors:      none detected

Test completed.
```

### TRUNK - MySQL 8.0.4 - InnoDB Engine

```
wp action-scheduler pressure-cooker --starting_queue_runners=30 --finishing_queue_runners=30 --initial_past_due_actions=60000

Deleting existing actions and claims.
Generating test actions  100% [===================================================================================================] 1:29 / 1:29
Created 60000 past due actions.

Elapsed:              120s (100%)
Generated actions:    64400
Processed actions:    31006 (654 fails)
Processing Rate:      258.4/s
Active queue runners: 29/30
Replacement rate      0/29
Database errors:      766 (766 deadlocks)

Test completed.
```

### This PR - MySQL 8.0.4 - MyISAM Engine (SKIP LOCKED ignored)

This was really just run to verify no issues were caused on systems using MyISAM engines.

```
wp action-scheduler pressure-cooker --starting_queue_runners=30 --finishing_queue_runners=30 --initial_past_due_actions=60000

Deleting existing actions and claims.
Generating test actions  100% [===================================================================================================] 1:29 / 1:29
Created 60000 past due actions.

Elapsed:              120s (100%)
Generated actions:    64324
Processed actions:    16707
Processing Rate:      139.2/s
Active queue runners: 30/30
Replacement rate      0/30
Database errors:      none detected

Test completed.
```

classes/data-stores/ActionScheduler_DBStore.php
```diff
@@ -1025,6 +1028,36 @@ protected function claim_actions( $claim_id, $limit, ?DateTime $before_date = nu
...
 	}
 
+	/**
+	 * Determines whether the database supports using SKIP LOCKED.
+	 *
+	 * SKIP_LOCKED support was added to MariaDB in 10.6.0 and to MySQL in 8.0.1
+	 *
+	 * @return bool
+	 */
+	private function db_supports_skip_locked() {
```
WPprodigy: Should maybe cache this check so it doesn't have to query and confirm the version support before every claim. Can do local runtime cache, but I'd prolly just do object cache with some relatively short TTL like 5 minutes even.

classes/schema/ActionScheduler_StoreSchema.php
WPprodigy: I would think that `status_claim_id` should be covered by `claim_id_status_priority_scheduled_date_gmt` since both claim id and status are the first two columns in the index.

For `QueueCleaner::clean_actions()`, this is worth a separate discussion/PR probably but this is the other enhancement that I've hacked into the plugin for high queue concurrency. It ends up trying to clean too often with high queue concurrency. Simply placing behind a `wp_cache_add()` style lock with a couple minutes TTL really helps cut down on the queries that have to run (and often conflict with each). It's really not built for concurrency, because if two new queues try to cleanup at the same time, they can both get a similar list of actions to delete/change status, and then they end up fighting through the list and can end up with DB errors when the other queue got to the action first.

In general, I'd be careful adding too many indexes to the table as it'll slow down the write times and possibly lead to worsening deadlock contention (though `SKIP LOCKED` really is awesome 😎)

classes/data-stores/ActionScheduler_DBStore.php
prettyboymp: @WPprodigy, I don't think that is necessary.  My `mysqli::$server_info` is a read-only property of the underlying `mysqli` object.  I can't find documentation explicitly saying it, but I believe it's set when the connection is made or at the very least would be saved within the connected `mysqli` object property between calls. https://www.php.net/manual/en/class.mysqli.php

classes/data-stores/ActionScheduler_DBStore.php
```diff
@@ -996,23 +990,32 @@ protected function claim_actions( $claim_id, $limit, ?DateTime $before_date = nu
...
-		 * @param array  $hooks Hooks to filter for.
+		 * @param array  $hooks    Hooks to filter for.
+		 *
+		 * @since 3.8.3 Made $claim_id and $hooks available.
+		 * @since 3.4.0
 		 */
-		$order    = apply_filters( 'action_scheduler_claim_actions_order_by', 'ORDER BY priority ASC, attempts ASC, scheduled_date_gmt ASC, action_id ASC', $claim_id, $hooks );
-		$params[] = $limit;
+		$order       = apply_filters( 'action_scheduler_claim_actions_order_by', 'ORDER BY priority ASC, attempts ASC, scheduled_date_gmt ASC, action_id ASC', $claim_id, $hooks );
+		$skip_locked = $this->db_supports_skip_locked() ? ' SKIP LOCKED' : '';
```
kalessil, line 1001: Sweet memories of Oracle :smile: If the MySQL/MariaDB implementation is equally good, we scored big by adopting it for this project!

classes/data-stores/ActionScheduler_DBStore.php
```diff
@@ -1025,6 +1028,43 @@ protected function claim_actions( $claim_id, $limit, ?DateTime $before_date = nu
...
 	}
 
+	/**
+	 * Determines whether the database supports using SKIP LOCKED.
+	 *
+	 * SKIP_LOCKED support was added to MariaDB in 10.6.0 and to MySQL in 8.0.1
+	 *
+	 * @return bool
+	 */
+	private function db_supports_skip_locked() {
```
kalessil, line 1038: Are we guaranteed to have InnoDB engine for the scheduler tables?

classes/data-stores/ActionScheduler_DBStore.php
```diff
@@ -1025,6 +1028,43 @@ protected function claim_actions( $claim_id, $limit, ?DateTime $before_date = nu
...
+		global $wpdb;
```
kalessil, line 1039: It looks a bit complicated to me. 

Dropping an idea here  to try running a probe `SELECT action_id FROM wp_actionscheduler_actions WHERE action_id=-1 FOR UPDATE SKIP LOCKED`— if it succeeds, consider the locking is supoprted. This would need some caching around, of course.

kalessil (commented):
Starting to dive in this PR: leaving some questions in comments

classes/data-stores/ActionScheduler_DBStore.php
prettyboymp: No, but MyISAM, or other non-transaction engines, will essentially ignore the SKIP LOCKED hint as long as the RBDMS supports it.   The update will be blocking no matter what since MyISAM does a full table lock.  I've tested this by running the as-pressure-cooker while having the actions table converted to MyISAM.

classes/data-stores/ActionScheduler_DBStore.php
prettyboymp: That was my first consideration since it gave more confidence that it was supported, but I was concerned with generating database errors on purpose.

However, I found that WordPress core does similar checks and mimicked it's implementation, e.g. 

https://github.com/WordPress/wordpress-develop/blob/71b2dd282aeb9d33d169f7b2dbab0b896a9d1068/src/wp-includes/class-wpdb.php#L4065-L4082

classes/data-stores/ActionScheduler_DBStore.php
kalessil: :+1:

classes/data-stores/ActionScheduler_DBStore.php
kalessil: Gotcha! Can we add the reference to the core method here for other folks?

kalessil: Also, following the link reminded me of custom DB-dropins. Did we have a chance to test this PR with our infra drop-ins?

kalessil (commented):
Thanks for all the answers @prettyboymp! I'll do the second review round.
Can you add some testing steps (I guess with the pressure cooker) so I can try them in my environment?

classes/data-stores/ActionScheduler_DBStore.php
prettyboymp: I don't have a local setup that will allow me to easily test it with HyperDB.  

However, I don't know of a scenario where it would cause issues.  HyperDB overrides `::db_server_info()`, but also adds an argument to it allowing the DBH or table name to be specified to use a different DBH than the last one connected to.  Because of the order of execution where the claim_id is inserted just before claiming actions, the default DBH should always be the same one that we need to check against.

prettyboymp:
> Thanks for all the answers @prettyboymp! I'll do the second review round. Can you add some testing steps (I guess with the pressure cooker) so I can try them in my environment?

Sorry, the testing instructions on in this comment: https://github.com/woocommerce/action-scheduler/pull/1259#issuecomment-2784466619

classes/schema/ActionScheduler_StoreSchema.php
prettyboymp: > I would think that status_claim_id should be covered by claim_id_status_priority_scheduled_date_gmt since both claim id and status are the first two columns in the index.

Normally `status_claim_id` is good enough, but it quickly falls apart in cases where there is a large backlog of overdue actions.  For example, when webhooks are setup and any sort of large scale batch processing happens like imports or API batching, the queued webhooks can quickly flood past due actions.  Since the claim query checks priority before scheduled date, the performance drops dramatically.

> In general, I'd be careful adding too many indexes to the table as it'll slow down the write times and possibly lead to worsening deadlock contention

Yes, I'd like to remove some of the current ones but need to save that for another PR since more research needs to be involved in determining how they are used.

kalessil (approved):
LGTM!

I tried my best running the tests, but the pressure cooker is not working properly under our default wp-env cli-container:

```log
Deleting existing actions and claims.
Generating test actions  100% [===================================================================================================================================================================================================================================================================] 1:26 / 1:26
Created 60000 past due actions.

Elapsed:              120s (100%)
Generated actions:    60000
Processed actions:    0
Processing Rate:      0/s
Active queue runners: 29/30
Replacement rate      15/29
Database errors:      none detected
```

Hence I'll rely to passing CI and not hold this PR further.

PR #1260, opened 2025-04-09
Add hook specifically for scheduling recurring actions.

prettyboymp:
# Changes proposed in this Pull Request

This is an upstream implementation of https://github.com/woocommerce/woocommerce/pull/56380.

The goal of this change is to provide a hook for other plugins and extensions to use specifically for ensuring that recurring actions are scheduled.

## The Problem

The current documentation for scheduling a recurring action is as follows:

```
/**
 * Schedule an action with the hook 'eg_midnight_log' to run at midnight each day
 * so that our callback is run then.
 */
function eg_schedule_midnight_log() {
	if ( false === as_has_scheduled_action( 'eg_midnight_log' ) ) {
		as_schedule_recurring_action( strtotime( 'tomorrow' ), DAY_IN_SECONDS, 'eg_midnight_log', array(), '', true );
	}
}
add_action( 'init', 'eg_schedule_midnight_log' );
```

Ideally, the inner part of the `eg_schedule_midnight_log()` function above would only be run during certain events like plugin activation or updates, however, there are environments where the activation/update flow are not used or crash scenarios could cause a recurring action to not get properly rescheduled.  So, the above pattern is used to guarantee those recurring actions required by the plugin are and stay scheduled.

There are some performance problems with this in that `as_has_scheduled_action( 'eg_midnight_log' )` requires a query to be run on every request and while this is mostly a high performant query, having every plugin run it for all of their scheduled recurring actions can add a bit of overhead to a request.

## Proposed Solution

This PR introduces a new hook called `action_scheduler_schedule_recurring_actions` that plugins should start using in order to schedule recurring actions.  This specific implementation schedules a new recurring action just to run this recurring hook - the idea being that one implementation with controlled best practices on when it checks to see if it is scheduled is better than each plugin doing it itself.

# New Proposed Integration

```php
function eg_schedule_midnight_log() {
	if ( false === as_has_scheduled_action( 'eg_midnight_log' ) ) {
		as_schedule_recurring_action( strtotime( 'tomorrow' ), DAY_IN_SECONDS, 'eg_midnight_log', array(), '', true );
	}
}

/**
 * If you plugin includes ActionScheduler and therefore knows that the version supports the
 * `action_scheduler_schedule_recurring_actions` hook, you can simply use
 */
add_action( 'action_scheduler_schedule_recurring_actions', 'eg_schedule_midnight_log' );

/**
 * If your plugin does not package ActionScheduler and is unsure if the `action_scheduler_schedule_recurring_actions` hook
 * is supported, use the following:
 */
if ( function_exists( 'as_supports' ) && as_supports( 'schedule_recurring_actions_hook' ) ) {
	add_action( 'action_scheduler_schedule_recurring_actions', 'eg_schedule_midnight_log' );
} else {
	add_action( 'action_scheduler_init', 'eg_schedule_midnight_log' );
}
```

# Testing Instructions

1. Go to WP-Admin -> Tools -> Scheduled Actions
2. You should be able to find an action scheduled with the hook of 'action_scheduler_run_recurring_actions_schedule_hook'
3. Add a snippet similar to the following somewhere on the site:
 ```php
	add_action( 'action_scheduler_schedule_recurring_actions', function () {
		if ( ! as_has_scheduled_action( 'my_random_callback_action' ) ) {
			as_schedule_recurring_action( time(), 5 * MINUTE_IN_SECONDS, 'my_random_callback_action' );
		}
	} );
} );

add_action( 'my_random_callback_action', function () {
	error_log( 'IT WORKED!!!!!!!' );
} );
 ```
 4. Run the `action_scheduler_run_recurring_actions_schedule_hook` action by hovering over it and clicking `Run` (otherwise you'd have to wait up to 24 hours).
  Screenshot 2025-04-09 at 1 04 36 PM
5. You should immediately see the `my_random_callback_action` Action scheduled as a new recurring action and checking your error log should show the `IT WORKED!!!!!!!` text.

### DIFF
--- a/classes/ActionScheduler_RecurringActionScheduler.php
+++ b/classes/ActionScheduler_RecurringActionScheduler.php
@@ -1,0 +1,44 @@
+<?php
+
+/**
+ * Class ActionScheduler_RecurringActionScheduler
+ *
+ * This class ensures that the `action_scheduler_schedule_recurring_actions` hook is triggered on a daily interval. This
+ * simplifies the process for other plugins to register their recurring actions without requiring each plugin to query
+ * or schedule actions independently on every request.
+ */
+class ActionScheduler_RecurringActionScheduler {
+
+	/**
+	 * Initialize the instance.  Should only be run on a single instance per request.
+	 *
+	 * @return void
+	 */
+	public function init(): void {
+		if ( is_admin() && ( ! defined( 'DOING_AJAX' ) || ! DOING_AJAX ) ) {
+			add_action( 'action_scheduler_init', [ __CLASS__, 'schedule_recurring_scheduler_hook' ] );
+		}
+	}
+
+	/**
+	 * Schedule the recurring `action_scheduler_schedule_recurring_actions` action if not already scheduled.
+	 *
+	 * @return void
+	 */
+	public function schedule_recurring_scheduler_hook(): void {
+		if ( false === wp_cache_get( 'as_is_recurring_scheduler_scheduled' ) ) {
+			if ( ! as_has_scheduled_action( 'action_scheduler_schedule_recurring_actions' ) ) {
+				as_schedule_recurring_action(
+					time(),
+					DAY_IN_SECONDS, // Hourly interval
+					'action_scheduler_schedule_recurring_actions',
+					[],
+					'ActionScheduler',
+					true,
+					20
+				);
+			}
+			wp_cache_set( 'as_is_recurring_scheduler_scheduled', true, HOUR_IN_SECONDS );
+		}
+	}
+}
--- a/classes/ActionScheduler_RecurringActionScheduler.php
+++ b/classes/ActionScheduler_RecurringActionScheduler.php
@@ -10,13 +10,22 @@
 class ActionScheduler_RecurringActionScheduler {
 
 	/**
+	 * @var string The hook of the scheduled recurring action that is run to trigger the
+	 *      `action_scheduler_schedule_recurring_actions` hook that plugins should use.  We can't directly have the
+	 *      scheduled action hook be the hook plugins should use because the actions will show as failed if no plugin
+	 *      was actively hooked into it.
+	 */
+	private const RUN_SCHEDULED_RECURRING_ACTIONS_HOOK = 'action_scheduler_run_recurring_actions_schedule_hook';
+
+	/**
 	 * Initialize the instance.  Should only be run on a single instance per request.
 	 *
 	 * @return void
 	 */
 	public function init(): void {
+		add_action( self::RUN_SCHEDULED_RECURRING_ACTIONS_HOOK, array( $this, 'run_recurring_scheduler_hook' ) );
 		if ( is_admin() && ( ! defined( 'DOING_AJAX' ) || ! DOING_AJAX ) ) {
-			add_action( 'action_scheduler_init', [ __CLASS__, 'schedule_recurring_scheduler_hook' ] );
+			add_action( 'action_scheduler_init', array( $this, 'schedule_recurring_scheduler_hook' ) );
 		}
 	}
 
@@ -27,11 +36,11 @@ class ActionScheduler_RecurringActionScheduler {
 	 */
 	public function schedule_recurring_scheduler_hook(): void {
 		if ( false === wp_cache_get( 'as_is_recurring_scheduler_scheduled' ) ) {
-			if ( ! as_has_scheduled_action( 'action_scheduler_schedule_recurring_actions' ) ) {
+			if ( ! as_has_scheduled_action( self::RUN_SCHEDULED_RECURRING_ACTIONS_HOOK ) ) {
 				as_schedule_recurring_action(
 					time(),
 					DAY_IN_SECONDS, // Hourly interval
-					'action_scheduler_schedule_recurring_actions',
+					self::RUN_SCHEDULED_RECURRING_ACTIONS_HOOK,
 					[],
 					'ActionScheduler',
 					true,
@@ -40,5 +49,14 @@ class ActionScheduler_RecurringActionScheduler {
 			}
 			wp_cache_set( 'as_is_recurring_scheduler_scheduled', true, HOUR_IN_SECONDS );
 		}
+	}
+
+	/**
+	 * Trigger the hook to allow other plugins to schedule their recurring actions.
+	 *
+	 * @return void
+	 */
+	public function run_recurring_scheduler_hook(): void {
+		do_action( 'action_scheduler_schedule_recurring_actions' );
 	}
 }
--- a/classes/ActionScheduler_RecurringActionScheduler.php
+++ b/classes/ActionScheduler_RecurringActionScheduler.php
@@ -3,7 +3,7 @@
 /**
  * Class ActionScheduler_RecurringActionScheduler
  *
- * This class ensures that the `action_scheduler_schedule_recurring_actions` hook is triggered on a daily interval. This
+ * This class ensures that the `action_scheduler_ensure_recurring_actions` hook is triggered on a daily interval. This
  * simplifies the process for other plugins to register their recurring actions without requiring each plugin to query
  * or schedule actions independently on every request.
  */
@@ -11,7 +11,7 @@ class ActionScheduler_RecurringActionScheduler {
 
 	/**
 	 * @var string The hook of the scheduled recurring action that is run to trigger the
-	 *      `action_scheduler_schedule_recurring_actions` hook that plugins should use.  We can't directly have the
+	 *      `action_scheduler_ensure_recurring_actions` hook that plugins should use.  We can't directly have the
 	 *      scheduled action hook be the hook plugins should use because the actions will show as failed if no plugin
 	 *      was actively hooked into it.
 	 */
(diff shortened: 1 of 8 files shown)


classes/ActionScheduler_RecurringActionScheduler.php
```diff
@@ -0,0 +1,62 @@
...
+		}
+	}
+
+	/**
+	 * Trigger the hook to allow other plugins to schedule their recurring actions.
+	 *
+	 * @return void
+	 */
+	public function run_recurring_scheduler_hook(): void {
+		do_action( 'action_scheduler_schedule_recurring_actions' );
```
jorgeatorres: It'd be nice to have a docblock here that documents this hook well.

classes/ActionScheduler_RecurringActionScheduler.php
```diff
@@ -0,0 +1,62 @@
...
+	 * Schedule the recurring `action_scheduler_schedule_recurring_actions` action if not already scheduled.
+	 *
+	 * @return void
+	 */
+	public function schedule_recurring_scheduler_hook(): void {
+		if ( false === wp_cache_get( 'as_is_recurring_scheduler_scheduled' ) ) {
+			if ( ! as_has_scheduled_action( self::RUN_SCHEDULED_RECURRING_ACTIONS_HOOK ) ) {
+				as_schedule_recurring_action(
+					time(),
+					DAY_IN_SECONDS, // Hourly interval
```
jorgeatorres: Due to the expiration set below for the cache we will _check_ that the recurring action exists at least hourly, but the action itself is not recurring hourly but daily.

jorgeatorres (commented):
Hey @prettyboymp 👋 

I'm not formally assigned as reviewer for this PR but wanted to leave some notes. I like the idea of centralizing this type of checks and I agree that this is much better than every plugin doing this on its own, potentially running tons of queries during `init`.

That said, the solution here seems more suited for recurring actions for which the first time the action runs isn't too important. If you used the new hook, you'd have to wait at least 24 hrs for the action to be even scheduled for the first time (and then some more for it to run), so the `$timestamp` arg for `as_schedule_recurring_action()` becomes pretty much irrelevant.

This might not be a problem most of the time, and of course, A-S doesn't make guarantees that any action will run at a particular moment, but it's a difference that I think we should properly document as the existence of an 'official' hook might give the impression that things remain the same, but you just use the new hook (vs init).

Assuming we have a site where the queue is processed frequently enough, this change means that a recurring action that would've been executed (for the first time) in the "next 5 mins" now will only be scheduled in the next 24 hrs, and possibly run for the first time before the 25th hour.

***

Slight tangent: I wonder if for such actions that 3rd parties need to schedule and for which the given frequency is more important than the _first_ time they run, we should have a different API function (that doesn't receive a `$timestamp` arg, even if internally it uses current time). This would also give us the ability to control things better but maybe would be less confusing. Just a thought.

prettyboymp:
@jorgeatorres , thank you for the feedback.

> That said, the solution here seems more suited for recurring actions for which the first time the action runs isn't too important. If you used the new hook, you'd have to wait at least 24 hrs for the action to be even scheduled for the first time (and then some more for it to run), so the $timestamp arg for as_schedule_recurring_action() becomes pretty much irrelevant.

Ideally, if an extension has a recurring action that needs to be run, it should first schedule it during an activation hook or a custom update check if adding a new action.  However, there are instances does not get rescheduled - because of failures, DB issues, etc.  Because of these cases, extensions have gotten into the pattern of adding the `as_has_scheduled_action` check on every page load.

I went back and forth on how often this should run.  Based on the use case, I think 24 hrs is proficient, especially if a plugin takes care of having their own activation hook.  One option could be to run the hook automatically on the `activated_plugin` action to cover any extension that don't have their own.

jorgeatorres:
Hey @prettyboymp!

> Ideally, if an extension has a recurring action that needs to be run, it should first schedule it during an activation hook or a custom update check if adding a new action. However, there are instances does not get rescheduled - because of failures, DB issues, etc. Because of these cases, extensions have gotten into the pattern of adding the `as_has_scheduled_action` check on every page load.

Yes, I understand what the goal is here and I 100% agree with you. That's why my feedback was about the hook itself, because the current name might seem to indicate that this hook is now the correct way to schedule the recurring actions, while in reality it's meant to be used for checking and ensuring that those actions remain scheduled.

Hopefully we can make that clearer with a docblock section or maybe some tweaking to the hook name.

Also, it'd be good to add this to our public docs so that the current (less than ideal) pattern of using `init` doesn't remain the only "official" way to schedule the jobs. This bit of documentation [lives in `docs/usage.md`](https://github.com/woocommerce/action-scheduler/blob/5590e3ae64bc876ce132057e618681470692d373/docs/usage.md?plain=1#L99-L129), which maybe we can update in this PR to reflect your suggestions:
- That ideally plugins should schedule those actions during an activation hook or while installing/updating or when a change in an option or something requires it.
- That they shouldn't check whether the jobs are scheduled on every page load, and (since AS 3.9.3) can rely on the new hook we're adding here.

What do you think?

prettyboymp:
@jorgeatorres, 100% agree.  I'll work on the updates to documentation and reevaluate the hook name.

prettyboymp:
@jorgeatorres,

I’ve updated the hook name to `action_scheduler_ensure_recurring_actions` to clarify that its purpose is to help plugins ensure their recurring actions remain scheduled, rather than serving as the primary mechanism for scheduling those actions.

The documentation has also been revised to better explain this distinction, emphasizing the use of activation hooks or upgrade routines for initially scheduling recurring actions.

**Other Note:** I considered triggering the `action_scheduler_ensure_recurring_actions` hook on `plugin_activated` as well, to help plugins that rely solely on this hook and do not implement their own activation routine. However, I decided against this approach because the `plugin_activated` hook does not fire in all scenarios (e.g., during network activation). Introducing a fallback that is not universally reliable could mask issues during typical testing, so I opted to leave this responsibility with the plugin developer and to lean more heavily in the best practice to use activation/update routines for initial scheduling.

jorgeatorres (approved):
LGTM. Thank you @prettyboymp!

docs/usage.md
```diff
@@ -96,36 +96,93 @@ Action Scheduler will later initialize itself on `'init'` with priority `1`.  Ac
...
- * Schedule an action with the hook 'eg_midnight_log' to run at midnight each day
- * so that our callback is run then.
+ * Ensure recurring action remains scheduled.
  */
-function eg_schedule_midnight_log() {
-	if ( false === as_has_scheduled_action( 'eg_midnight_log' ) ) {
-		as_schedule_recurring_action( strtotime( 'tomorrow' ), DAY_IN_SECONDS, 'eg_midnight_log', array(), '', true );
+add_action( 'action_scheduler_ensure_recurring_actions', function() {
+    if ( ! as_has_scheduled_action( 'my_plugin_recurring_action' ) ) {
+        as_schedule_recurring_action(
```
barryhughes, line 135: *(I realize this PR was merged already, I was just catching up on the latest changes ... nice work on this, btw!)*

This leads to a race condition, because a second process may create the same recurring action right after our check to see if it exists.

- What do you think of setting `unique` to true?
- Going off on a bit of a tangent, what do you think about adding support for `as_supports( 'unique-actions' )` (on the basis that, on those rare occasions where a plugin assumes unique actions available, but they are not, a flood of actions may be created)?
    - *Edit* → I suppose that's still awkward, because in such a scenario `as_supports()` may also be unavailable.

senadir:
Which version of Woo did this land on?

prettyboymp:
> Which version of Woo did this land on?

It isn't yet available in WooCommerce core.  We still need to cut a release of the latest A-S with these changes and upgrade Woo core. Hopefully we can do that before 10.1.

classes/ActionScheduler_RecurringActionScheduler.php
```diff
@@ -0,0 +1,81 @@
...
+					time(),
+					DAY_IN_SECONDS,
+					self::RUN_SCHEDULED_RECURRING_ACTIONS_HOOK,
+					[],
+					'ActionScheduler',
+					true,
+					20
+				);
+			}
+			wp_cache_set( 'as_is_ensure_recurring_actions_scheduled', true, HOUR_IN_SECONDS );
```
shivapoudel, line 50: @jorgeatorres Expiry was set for the cache group. Please fix this.
```
wp_cache_set( 'as_is_ensure_recurring_actions_scheduled', true, 'action-scheduler', HOUR_IN_SECONDS );
```

classes/ActionScheduler_RecurringActionScheduler.php
```diff
@@ -0,0 +1,81 @@
...
+		}
+	}
+
+	/**
+	 * Schedule the recurring `action_scheduler_ensure_recurring_actions` action if not already scheduled.
+	 *
+	 * @return void
+	 */
+	public function schedule_recurring_scheduler_hook(): void {
+		if ( false === wp_cache_get( 'as_is_ensure_recurring_actions_scheduled' ) ) {
```
shivapoudel, line 38: We can also define a group here.
```
wp_cache_get( 'as_is_ensure_recurring_actions_scheduled', 'action-scheduler' )
```

classes/ActionScheduler_RecurringActionScheduler.php
jorgeatorres: Thanks for the ping @shivapoudel and for noticing. I'm working [on a PR](https://github.com/woocommerce/action-scheduler/pull/1302) to address this.

Issue #1023, opened 2024-01-09
Hybrid store does not support unique actions


rrennick:
The action scheduler [migration gets re-enabled](https://github.com/woocommerce/action-scheduler/blob/trunk/classes/ActionScheduler_DataController.php#L169) every time a plugin is deactivated. In turn this switches AS to the hybrid data store which does not support unique actions.

I have a dev install that only runs AS actions through CLI that added ~1K copies of a unique action in a couple hours.

1- It should be safe to remove the code that re-enables the hybrid store as the DBStore has been default for almost 4 years.
2- The Hybrid store should support unique actions primarily for development/staging installs

barryhughes:
> It should be safe to remove the code that re-enables the hybrid store as the DBStore has been default for almost 4 years.

For whenever we take this up: let's be careful on this point. Though it was a couple of years back, we did see a relatively popular third party plugin still shipping 2.x (and there could definitely be unmaintained custom plugins in the same boat).

Definitely great if we can do it safely.

barryhughes:
I think #1039 is a satisfactory solution for this problem. The post store is generally not used, and unless we decide to retain it as a backup for other problems (like missing tables [[1]](https://github.com/woocommerce/action-scheduler/issues/1059) [[2]](https://github.com/woocommerce/action-scheduler/issues/973) [[3]](https://github.com/woocommerce/action-scheduler/issues/1032)) then I don't think it's worth investing too much more time into the provision of full-fledged unique actions.

PR #1265, opened 2025-04-11
Make `$unique` available to all pre-creation/short-circuit hooks.

### LINKED ISSUE #1263: `pre_as_schedule_single_action` hook does not supply `$unique`
When actions are created, Action Scheduler triggers pre-creation actions that follow the standard WordPress 'short-circuit' pattern, allowing other plugins to intervene and takeover action creation, etc. [`pre_as_enqueue_async_action`](https://github.com/woocommerce/action-scheduler/blob/3.9.2/functions.php#L40C25-L40C52) is a pretty good example.

For all of these (except the `async` variant mentioned at the start), though, we seem to have forgotten to supply the `$unique` param:

- [`pre_as_schedule_single_action`](https://github.com/woocommerce/action-scheduler/blob/3.9.2/functions.php#L90)
- [`pre_as_schedule_recurring_action`](https://github.com/woocommerce/action-scheduler/blob/3.9.2/functions.php#L163)
- [`pre_as_schedule_cron_action`](https://github.com/woocommerce/action-scheduler/blob/3.9.2/functions.php#L229)

This is a little limiting, and makes it hard for third party code to safely use these hooks.

⛑ The 'safety' factor comes from the fact that callbacks can't know if an action is intended to be unique or not ... so if the job of the callback is to modify the scheduled action in some way before persisting it, and the action happens to be unique in nature, we'd potentially end up creating unwanted duplicates.

barryhughes:
Makes `$unique` available to consumers of the following filter hooks:

- `pre_as_schedule_single_action`
- `pre_as_schedule_recurring_action`
- `pre_as_schedule_cron_action`

#### Why is this needed?

Unique actions *('there should be only one')* have been a feature of Action Scheduler since the 3.6.0 release. 

However, currently, callbacks using any of the three hooks above will not be able to tell if an action should be unique or not ... in some cases this may lead to them (re-)creating the actions as if they were non-unique. This is quite risky—please refer back to [this historic issue](https://github.com/woocommerce/action-scheduler/issues/1023) (which was [partly solved already](https://github.com/woocommerce/action-scheduler/pull/1039), though we didn't think about this specific dimension at the time).

#### Testing

Code review is probably sufficient here. We're supplying an additional param to the hooks, but are not changing the order of the params—so existing consumers should be unaffected. 

That said, you can simulate the problem and see the change by running through the following steps. Start by checking out this branch and then add the following code (I'm using PHP 8 syntax for readability, please adapt as needed if you are using an older version) to a mu-plugin or other suitable location. Note that, initially, we only accept 7 params, so the callback is unaware of whether or not an action should be unique. This lets us simulate the problem:

```php
add_filter(
	'pre_as_schedule_recurring_action',
	function ( $short_circuit, $timestamp, $interval_in_seconds, $hook, $args, $group, $priority ) {
		static $lock = false;

		// Prevent infinite loops of doom 🧙‍♂️
		if ( $lock ) {
			return $short_circuit;
		}

		$lock   = true;
		$result = as_schedule_recurring_action(
			timestamp:           $timestamp,
			interval_in_seconds: $interval_in_seconds,
			hook:                $hook,
			args:                $args,
			group:               $group,
			priority:            $priority,
		);
		$lock   = false;

		return $result;
	},
	10,
	7
);
```

Try creating a *unique* recurring action. Via WP CLI:

```
wp eval "as_schedule_recurring_action( time() + 600, HOUR_IN_SECONDS, 'test_action', [], '', true );"
```

Do this three times, then run:

```
wp action-scheduler action list --hook=test_action
```

You should see that we have three of these test actions. This is wrong/bad/risky, because the action is intended to be unique (there should only ever be one pending instance):

```
% wp action-scheduler action list --hook=test_action
                                                
+--------+-------------+---------+-------+-----------+---------------------------+
| id     | hook        | status  | group | recurring | scheduled_date            |
+--------+-------------+---------+-------+-----------+---------------------------+
| 665969 | test_action | pending |       | yes       | 2025-04-11 18:25:16 +0000 |
| 665972 | test_action | pending |       | yes       | 2025-04-11 18:30:59 +0000 |
| 665974 | test_action | pending |       | yes       | 2025-04-11 18:31:32 +0000 |
+--------+-------------+---------+-------+-----------+---------------------------+
```

Now modify the snippet to the following (this demonstrates the fix):

```php
add_filter(
	'pre_as_schedule_recurring_action',
	function ( $short_circuit, $timestamp, $interval_in_seconds, $hook, $args, $group, $priority, $unique ) {
		static $lock = false;

		// Prevent infinite loops of doom 🧙‍♂️
		if ( $lock ) {
			return $short_circuit;
		}

		$lock   = true;
		$result = as_schedule_recurring_action(
			timestamp:           $timestamp,
			interval_in_seconds: $interval_in_seconds,
			hook:                $hook,
			args:                $args,
			group:               $group,
			priority:            $priority,
			unique:              $unique,
		);
		$lock   = false;

		return $result;
	},
	10,
	8
);
```

Repeat the test. No further actions should be created (uniqueness is now respected). If you wish, you can now switch to a different hook name and verify that the creation of *new* unique actions is not impeeded:

```
% wp eval "as_schedule_recurring_action( time() + 600, HOUR_IN_SECONDS, 'test_action2', [], '', true );"
% wp eval "as_schedule_recurring_action( time() + 600, HOUR_IN_SECONDS, 'test_action2', [], '', true );"
% wp action-scheduler action list --hook=test_action2     
                                              
+--------+--------------+---------+-------+-----------+---------------------------+
| id     | hook         | status  | group | recurring | scheduled_date            |
+--------+--------------+---------+-------+-----------+---------------------------+
| 665977 | test_action2 | pending |       | yes       | 2025-04-11 18:43:54 +0000 |
+--------+--------------+---------+-------+-----------+---------------------------+
```

Closes #1263

### Changelog

> Dev - Makes the `$unique` param available via the `pre_as_schedule_*_action` hooks.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -86,8 +86,9 @@ function as_schedule_single_action( $timestamp, $hook, $args = array(), $group =
 	 * @param array    $args       Action arguments.
 	 * @param string   $group      Action group.
 	 * @param int      $priorities Action priority.
+	 * @param bool     $unique     Unique action.
 	 */
-	$pre = apply_filters( 'pre_as_schedule_single_action', null, $timestamp, $hook, $args, $group, $priority );
+	$pre = apply_filters( 'pre_as_schedule_single_action', null, $timestamp, $hook, $args, $group, $priority, $unique );
 	if ( null !== $pre ) {
 		return is_int( $pre ) ? $pre : 0;
 	}
@@ -159,8 +160,9 @@ function as_schedule_recurring_action( $timestamp, $interval_in_seconds, $hook, 
 	 * @param array    $args                Action arguments.
 	 * @param string   $group               Action group.
 	 * @param int      $priority            Action priority.
+	 * @param bool     $unique              Unique action.
 	 */
-	$pre = apply_filters( 'pre_as_schedule_recurring_action', null, $timestamp, $interval_in_seconds, $hook, $args, $group, $priority );
+	$pre = apply_filters( 'pre_as_schedule_recurring_action', null, $timestamp, $interval_in_seconds, $hook, $args, $group, $priority, $unique );
 	if ( null !== $pre ) {
 		return is_int( $pre ) ? $pre : 0;
 	}
@@ -225,8 +227,9 @@ function as_schedule_cron_action( $timestamp, $schedule, $hook, $args = array(),
 	 * @param array    $args       Action arguments.
 	 * @param string   $group      Action group.
 	 * @param int      $priority   Action priority.
+	 * @param bool     $unique     Unique action.
 	 */
-	$pre = apply_filters( 'pre_as_schedule_cron_action', null, $timestamp, $schedule, $hook, $args, $group, $priority );
+	$pre = apply_filters( 'pre_as_schedule_cron_action', null, $timestamp, $schedule, $hook, $args, $group, $priority, $unique );
 	if ( null !== $pre ) {
 		return is_int( $pre ) ? $pre : 0;
 	}


barryhughes:
💡 My testing instructions might seem a little contrived: they intercept the creation of the action, just to create the very same action. So you have to use some imagination and imagine this logic exists to A) store the action—or a copy of it—somewhere else or B) it's doing this to modify the priority or cadence of specific actions (which, in fact, is the reason I hit upon this problem).

prettyboymp (approved):
This all looks good to me.

jorgeatorres (approved):
LGTM.

PR #1274, opened 2025-06-12
Update minimum supported php and phpunit versions

joshuatf:
Bumps the minimum supported PHP version to 7.2 and the phpunit version to ^8.5.

[Phpunit 7 has reached its end of life](https://phpunit.de/supported-versions.html) for some time and we need phpunit 8 (requires PHP >=7.2) in order to more easily test across all PHP versions.

This PR also updates test to fix some typing issues that occur due to the phpunit bump.

## Testing

1. Confirm that the bump to PHP 7.2 is acceptable.
2. Confirm that all phpunit tests pass.

### DIFF
--- a/.github/workflows/pr-unit-tests.yml
+++ b/.github/workflows/pr-unit-tests.yml
@@ -10,17 +10,17 @@ jobs:
       fail-fast: false
       matrix:
         # We test against the earliest and latest PHP versions for each major supported version.
-        php: [ '7.1', '7.4', '8.0', '8.3' ]
+        php: [ '7.2', '7.4', '8.0', '8.3' ]
         wp: [ '6.5', '6.6', 'latest', 'nightly' ]
         multisite: [ '0', '1' ]
         db: [ 'mysql:5.6', 'mysql:8.1', 'mariadb:10.4', 'mariadb:10.6']
         exclude:
           # WordPress 6.6+ requires PHP 7.2+
-          - php: 7.1
+          - php: 7.2
             wp: 6.6
-          - php: 7.1
+          - php: 7.2
             wp: latest
-          - php: 7.1
+          - php: 7.2
             wp: nightly
     services:
       database:
--- a/.github/workflows/pr-unit-tests.yml
+++ b/.github/workflows/pr-unit-tests.yml
@@ -61,27 +61,6 @@ jobs:
       - name: Install dependencies
         run: composer install --prefer-dist
 
-      - name: Setup PHPUnit 
-        run: |
-          # PHPUnit 7.5 when using PHP 7.1 - 7.4.
-          if [ "$(php -r "echo version_compare( PHP_VERSION, '8.0', '<' );")" ]; then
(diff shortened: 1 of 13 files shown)


.github/workflows/pr-unit-tests.yml
```diff
@@ -10,17 +10,17 @@ jobs:
       fail-fast: false
       matrix:
         # We test against the earliest and latest PHP versions for each major supported version.
-        php: [ '7.1', '7.4', '8.0', '8.3' ]
+        php: [ '7.2', '7.4', '8.0', '8.3' ]
```
barryhughes, line 13: Not necessarily something to address in this PR 🙌🏼 ... but at some point we probably should replace `'8.3'` with `'8.4'`.

.github/workflows/pr-unit-tests.yml
```diff
@@ -61,27 +61,6 @@ jobs:
...
-            rm ./vendor/bin/phpunit
-            ln -sf /tmp/phpunit-fork/phpunit-add-compatibility-with-php8-to-phpunit-7/phpunit ./vendor/bin/phpunit
-          fi
-          
-          if [ $OVERWRITE ]; then
-            rm ./vendor/bin/phpunit
-            chmod +x /tmp/phpunit
-            mv /tmp/phpunit ./vendor/bin/phpunit
-          fi
-
```
barryhughes, line 84: Nice to see this removed!

barryhughes (approved):
Thanks! Looks good to me.

✍🏼 We'll want to be doubly sure we include a changelog entry for this, once we ship the next update. Could it also be worth posting a dev advisory? I don't have a strong feeling on that, but thought I'd pose the question (although I guess we could consider it to be a 'soft bump', because functionally nothing has changed just yet, we wouldn't want this to go unnoticed by library consumers).

:hurtrealbad: I notice the following warnings popping up ... not something we need to address here in this PR, but worth being cognizant of:

> The optional $delta parameter of assertEquals() is deprecated and will be removed in PHPUnit 9. Refactor your test to use assertEqualsWithDelta() instead.

joshuatf:
>We'll want to be doubly sure we include a changelog entry for this

@barryhughes Showing my naivety, but how does the changelog work in this repo?  Doesn't look like we use changelogger; is there an unreleased section usually?

barryhughes:
> Showing my naivety, but how does the changelog work in this repo?

No, it's a great question, and that's really why I flagged it—as there unfortunately hasn't been as structured an approach (or the same tooling) as for Woo Core. Mostly it's up to the release lead to build/verify/correct the changelog, based on PR descriptions/merge commit messages. That obviously isn't great, and has led to sub-par changelogs in the past.

I think aligning it with Woo Core's approach, or else updating the changelog on a PR-by-PR basis (and resolving conflicts as needed) may be better, but will leave that to you all to figure out.

... Also (excuse the use of an internal link) Pf59Ax-ZD-p2 may be of interest.

joshuatf:
Thanks for the context, @barryhughes!  Added an issue to implement the changelogger [here](https://github.com/woocommerce/action-scheduler/issues/1275).

PR #1283, opened 2025-08-12
Include database error in the exception message.

dd32:
In #966 the "Database error." exception was improved to include the actual DB error, however, it only updated one location where this exception is used.

This sync's the code to the other location.

100% untested.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -2,6 +2,7 @@
 
 = 4.1.0 - xxxx-xx-xx =
 * Use `wp_admin_notice()` to render admin notices, instead of custom HTML.
+* Dev - Updates the `wpPostStore` to capture any database errors that might occur while claiming actions.
 
 = 4.0.0 - 2026-06-16 =
 * Breaking change: action args are taken into account when scheduling unique actions.
--- a/classes/data-stores/ActionScheduler_wpPostStore.php
+++ b/classes/data-stores/ActionScheduler_wpPostStore.php
@@ -697,10 +697,12 @@ class ActionScheduler_wpPostStore extends ActionScheduler_Store {
 				? _x( 'unknown', 'database error', 'action-scheduler' )
 				: $wpdb->last_error;
 			throw new RuntimeException(
-				sprintf(
-					/* translators: %s database error. */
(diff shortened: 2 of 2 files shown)


barryhughes (commented):
Nice, low overhead and probably worth including for improved consistency with the `DBStore`.

Note that the `wpPostStore` mostly sits unused on modern installations (the current default is `DBStore`). A temporary exception to this crops up when a plugin is activated/deactivated and, though generally only for a very short period of time, it will then be invoked via the `HybridStore`.

So, to test, I forced the use of `wpPostStore` with:

```php
add_filter( 'action_scheduler_store_class', fn () => ActionScheduler_wpPostStore::class, 1000 );
```

I then deliberately introduced an error to the claim process using:

```php
add_filter( 'query', function ( $query ) {
	return class_exists( ActionScheduler_wpPostStore::class ) && str_contains( $query, ActionScheduler_wpPostStore::POST_TYPE )
		? $query . ' __SCUPPERED__'
		: $query;
} );
```

And triggered a queue run via:

```sh
wp action-scheduler run
```

The resulting database error was dutifully reported in the command output:

```
Error: There was an error running the action scheduler: Unable to claim actions. Database error: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '__SCUPPERED__' at line 1.
```

As probably true of the original work for the `DBStore`, this may anyway be caught and logged in various environments because `wpdb` makes its own `error_log()` call, but again it improves consistency and helps ensure the problem is surfaced *(perhaps* at the cost of adding noise, in some environments).

barryhughes (approved):

barryhughes:
Current crop of test failures [(example)](https://github.com/woocommerce/action-scheduler/actions/runs/28668396381/job/85036007219?pr=1283) look unrelated to the change itself:

```
⯆ Init DB and WP

  ...

  svn: E170013: Unable to connect to a repository at URL 'https://core.svn.wordpress.org/trunk'
  svn: E175002: Unexpected HTTP status 504 'Gateway Time-out' on '/trunk'
  Error: Process completed with exit code 1.
```

Will re-run.

PR #1285, opened 2025-08-27
Fix number formating for action numbers in list view

### LINKED ISSUE #1279: Missing formatting for action count in list view
While working on a client website I noticed that the action count number of the list view under "Tools" is not formatted. This makes it very hard to read if the number is big. 4933472 or 4933472529 is a huge difference but not very easy to see. 
WordPress is formatting this number on other table views. Also the number inside the pagination has a nice format.
I think the number should be formatted too.

Image

I created this patch to fix it:
[fix_formating_of_action_count.patch](https://github.com/user-attachments/files/20969777/fix_formating_of_action_count.patch)

Drivingralle:
Fixes #1279 
Changes to format the count nicely


prettyboymp (approved):
Changes look great!  Thank you for submitting this fix.

PR #1288, opened 2025-09-17
Import missing WP_CLI class to prevent fatal error

szepeviktor:

1. Run `wp action-scheduler action list` and take note of the `id` of any action.
2. Run `wp action-scheduler action logs <id>`. No fatal error should occur and the logs for the action should appear.
3. (Optional) On `trunk`, run `wp action-scheduler action logs <id>` and you get a fatal error.

https://github.com/woocommerce/action-scheduler/blob/628fd3ccf87649ba470240afa9abc4adb8bce7e0/classes/WP_CLI/Action_Command.php#L284-L285

### DIFF
--- a/classes/WP_CLI/Action_Command.php
+++ b/classes/WP_CLI/Action_Command.php
@@ -2,6 +2,8 @@
 
 namespace Action_Scheduler\WP_CLI;
 
+use WP_CLI;
+
 /**
  * Action command for Action Scheduler.
  */


classes/WP_CLI/Action_Command.php
```diff
@@ -2,6 +2,8 @@
 
 namespace Action_Scheduler\WP_CLI;
 
+use WP_CLI;
+
```
jorgeatorres, line 6: Shouldn't be enough to change the `WP_CLI::runcommand()` line to `\WP_CLI::runcommand()` instead of importing `WP_CLI` here?

jorgeatorres (changes requested):
Hey @szepeviktor!

Thanks for your contribution. I left some minor comment, but if you don't mind, could you please update the PR description with details on why the change is needed and testing instructions.

I understand it's to prevent a fatal error, but it'd be good to properly document it for future reference here.

Thanks again!

classes/WP_CLI/Action_Command.php
szepeviktor: It depends on whether importing classes is your thing.

jorgeatorres: Right. In this case, given nothing in `Action_Command` requires `\WP_CLI` I think that would be preferable, but I agree it's optional.

jorgeatorres:
@szepeviktor: Thanks for responding. Let us know if you have the time to add a description to the PR with the changes and also some testing instructions and we'll be more than happy to review again. Thanks again!

szepeviktor:
@jorgeatorres Done.

jorgeatorres (approved):
LGTM. Thanks @szepeviktor!

(I've expanded the testing instructions for clarity)

szepeviktor:
Glad to contribute.

PR #1289, opened 2025-09-25
Option lock stuck locked with ''

superdav42:
Somehow (maybe old AS version?) the option value can get set to `''`. In this case the code will attempt to insert the option but it will give a duplicate key error. So the lock can never be set until someone manually deletes this option. 

With this cade change the option will be updated if it's `''` an update will be used instead of an insert.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -2,6 +2,7 @@
 
 = 4.1.0 - xxxx-xx-xx =
 * Use `wp_admin_notice()` to render admin notices, instead of custom HTML.
+* Fix an oversight in the lock implementation (used to rate-limit async request runners) that could result in a database error in unusual cases.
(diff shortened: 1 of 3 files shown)


barryhughes:
Sorry for the delay here, @superdav42 ... this looks like a promising fix. 

Possibly, this was the result of an earlier version of the lock implementation [(authored by me 😢)](https://github.com/woocommerce/action-scheduler/commit/636e298#diff-c5062d1cfe68a6b17562f5bcc5a0e9ca63fce39c4077b3f1215ad46f808256a5R102-R124) which created the new lock, initially with an empty string, and *then* updated it with an expiration (of course, that creates the potential for the lock to be left empty, if a server crash or similar occurs in between those two operations).

Though, there may also be some other path that would lead to the problem, and so it's worth guarding against it 👍🏼

barryhughes (approved):

PR #1297, opened 2025-11-15
Correctly implement wp_cache_set in ActionScheduler_RecurringActionScheduler

### LINKED ISSUE #1296: AS 3.9.3 incorrectly implemented wp_cache_set in ActionScheduler_RecurringActionScheduler
Check this code from the latest 3.9.3:

```php
wp_cache_set( 'as_is_ensure_recurring_actions_scheduled', true, HOUR_IN_SECONDS );
```

It sets the value `3600` as the cache group name.

`wp_cache_set()` expects the expiration to be the 4th parameter.

For your own jobs, you use `ActionScheduler` as the group name. So maybe here it should also be:

```php
wp_cache_set( 'as_is_ensure_recurring_actions_scheduled', true, 'ActionScheduler', HOUR_IN_SECONDS );
```

Without this change (see #1297) the cache will never expire.

slaFFik:
Use the correct unique cache group `'ActionScheduler'` in the single `wp_cache_set()` function call.

Fix #1296

### DIFF
--- a/classes/ActionScheduler_RecurringActionScheduler.php
+++ b/classes/ActionScheduler_RecurringActionScheduler.php
@@ -35,7 +35,7 @@ class ActionScheduler_RecurringActionScheduler {
 	 * @return void
 	 */
 	public function schedule_recurring_scheduler_hook(): void {
-		if ( false === wp_cache_get( 'as_is_ensure_recurring_actions_scheduled' ) ) {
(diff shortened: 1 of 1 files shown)


dd32:
The `wp_cache_get( 'as_is_ensure_recurring_actions_scheduled' )` a few lines higher will need to be adjusted to match.

I suspect this should probably be a `get_transient()`/`set_transient()` call instead. _(as transients are used elsewhere in action-scheduler, and matches the incorreect wp_cache_set() syntax used here)_

PR #1300, opened 2025-12-01
Fix partial args matching

### LINKED ISSUE #1299: Partial Match checks "args" instead of "extended_args"
The partial match functionality currently checks `args` for both `json` and `like` mode.

I think this is a major bug since the arguments if > 191 chars are stored in the `extended_args` database columns.

The fix is to change:
- https://github.com/woocommerce/action-scheduler/blob/f5550b2176d604926783947822b752e56de03b02/classes/data-stores/ActionScheduler_DBStore.php#L513
- https://github.com/woocommerce/action-scheduler/blob/f5550b2176d604926783947822b752e56de03b02/classes/data-stores/ActionScheduler_DBStore.php#L520

Both to  `a.extended_args`

JUVOJustin:
`partial_args_matching` (`json` and `like` modes) fails to find actions when arguments exceed 191 characters because it only queries the `args` column, which contains an MD5 hash for long args while the actual JSON is in `extended_args`.

## Changes

- **`json` mode**: Use `COALESCE(a.extended_args, a.args)` so `JSON_EXTRACT` operates on actual JSON data
- **`like` mode**: Check both columns following the existing `search` query pattern: `((a.extended_args IS NULL AND a.args LIKE %s) OR a.extended_args LIKE %s)`

## Example

```php
// Before: only checks args column (contains MD5 hash for long args)
$sql .= ' AND JSON_EXTRACT(a.args, %s)=' . $placeholder;

// After: checks extended_args first if present
$sql .= ' AND JSON_EXTRACT(COALESCE(a.extended_args, a.args), %s)=' . $placeholder;
```

## Tests

Added test coverage for both `like` and `json` modes with:
- Short args (≤191 chars, stored in `args`)
- Long args (>191 chars, stored in `extended_args`)

Fix https://github.com/woocommerce/action-scheduler/issues/1299

### DIFF
--- a/classes/data-stores/ActionScheduler_DBStore.php
+++ b/classes/data-stores/ActionScheduler_DBStore.php
@@ -510,15 +510,16 @@ AND `group_id` = %d
 								)
 							);
 						}
-						$sql         .= ' AND JSON_EXTRACT(a.args, %s)=' . $placeholder;
+						$sql         .= ' AND JSON_EXTRACT(COALESCE(a.extended_args, a.args), %s)=' . $placeholder;
 						$sql_params[] = '$.' . $key;
 						$sql_params[] = $value;
 					}
 					break;
 				case 'like':
 					foreach ( $query['args'] as $key => $value ) {
-						$sql         .= ' AND a.args LIKE %s';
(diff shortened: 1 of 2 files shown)


JUVOJustin:
To be clear: This PR is done by github copilot. I reviewed and tested locally again.
Fixes https://github.com/woocommerce/action-scheduler/issues/1299

PR #1309, opened 2026-03-07
Performance: address side-effects of processing corrupted entries (logs table bloat, hard-working queue processing)

### LINKED ISSUE #1311: Logs tables bloat caused by corrupted action schedules
The issue resurfaced in p1772090743156589-slack-C08SVLULL3B: A-S built-in migration can potentially cause schedule data corruption (serialized data gets truncated), and A-S will keep trying to process it, spawning more logs.

kalessil:
Fixes #1311 ACTSCH-92

This update follows up on p1772090743156589-slack-C08SVLULL3B and addresses the handling of actions with corrupted data:
- Identifies actions with corrupted data and cancels their execution to allow for rescheduling.
- Removes excessive logs generated during attempts to process corrupted actions (gradually if it's the target bloat case).
- Makes the cancelled action available in the UI for investigation.

In this PR, we identify actions that fail to construct an action object after two attempts and cancel them, including purging their associated logs (gradually if needed).

From an architectural perspective, we are aligning replication-friendly cleanup methods with the WooCommerce core work I [completed](https://github.com/woocommerce/woocommerce/pull/60711) earlier to meet the performance requirements of HVM stores and addressing a known reliability issue.

Testing Instructions:

- Build the plugin from this branch and install it on your test site.
- Modify any pending `action_scheduler_run_recurring_actions_schedule_hook` by truncating part of the `schedule` field and updating its scheduled date to a past date.
- Navigate to the admin area under `Tools -> Action Scheduler`.
- Verify that the action status has changed and that there is a single log entry mentioning data corruption, which must appear in the UI.
- Delete the `as_is_ensure_recurring_actions_scheduled` transient and refresh the page.
- Verify that `action_scheduler_run_recurring_actions_schedule_hook` is rescheduled for 3 a.m. the next day.

### DIFF
--- a/classes/abstracts/ActionScheduler_Abstract_QueueRunner.php
+++ b/classes/abstracts/ActionScheduler_Abstract_QueueRunner.php
@@ -84,10 +84,9 @@ abstract class ActionScheduler_Abstract_QueueRunner extends ActionScheduler_Abst
 		 * exceptions (and an exception thrown from a catch block cannot be caught by a later catch block in the *same*
 		 * structure).
 		 */
+		$valid_action = true;
 		try {
 			try {
-				$valid_action = true;
-
 				do_action( 'action_scheduler_before_execute', $action_id, $context );
 
 				if ( ActionScheduler_Store::STATUS_PENDING !== $this->store->get_status( $action_id ) ) {
@@ -98,10 +97,18 @@ abstract class ActionScheduler_Abstract_QueueRunner extends ActionScheduler_Abst
 
 				do_action( 'action_scheduler_begin_execute', $action_id, $context );
 
-				$action = $this->store->fetch_action( $action_id );
+				$action = $this->fetch_complete_action( $action_id );
+				if ( null === $action ) {
+					$valid_action = false;
+					$this->cancel_corrupted_action( $action_id );
+					return;
+				}
+
 				$this->store->log_execution( $action_id );
 				$action->execute();
+
 				do_action( 'action_scheduler_after_execute', $action_id, $action, $context );
+
 				$this->store->mark_complete( $action_id );
 			} catch ( Throwable $e ) {
 				// Throwable is defined when executing under PHP 7.0 and up. We convert it to an exception, for
@@ -118,6 +125,44 @@ abstract class ActionScheduler_Abstract_QueueRunner extends ActionScheduler_Abst
 		if ( isset( $action ) && is_a( $action, 'ActionScheduler_Action' ) && $action->get_schedule()->is_recurring() ) {
 			$this->schedule_next_instance( $action, $action_id );
 		}
+	}
+
+	/**
+	 * Fetches the action instance. On failure returns null instead of ActionScheduler_NullAction.
+	 *
+	 * @param int $action_id Action ID.
+	 * @return ActionScheduler_Action|null
+	 */
+	private function fetch_complete_action( int $action_id ) {
+		// Handle corrupted entries and potential DB read failures (corrupted arguments or schedule data; a known trigger is incomplete A-S migration).
+		$attempts_to_populate = 0;
+		do {
+			$action  = $this->store->fetch_action( $action_id );
+			$failure = $action instanceof ActionScheduler_NullAction;
+			// Sleep 0.01 second before retrying.
+			usleep( 10000 * (int) $failure );
+		} while ( $failure && ++$attempts_to_populate < 2 );
+
+		return $action instanceof ActionScheduler_NullAction ? null : $action;
+	}
+
+	/**
+	 * Cancels the corrupted actions and performs the necessary cleanup to address identified side effects.
+	 *
+	 * @param int $action_id Action ID.
+	 * @return void
+	 */
+	private function cancel_corrupted_action( int $action_id ) {
+		$this->store->cancel_action( $action_id );
+
+		// Corrupted entries typically contain an excessive number of logs, which we intend to remove immediately.
+		do_action( 'action_scheduler_cancelled_corrupted_action', $action_id );
+
+		// Add a comment to clarify the missing logs and explain the reason for the cancellation.
+		ActionScheduler_Logger::instance()->log(
+			$action_id,
+			__( 'This action data appears to be corrupt. We are cancelling it to allow recurring actions to be re-created by extensions.', 'action-scheduler' )
+		);
 	}
 
 	/**
--- a/classes/abstracts/ActionScheduler_Abstract_QueueRunner.php
+++ b/classes/abstracts/ActionScheduler_Abstract_QueueRunner.php
@@ -156,7 +156,7 @@ abstract class ActionScheduler_Abstract_QueueRunner extends ActionScheduler_Abst
 		$this->store->cancel_action( $action_id );
 
 		// Corrupted entries typically contain an excessive number of logs, which we intend to remove immediately.
-		do_action( 'action_scheduler_cancelled_corrupted_action', $action_id );
+		do_action( 'action_scheduler_canceled_corrupted_action', $action_id );
 
 		// Add a comment to clarify the missing logs and explain the reason for the cancellation.
 		ActionScheduler_Logger::instance()->log(
--- a/classes/data-stores/ActionScheduler_DBLogger.php
+++ b/classes/data-stores/ActionScheduler_DBLogger.php
@@ -121,7 +121,7 @@ class ActionScheduler_DBLogger extends ActionScheduler_Logger {
 	 * @param int $action_id Action ID.
 	 */
 	public function clear_deleted_action_logs( $action_id ) {
-		$this->clear_deleted_action_logs_single_batch( $action_id, -1, 1, 100 );
+		$this->clear_deleted_action_logs_single_batch( $action_id, -1, 1, 4000 );
 	}
 
 	/**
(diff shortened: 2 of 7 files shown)


classes/data-stores/ActionScheduler_HybridStore.php
```diff
@@ -400,9 +400,14 @@ protected function get_store_from_action_id( $action_id, $primary_first = false
...
-			if ( ! is_a( $action, 'ActionScheduler_NullAction' ) ) {
-				return $store;
+			try {
+				// Probe by fetching the action status: if entry is corrupted, the object construction is not feasible.
+				if ( null !== $store->get_status( $action_id ) ) {
+					return $store;
+				}
+			} catch ( \InvalidArgumentException $e ) {
+				// Do nothing. In this context, not finding status for the given ID is not an error.
+				unset( $e ); // Make the code sniffer happy.
```
jorgeatorres: Not that it makes a ton of difference, but maybe we just add a `// phpcs:ignore` here. The `unset()` pattern is a bit strange.

classes/abstracts/ActionScheduler_Abstract_QueueRunner.php
```diff
@@ -120,6 +127,44 @@ function ( $type, $message ) {
...
+	 * Cancels the corrupted actions and performs the necessary cleanup to address identified side effects.
+	 *
+	 * @param int $action_id Action ID.
+	 * @return void
+	 */
+	private function cancel_corrupted_action( int $action_id ) {
+		$this->store->cancel_action( $action_id );
+
+		// Corrupted entries typically contain an excessive number of logs, which we intend to remove immediately.
+		do_action( 'action_scheduler_cancelled_corrupted_action', $action_id );
```
jorgeatorres: For consistency with other 'canceled' in A-S, can we use the American spelling?

classes/abstracts/ActionScheduler_Abstract_QueueRunner.php
```diff
@@ -120,6 +127,44 @@ function ( $type, $message ) {
...
+	 */
+	private function fetch_complete_action( int $action_id ) {
+		// Handle corrupted entries and potential DB read failures (corrupted arguments or schedule data; a known trigger is incomplete A-S migration).
+		$attempts_to_populate = 0;
+		do {
+			$action  = $this->store->fetch_action( $action_id );
+			$failure = $action instanceof ActionScheduler_NullAction;
+			// Sleep 0.01 second before retrying.
+			usleep( 10000 * (int) $failure );
+		} while ( $failure && ++$attempts_to_populate < 2 );
```
jorgeatorres, line 144: Is the sleep really useful here? I might be ignoring some ways for an action to become corrupted, but the most common scenario is that the schedule data is invalid. If that's the case, re-reading won't help (the row won't "fix itself").

As for transient db connection errors, `$wpdb` already retries the connection before returning, at least when the connection handle is dropped or when we get the "server has gone away" error, which is a bit more robust than the 10ms wait. In that particular scenario we wouldn't be adding much.

Did you have some other failure in mind?

jorgeatorres (commented):
Hey @kalessil!

I like where this is going. Left some comments for discussion. Let me know what you think!

classes/data-stores/ActionScheduler_HybridStore.php
kalessil: Yep, we can do that as well. I'll update it as I start my week.

classes/abstracts/ActionScheduler_Abstract_QueueRunner.php
kalessil: I went extremely defensive here against errors during reads: given we'll be cancelling the action, I aim for 200% to rule out failed read operations and replication lags (I burned on this exact pattern in the past).

classes/abstracts/ActionScheduler_Abstract_QueueRunner.php
kalessil: Sure, here we go: bdb88b55d081fb49049aafdc86a077c0a0152382

classes/data-stores/ActionScheduler_HybridStore.php
kalessil: Why not, here we go: cdeacc176b3bc050e4a78e06bbb779338ad5c57d

classes/data-stores/ActionScheduler_DBLogger.php
```diff
@@ -110,17 +110,43 @@ public function init() {
...
 	public function clear_deleted_action_logs( $action_id ) {
-		/** @var \wpdb $wpdb */ //phpcs:ignore Generic.Commenting.DocComment.MissingShort
+		$this->clear_deleted_action_logs_single_batch( $action_id, -1, 1, 100 );
+	}
+
+	/**
+	 * Delete the action logs batch for an action.
+	 *
+	 * @param int  $action_id      Action id.
+	 * @param int  $cutoff_log_id  Cutoff log ID. Retain logs generated after the cleanup begins. A value of -1 indicates the start of the cleanup.
```
prettyboymp, line 131: A batch size of 100 is very conservative here given the cost of each batch. Each continuation schedules a full AS action, which involves inserting an action row, creating 3 lifecycle log entries (stored/started/completed), claiming, executing, and marking complete. That per-batch overhead almost certainly exceeds the cost of just deleting more rows.

For the scenario this PR targets (169M log rows), batch_size=100 creates ~1.69M cleanup actions and ~5.07M compounding log entries — turning cleanup into a data amplification problem.

For comparison, the WooCommerce core PR #60711 (referenced in this PR's description) used an in-process `do/while` loop with `usleep()` between batches, creating zero additional AS actions. If per-action scheduling is intentional here (to avoid PHP timeout on very large tables), the batch size should be significantly higher to offset the scheduling overhead.

classes/data-stores/ActionScheduler_DBLogger.php
```diff
@@ -110,17 +110,43 @@ public function init() {
...
+
+		if ( -1 === $cutoff_log_id ) {
+			$cutoff_log_id = 1 + (int) $wpdb->get_var( $wpdb->prepare( "SELECT MAX(log_id) FROM {$wpdb->actionscheduler_logs} WHERE action_id = %d", $action_id ) );
+		}
+
+		$deleted  = (int) $wpdb->query( $wpdb->prepare( "DELETE FROM {$wpdb->actionscheduler_logs} WHERE action_id = %d AND log_id < %d LIMIT %d", $action_id, $cutoff_log_id, $batch_size ) );
+		$continue = $deleted === $batch_size;
+		if ( $continue ) {
+			// Schedule immediately, as this action will not be selected during the current run and will have lower than normal priority.
+			as_schedule_single_action( time(), 'action_scheduler_clear_deleted_action_logs_hook', array( $action_id, $cutoff_log_id, ++$batch, $batch_size ), '', false, 9 );
```
prettyboymp: Priority 9 means these cleanup actions are claimed *before* default-priority (10) actions, since the claim query uses `ORDER BY priority ASC`. This means log cleanup would preempt business-critical scheduled work like subscription renewals.

Background maintenance like log deletion should yield to store operations, not the other way around. Consider using a priority > 10 (e.g., 11 or higher) so cleanup runs when the queue isn't busy with higher-priority work.

prettyboymp (changes requested):
Two concerns with the batched log cleanup approach in `ActionScheduler_DBLogger`:

1. **Batch size of 100 is likely too low given the per-batch overhead** — Each continuation batch schedules a full AS action (INSERT into actions table + 3 lifecycle log entries + claim + execute + mark complete). That overhead almost certainly exceeds the cost of simply deleting more rows per batch. For the 169M row scenario this PR targets, a batch size of 100 produces ~1.69M cleanup actions and ~5.07M compounding log entries. Increasing the batch size would significantly reduce the scheduling overhead while still being replication-friendly.

2. **Cleanup action priority 9 preempts business-critical work** — Priority 9 is higher than the default 10 (`ORDER BY priority ASC`), so cleanup actions are claimed *before* default-priority actions like subscription renewals. Background log cleanup shouldn't take precedence over store operations.

classes/data-stores/ActionScheduler_DBLogger.php
kalessil: @prettyboymp, quick question: do you check the data volume impact manually, or does your Claude setup handle it automatically during reviews?

kalessil: In commit 548fcb2e677e71e98950ef9e5af17b5b739008b1, I've increased the batch size to 4000 based on the following constraints provided to Claude for selecting the optimal value:
- The replication log should not be heavily impacted. The maximum allowable batch size is 5000.
- The batch size should allow clearing the largest expected case, a 30GB logs table, within a maximum of 30 days.

For the scenario this PR targets (169M log rows), batch_size=4000 creates ~42,250 cleanup actions and ~126,750 compounding log entries — a 40x reduction in scheduling overhead and amplification of log rows compared to batch_size=100.

classes/data-stores/ActionScheduler_DBLogger.php
kalessil: Addressed in c1fae5de78c30dc5d8d3dee9ec9e1a0e7403f85f.

```
Suggest: 20

  Reasons:
  - Clears all business operations (0, 1, 5, 10) without competing
  - Matches AS's own established convention for background maintenance (action_scheduler_run_recurring_actions_schedule_hook is explicitly infrastructure-level)
  - On HVM stores, queue slots aren't 100% saturated at all times — priority-20 actions still get picked up reliably enough to hit the 30-day target
  - Going higher (50, 100) risks starving the cleanup on a genuinely busy store and breaking the 30-day assumption

```

prettyboymp (approved):

Issue #912, opened 2023-02-22
Clean-up enhancements


barryhughes:
Especially in the context of very busy sites, it is entirely possible (and we have seen real world examples of this) for the queue cleaner—which normally takes care of removing completed and expired actions from the database—to 'fall behind', meaning the actions table continues to grow. We wish to introduce a number of changes to help combat this.

Internal link: p7bje6-4ph-p2

Goals:

- [x] Dedicated queue cleaning process (instead of 'piggy backing' on the queue runner).
- [x] Ability to trigger the above via WP CLI.
- [ ] Ability to modify the retention period via the admin UI.
- [ ] Consider a flexible retention period (ie, if expired actions are not old enough to be cleaned up, but the total number of expired actions exceeds a certain threshold, start cleaning anyway).
- [x] Look into clean-up of *failed* actions.

marsjaninzmarsa:
great to hear that, we've recently encountered mentioned issues with one of the client sites - the queue runner wasn't able to finish processing the queue due to a massive load of uncleaned jobs bloating database, which was a very unpleasant surprise with in other ways fantastic solution

barryhughes:
https://github.com/woocommerce/action-scheduler/pull/907 has been merged and introduces `wp action-scheduler clean`.

kalessil:
#1310 will address `Look into clean-up of failed actions`.

kalessil:
https://github.com/woocommerce/action-scheduler/pull/1312 will address `Dedicated queue cleaning process (instead of 'piggy backing' on the queue runner).`

PR #1310, opened 2026-03-07
Performance: cap failed actions retention period to three months by default

kalessil:
Partially addresses ACTSCH-15 #912.

The issue resurfaces in stats shared under p1772090743156589-slack-C08SVLULL3B. This PR limits the retention period for failed actions to three months to prevent failed actions from continuously increasing the A-S table size.

We are addressing A-S scalability issues in HVM caused by the accumulation of failed actions, which increases database table size and slows A-S performance. As the database is often the main bottleneck, slower A-S queries further impact performance, especially during BFCM.

> Note: The actions scheduler allows you to customize retention periods for action statuses, but these settings apply to all statuses collectively. However, failed statuses require a longer retention period than others to support troubleshooting while maintaining a manageable actions table size. This PR addresses this need.

> BC break note: By default, failed actions are purged after three months. Values returned by `action_scheduler_retention_period` are enforced more strictly. If the value cannot be resolved as a positive integer, a one-month retention period will be applied as a fallback.

Testing Instructions:

- Build the plugin from this branch and install it on your test system.
- Modify a completed action by setting its status to failed and updating all date fields to twelve months in the past.
- Navigate to the admin area.
- Confirm that the action record has been deleted.

### DIFF
--- a/classes/ActionScheduler_QueueCleaner.php
+++ b/classes/ActionScheduler_QueueCleaner.php
@@ -107,7 +107,7 @@ class ActionScheduler_QueueCleaner {
 		$deleted_failed_entries = array();
 		// Backward compatibility note: if store already purging the failed statuses, don't change the behaviour.
 		if ( $lifespan_failed > 0 && ! in_array( ActionScheduler_Store::STATUS_FAILED, $statuses_to_purge, true ) ) {
-			// use a fixed default batch size to ensure that the cleanup of failed actions does not interfere with the regular cleanup.
+			// Use a fixed default batch size to ensure that the cleanup of failed actions does not interfere with the regular cleanup.
 			$deleted_failed_entries = $this->clean_actions( array( ActionScheduler_Store::STATUS_FAILED ), $cutoff_failed, 20 );
 		}
 
--- a/classes/ActionScheduler_QueueCleaner.php
+++ b/classes/ActionScheduler_QueueCleaner.php
@@ -107,7 +107,7 @@ class ActionScheduler_QueueCleaner {
 		$deleted_failed_entries = array();
 		// Backward compatibility note: if store already purging the failed statuses, don't change the behaviour.
 		if ( $lifespan_failed > 0 && ! in_array( ActionScheduler_Store::STATUS_FAILED, $statuses_to_purge, true ) ) {
-			// We will use a hard-coded default batch size, as recommended in the code review.
+			// use a fixed default batch size to ensure that the cleanup of failed actions does not interfere with the regular cleanup.
 			$deleted_failed_entries = $this->clean_actions( array( ActionScheduler_Store::STATUS_FAILED ), $cutoff_failed, 20 );
 		}
 
--- a/classes/ActionScheduler_QueueCleaner.php
+++ b/classes/ActionScheduler_QueueCleaner.php
@@ -50,6 +50,7 @@ class ActionScheduler_QueueCleaner {
 	/**
 	 * Default queue cleaner process used by queue runner.
 	 *
+	 * @since 3.9.4 by default, failed actions are removed after three months.
 	 * @return array
 	 */
 	public function delete_old_actions() {
@@ -60,8 +61,28 @@ class ActionScheduler_QueueCleaner {
 		 */
 		$lifespan = apply_filters( 'action_scheduler_retention_period', $this->month_in_seconds );
 
+		/**
+		 * Set the retention period, in seconds, for actions with a status returned by the action_scheduler_default_cleaner_statuses filter.
+		 *
+		 * @param int $retention_period Retention period in seconds.
+		 */
+		$lifespan_default = max( 0, (int) apply_filters( 'action_scheduler_retention_period_by_default', $lifespan ) );
+		$lifespan_default = $lifespan_default > 0 ? $lifespan_default : $this->month_in_seconds;
+
+		/**
+		 * Set the retention period in seconds for actions with a failed status. If the action_scheduler_default_cleaner_statuses filter includes
+		 * a failed status, this filter result will be ignored, and the retention period for failed actions will match that of other statuses.
+		 *
+		 * @param int $retention_period Retention period in seconds.
+		 */
+		$lifespan_failed = max( 0, (int) apply_filters( 'action_scheduler_retention_period_for_failed', 3 * $this->month_in_seconds ) );
+		// We considered 12-month, 3-month, and 1-month options for failed action retention and selected a 3-month period
+		// to align with the quarterly accounting cycle. Store owners may adjust the retention period to achieve PCI DSS
+		// compliance or to align with a different accounting cycle, as needed.
+
 		try {
-			$cutoff = as_get_datetime_object( $lifespan . ' seconds ago' );
+			$cutoff_failed  = as_get_datetime_object( $lifespan_failed . ' seconds ago' );
+			$cutoff_default = as_get_datetime_object( $lifespan_default . ' seconds ago' );
 		} catch ( Exception $e ) {
 			_doing_it_wrong(
 				__METHOD__,
@@ -83,7 +104,16 @@ class ActionScheduler_QueueCleaner {
 		 */
 		$statuses_to_purge = (array) apply_filters( 'action_scheduler_default_cleaner_statuses', $this->default_statuses_to_purge );
 
-		return $this->clean_actions( $statuses_to_purge, $cutoff, $this->get_batch_size() );
+		$deleted_failed_entries = array();
+		// Backward compatibility note: if store already purging the failed statuses, don't change the behaviour.
+		if ( $lifespan_failed > 0 && ! in_array( ActionScheduler_Store::STATUS_FAILED, $statuses_to_purge, true ) ) {
+			// We will use a hard-coded default batch size, as recommended in the code review.
+			$deleted_failed_entries = $this->clean_actions( array( ActionScheduler_Store::STATUS_FAILED ), $cutoff_failed, 20 );
+		}
+
+		$deleted_entries = $this->clean_actions( $statuses_to_purge, $cutoff_default, $this->get_batch_size() );
+
+		return array_merge( $deleted_failed_entries, $deleted_entries );
 	}
 
 	/**
@@ -105,7 +135,6 @@ class ActionScheduler_QueueCleaner {
 		}
 
 		$deleted_actions = array();
-
 		foreach ( $statuses_to_purge as $status ) {
 			$actions_to_delete = $this->store->query_actions(
 				array(
@@ -116,11 +145,10 @@ class ActionScheduler_QueueCleaner {
 					'orderby'          => 'none',
 				)
 			);
-
-			$deleted_actions = array_merge( $deleted_actions, $this->delete_actions( $actions_to_delete, $lifespan, $context ) );
+			$deleted_actions[] = $this->delete_actions( $actions_to_delete, $lifespan, $context );
 		}
 
-		return $deleted_actions;
+		return array_merge( array(), ...$deleted_actions );
 	}
 
 	/**
--- a/classes/ActionScheduler_QueueRunner.php
+++ b/classes/ActionScheduler_QueueRunner.php
@@ -153,8 +153,10 @@ class ActionScheduler_QueueRunner extends ActionScheduler_Abstract_QueueRunner {
 
 		$this->processed_actions_count = 0;
 		if ( false === $this->has_maximum_concurrent_batches() ) {
+			$batch_size = apply_filters( 'action_scheduler_queue_runner_batch_size', 25 );
+			// Note: gc_collect_cycles() was considered here and in do_batch/clear_caches, but rejected:
+			// upside is speculative, GC sweep cost scales with object count and can cause pauses.
 			do {
-				$batch_size                     = apply_filters( 'action_scheduler_queue_runner_batch_size', 25 );
 				$processed_actions_in_batch     = $this->do_batch( $batch_size, $context );
 				$this->processed_actions_count += $processed_actions_in_batch;
 			} while ( $processed_actions_in_batch > 0 && ! $this->batch_limits_exceeded( $this->processed_actions_count ) ); // keep going until we run out of actions, time, or memory.
@@ -180,11 +182,13 @@ class ActionScheduler_QueueRunner extends ActionScheduler_Abstract_QueueRunner {
 		$this->monitor->attach( $claim );
 		$processed_actions = 0;
 
+		$claim_id = $claim->get_id();
 		foreach ( $claim->get_actions() as $action_id ) {
 			// bail if we lost the claim.
-			if ( ! in_array( $action_id, $this->store->find_actions_by_claim_id( $claim->get_id() ), true ) ) {
+			if ( $claim_id !== $this->store->get_claim_id( $action_id ) ) {
 				break;
 			}
+
 			$this->process_action( $action_id, $context );
 			$processed_actions++;
 
--- a/classes/data-stores/ActionScheduler_DBStore.php
+++ b/classes/data-stores/ActionScheduler_DBStore.php
@@ -1231,9 +1231,13 @@ AND `group_id` = %d
 
 		$updated = $wpdb->update(
 			$wpdb->actionscheduler_actions,
-			array( 'status' => self::STATUS_FAILED ),
+			array(
+				'status'             => self::STATUS_FAILED,
+				'last_attempt_gmt'   => current_time( 'mysql', true ),
+				'last_attempt_local' => current_time( 'mysql' ),
+			),
 			array( 'action_id' => $action_id ),
-			array( '%s' ),
+			array( '%s', '%s', '%s' ),
 			array( '%d' )
 		);
 		if ( empty( $updated ) ) {
@@ -1301,7 +1305,7 @@ AND `group_id` = %d
 				'last_attempt_local' => current_time( 'mysql' ),
 			),
 			array( 'action_id' => $action_id ),
-			array( '%s' ),
+			array( '%s', '%s', '%s' ),
 			array( '%d' )
 		);
 		if ( empty( $updated ) ) {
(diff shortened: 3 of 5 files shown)


classes/data-stores/ActionScheduler_DBStore.php
```diff
@@ -1222,7 +1222,11 @@ public function mark_failure( $action_id ) {
 
 		$updated = $wpdb->update(
 			$wpdb->actionscheduler_actions,
-			array( 'status' => self::STATUS_FAILED ),
+			array(
```
kalessil, line 1234: Bug-fix: align with marking completion; surfaced in failing tests.

classes/ActionScheduler_QueueCleaner.php
```diff
@@ -116,11 +146,10 @@ public function clean_actions( array $statuses_to_purge, DateTime $cutoff_date,
 					'orderby'          => 'none',
 				)
 			);
-
-			$deleted_actions = array_merge( $deleted_actions, $this->delete_actions( $actions_to_delete, $lifespan, $context ) );
+			$deleted_actions[] = $this->delete_actions( $actions_to_delete, $lifespan, $context );
 		}
 
-		return $deleted_actions;
+		return array_merge( ...$deleted_actions );
```
kalessil: Lower algorithmic complexity here - if batch size increases, this holds on solid.

classes/ActionScheduler_QueueRunner.php
```diff
@@ -180,11 +180,13 @@ protected function do_batch( $size = 100, $context = '' ) {
 		$this->monitor->attach( $claim );
 		$processed_actions = 0;
 
+		$claim_id = $claim->get_id();
 		foreach ( $claim->get_actions() as $action_id ) {
 			// bail if we lost the claim.
-			if ( ! in_array( $action_id, $this->store->find_actions_by_claim_id( $claim->get_id() ), true ) ) {
+			if ( $claim_id !== $this->store->get_claim_id( $action_id ) ) {
```
kalessil, line 188: Lower algorithmic complexity here - if batch size increases, this holds on solid and performs instantly.

classes/ActionScheduler_QueueCleaner.php
```diff
@@ -83,7 +103,18 @@ public function delete_old_actions() {
...
 
-		return $this->clean_actions( $statuses_to_purge, $cutoff, $this->get_batch_size() );
+		$batch_size             = $this->get_batch_size();
+		$deleted_failed_entries = array();
+		// Backward compatibility note: if store already purging the failed statuses, don't change the behaviour.
+		if ( ! in_array( ActionScheduler_Store::STATUS_FAILED, $statuses_to_purge, true ) ) {
+			$deleted_failed_entries = $this->clean_actions( array( ActionScheduler_Store::STATUS_FAILED ), $cutoff_failed, $batch_size );
+			$count_statuses         = empty( $statuses_to_purge ) ? count( $this->default_statuses_to_purge ) : count( $statuses_to_purge );
+			$batch_size             = (int) ( ( ( $count_statuses * $batch_size ) - count( $deleted_failed_entries ) ) / $count_statuses );
+		}
```
prettyboymp, line 112: The batch_size recalculation here can produce zero when `$count_statuses` is 1 (a single status via the `action_scheduler_default_cleaner_statuses` filter) and all `$batch_size` items are deleted. A `per_page` of 0 in `query_actions()` skips the `LIMIT` clause entirely, resulting in an unbounded query that returns all matching rows for deletion.

With the default 2 statuses this can't happen (minimum result is 10), but an early return when the budget is exhausted would be safer:
```php
if ( $batch_size <= 0 ) {
    return $deleted_failed_entries;
}
```

classes/ActionScheduler_QueueCleaner.php
```diff
@@ -50,6 +50,7 @@ public function __construct( ?ActionScheduler_Store $store = null, $batch_size =
 	/**
 	 * Default queue cleaner process used by queue runner.
 	 *
+	 * @since 3.9.4 by default, failed actions are removed after three months.
```
prettyboymp, line 53: This is a backward-incompatible change to documented behavior. `docs/perf.md` currently states: *"By default, Action Scheduler does not automatically delete old failed actions."* — and this PR directly reverses that guarantee.

The change is well-justified (failed action accumulation is a documented problem across P2s and merchant cases), but it would be good to:
1. Update `docs/perf.md` to reflect the new default behavior.
2. Document a clean opt-out path for sites that intentionally keep failed actions (e.g., for debugging, auditing, or compliance). Currently returning a very large value from `action_scheduler_retention_period_for_failed` is the only workaround, which isn't obvious.
3. Consider a changelog/upgrade notice since this changes data retention defaults.

classes/ActionScheduler_QueueCleaner.php
```diff
@@ -142,6 +171,9 @@ private function delete_actions( array $actions_to_delete, $lifespan = null, $co
 			try {
 				$this->store->delete_action( $action_id );
 				$deleted_actions[] = $action_id;
```
prettyboymp, line 172: What's the reasoning for adding this unconditionally at the per-action level? This affects all deletion paths including WP CLI `action-scheduler clean`, which already has its own `--pause` flag for inter-batch throttling. At default batch sizes (~60 actions) it adds ~60ms which is fine, but WP CLI operations processing thousands of actions accumulate seconds of pure sleep time (10K actions = +10 seconds).

Would it make more sense to apply this at the batch level (e.g., between status groups in `clean_actions()`) rather than per-action? That would still throttle replication log pressure while keeping the total overhead constant regardless of batch size.

classes/ActionScheduler_QueueCleaner.php
kalessil: I'll update the file to include a note about the BC break and instructions for replicating the original behavior. I have also added the corresponding filter for this case.

classes/ActionScheduler_QueueCleaner.php
kalessil: I evaluated both per-action and per-batch approaches. The per-item option maintains a consistent stress level for replication, regardless of batch size.

For example, running the CLI with 10,000 actions demonstrates that the per-batch option causes spikes in the replication log, while the per-item option ensures a steady flow without excessive stress.

classes/ActionScheduler_QueueCleaner.php
kalessil: Good defensive addition: added in af7d7012afc09c6ff58c8113bc27e0193af97b58

classes/ActionScheduler_QueueCleaner.php
kalessil: Updating the docs in af7d7012afc09c6ff58c8113bc27e0193af97b58

classes/ActionScheduler_QueueCleaner.php
kalessil: We can also use half of that value, and the change should still be effective.

kalessil: Removing this delay in 44d2096e1bb8f995439e1916b99eb0beca140236: this needs to be refined as part of https://github.com/woocommerce/action-scheduler/pull/1312, which will delete actions on schedule instead of ad-hoc (ad-hoc is not really suited for HVMs/BFCM).

classes/ActionScheduler_QueueCleaner.php
```diff
@@ -60,8 +61,27 @@ public function delete_old_actions() {
...
+		 */
+		$lifespan_default = apply_filters( 'action_scheduler_retention_period_by_default', $lifespan );
+
+		/**
+		 * Set the retention period in seconds for actions with a failed status. If the action_scheduler_default_cleaner_statuses filter includes
+		 * a failed status, this filter result will be ignored, and the retention period for failed actions will match that of other statuses.
+		 *
+		 * @param int $retention_period Retention period in seconds.
+		 */
+		$lifespan_failed = apply_filters( 'action_scheduler_retention_period_for_failed', 3 * $this->month_in_seconds );
```
jorgeatorres: Probably excessive caution, an the existing `$lifespan` didn't consider this, but I wonder if we should pass these lifespans through `absint()` or at least `abs()` so there's no chance we end up with a date in the future.

classes/ActionScheduler_QueueCleaner.php
```diff
@@ -83,7 +103,21 @@ public function delete_old_actions() {
...
+		if ( ! in_array( ActionScheduler_Store::STATUS_FAILED, $statuses_to_purge, true ) ) {
+			$deleted_failed_entries = $this->clean_actions( array( ActionScheduler_Store::STATUS_FAILED ), $cutoff_failed, $batch_size );
+			$count_statuses         = empty( $statuses_to_purge ) ? count( $this->default_statuses_to_purge ) : count( $statuses_to_purge );
+			$batch_size             = (int) ( ( ( $count_statuses * $batch_size ) - count( $deleted_failed_entries ) ) / $count_statuses );
+			if ( $batch_size <= 0 ) {
+				return $deleted_failed_entries;
+			}
+		}
+
+		$deleted_entries = $this->clean_actions( $statuses_to_purge, $cutoff_default, $batch_size );
```
jorgeatorres: While this is clever, I wonder if we should always let actions in other statuses be deleted. If there's a large backlog of failed actions, we will only be removing failed actions for a while. Maybe we can split the 'budget' since the beginning (even if we give more to failed) so that we always clean up at least some actions of the other statuses?

docs/perf.md
```diff
@@ -129,7 +129,16 @@ add_action( 'wp_ajax_nopriv_eg_create_additional_runners', 'eg_create_additional
...
 ## Cleaning Failed Actions
 
-By default, Action Scheduler does not automatically delete old failed actions. There are two optional methods of removing these actions:
+As of version 3.9.4, Action Scheduler retains failed actions for three months by default. If 'failed' is included in the purge statuses, the original behavior remains.
+The retention period for failed actions can be customized:
+
+```php
+add_filter( 'action_scheduler_retention_period_for_failed', function( $retention_period ) {
+    return 12 * MONTH_IN_SECONDS;
+} );
```
jorgeatorres, line 138: I wonder if we could change the filter so that `0` actually prevents removal and we document it here as a way to "bring back" the old behavior. I prefer the new one, but might be a way to soften the backwards incompatible change.

jorgeatorres (commented):
I left some comments for discussing @kalessil, but nothing too serious. I really like the improvements! 💯

classes/ActionScheduler_QueueCleaner.php
kalessil: Do you mean only for this specific filter @jorgeatorres ?

classes/ActionScheduler_QueueCleaner.php
kalessil: Removing failed actions will use part of the execution budget, but at least half will remain available for cleaning other statuses.

`$batch_size = (int) ( ( ( $count_statuses * $batch_size ) - count( $deleted_failed_entries ) ) / $count_statuses );` ensures that, even after removing failed entries, the remaining batch size is distributed evenly among the other statuses.

This approach ensures we remain within the same execution budget and that cleaning failed statuses does not prevent other statuses from being processed.

docs/perf.md
kalessil: Sure, that leaves a loophole where A-S can be blamed for bloating tables, but it's a fair point to provide an opportunity to restore original behaviour. I'll add the necessary changes today.

classes/ActionScheduler_QueueCleaner.php
kalessil: For improved execution budget management, please refer to https://github.com/woocommerce/action-scheduler/pull/1312. In this case (here), I prefer a simpler approach.

classes/ActionScheduler_QueueCleaner.php
jorgeatorres: I think it should apply to all lifespan variables, unless I'm missing something. My concern is that a negative value could result in a date in the future, so we end up deleting more than necessary. Am I wrong?

docs/perf.md
jorgeatorres: Yes, I don't think this is a blocker, but the filter exists so if for some reason someone needs to do this they can use it instead of having to do very strange things. And, in any case, it'd be a conscious choice.

classes/ActionScheduler_QueueCleaner.php
jorgeatorres: Yes, that's true when there are at least 2 different statuses returned by filter `action_scheduler_default_cleaner_statuses` (which is the default), but when there's just one, the total budget will be `$batch_size * 1 = $batch_size` and the failed budget will be `$batch_size` so we're left with no budget to process the remaining statuses, assuming there were at least `$batch_size` failed actions, which wouldn't be uncommon.

kalessil: Gotcha, I see your point. I'll look into handling this edge case better.

classes/ActionScheduler_QueueCleaner.php
kalessil: With the piggy-back style of cleaning as it is now, that's technically possible. I'll go ahead with extra defence against the mentioned cases.

docs/perf.md
kalessil: Addressed in 24eab6106fe8898f03170e9dcba08cc4f5ce1e28

classes/ActionScheduler_QueueCleaner.php
kalessil: Addressed in 5c4deedfc420038d70acd8205729e6203a2d8d8b

classes/ActionScheduler_QueueCleaner.php
kalessil: Addressed in 24eab6106fe8898f03170e9dcba08cc4f5ce1e28

classes/ActionScheduler_QueueCleaner.php
```diff
@@ -83,7 +103,23 @@ public function delete_old_actions() {
...
+			if ( $batch_size <= 0 ) {
+				// In this edge case, $statuses_to_purge contains a single status. The budget formula assumes
+				// two statuses, so the remaining budget is reset to the default to ensure this status is processed.
+				$batch_size = $this->get_batch_size();
+			}
+		}
+
+		$deleted_entries = $this->clean_actions( $statuses_to_purge, $cutoff_default, $batch_size );
+
+		return array_merge( $deleted_failed_entries, $deleted_entries );
```
jorgeatorres: I appreciate the changes but I'm still unsure if all this logic is really needed. Do you think just running the two deletions with the same batch size could be problematic? I don't think an extra 20 deletions would be a problem, and we wouldn't have to do any arithmetic or explaining.

What do you think?

classes/data-stores/ActionScheduler_DBStore.php
```diff
@@ -1222,7 +1222,11 @@ public function mark_failure( $action_id ) {
...
+				'status'             => self::STATUS_FAILED,
+				'last_attempt_gmt'   => current_time( 'mysql', true ),
+				'last_attempt_local' => current_time( 'mysql' ),
+			),
```
jorgeatorres, line 1238: Note that this requires some changes to the `$format` argument to `$wpdb->update()` below.

classes/ActionScheduler_QueueCleaner.php
kalessil: While a default batch size of 20 would allow for simplification, the batch size can be [customized](https://github.com/woocommerce/action-scheduler/blob/08d69eaf9c4434939e57c836c50e7b832afda4bf/classes/ActionScheduler_QueueCleaner.php#L252C34-L252C69) using the `action_scheduler_cleanup_batch_size` filter. Many stores likely use this option to manage the size of their scheduler tables.

This is the primary reason for the added complexity.

classes/data-stores/ActionScheduler_DBStore.php
kalessil: I'll look into this one tomorrow.

classes/ActionScheduler_QueueCleaner.php
jorgeatorres: Yes, I get that, but we can clamp the value to 20 if that's our concern or some other reasonable value. Would that still be problematic in your opinion?

kalessil: If you suggest using a hard-coded batch size of 20 for failed actions deletion, that's a safe, doable, and elegantly grounded approach. Can you confirm, so I can made the necessary changes tomorrow?

jorgeatorres: That's what I'm suggesting, yes.

kalessil: Great suggestion, I'll apply changes and ping you back for the review.

kalessil: Addressed in d1931973e39bb7d863a8819e2b23b5d231c19867

classes/data-stores/ActionScheduler_DBStore.php
kalessil: Addressed in f8e6ff1f7791c961b63a7b45617f1356c2b1d382

classes/ActionScheduler_QueueCleaner.php
```diff
@@ -60,8 +61,27 @@ public function delete_old_actions() {
 		 */
 		$lifespan = apply_filters( 'action_scheduler_retention_period', $this->month_in_seconds );
```
jorgeatorres, line 62: Note that if this returns something like a negative integer or an invalid string, we will transform it to `0` below and will end up deleting all complete/failed actions regardless of their last attempted date. Maybe we should ensure that this is always `> 0` (otherwise use our default of 1 month) so there's always at least _some_ retention period or we handle invalid values more gracefully than assuming "delete now".

jorgeatorres (approved):
LGTM. Confirmed that failed actions attempted more than 3 months ago are deleted as expected, leaving the rest untouched.

I'm leaving an otional comment, but pre-approving. Thank you @kalessil!

classes/ActionScheduler_QueueCleaner.php
kalessil: Added in abb71d7c3e6c1f38f65c11b1d905ddc6b7490f7a

PR #1312, opened 2026-03-12
Performance: extract action deletions into a separate task

kalessil:

Partially addresses ACTSCH-15 https://github.com/woocommerce/action-scheduler/issues/912.

A new `action_scheduler_run_actions_cleanup_hook` recurring action has been introduced to clean scheduler tables outside of business hours (3 am). This process uses predictable throughput, cleans batches of 250 (minimum) actions per status, and optimizes execution budgets.

> Note: executed with high priority as the task is fast and critical for maintaining Action Scheduler performance independently of how full the queue is.

If a custom cleaner is in use, the cleanup behavior remains unchanged. Actions are cleaned in batches of 20 (by default) per status, with cleanup performed at the start of the queue run via cron or CLI.

The main changes are contained in `\ActionScheduler_QueueCleaner::clean_actions`, which alternates between piggy-back and scheduled cleanup to ensure backward compatibility and predictable throughput accordingly.

This change primarily targets WooCommerce HVMs, ensuring that cleanup parameters are predictable and do not disrupt business operations during working hours.

> Backward compatibility: These changes apply only when the standard ActionScheduler_QueueCleaner is in use. Custom cleaner subclasses preserve the previous cleanup behavior.

Testing the New Default Behavior:
- Build the scheduler from this branch (e.g., `pnpm build`) and install it on your test site as a standard WordPress plugin.
- Navigate to the admin area, then go to `Tools -> Scheduled Actions`.
- Search for `action_scheduler_run_recurring_actions_schedule_hook` and execute it.
- Verify `action_scheduler_run_actions_cleanup_hook` recurring action appeared and scheduled to 3am the next day (site TZ)
- Run the action and confirm it completes successfully. Although trailing actions are covered by unit tests, your testing site may generate multiple trailing cleanup actions. Please allow time for these actions to finish, and ensure that action spawning stops. At the end, verify that only one pending `action_scheduler_run_actions_cleanup_hook` action remains.
- Verify that `wp action-scheduler run && wp action-scheduler clean` completes successfully.

Testing original behavior with a custom cleaner:
- Add the code from below, e.g., as an MU-plugin
- Attempt to run `action_scheduler_run_actions_cleanup_hook` from admin again and verify its execution log mentions missing callback hook (if customer has this customization already in place, the action will not be scheduled at all, so failure here is expected)
- Verify that `wp action-scheduler run && wp action-scheduler clean` completes successfully.

```php
add_filter(
	'action_scheduler_queue_runner_class',
	function () {
		class CustomQueueRunner extends ActionScheduler_QueueRunner {
			public function __construct( ?ActionScheduler_Store $store = null, ?ActionScheduler_FatalErrorMonitor $monitor = null, ?ActionScheduler_QueueCleaner $cleaner = null, ?ActionScheduler_AsyncRequest_QueueRunner $async_request = null ) {
				parent::__construct( $store, $monitor, new class() extends ActionScheduler_QueueCleaner {}, $async_request );
			}
		}

		return CustomQueueRunner::class;
	}
);
```

### DIFF
--- a/classes/ActionScheduler_QueueCleaner.php
+++ b/classes/ActionScheduler_QueueCleaner.php
@@ -85,7 +85,7 @@ class ActionScheduler_QueueCleaner {
 	}
 
 	/**
-	 * Default queue cleaner process used by queue runner.
+	 * Handles action deletion for both the scheduled cleanup task and the inline cleanup cycle.
 	 *
 	 * @since 3.9.4 by default, failed actions are removed after three months.
 	 * @return array
--- a/classes/ActionScheduler_QueueCleaner.php
+++ b/classes/ActionScheduler_QueueCleaner.php
@@ -226,7 +226,7 @@ class ActionScheduler_QueueCleaner {
 
 			$fetched_actions_count      = count( $actions_to_delete );
 			$iteration_unused_budget    = $is_scheduled_cleanup ? ( $iteration_execution_budget - $fetched_actions_count ) : 0;
-			$continue_scheduled_cleanup = $continue_scheduled_cleanup || ( $fetched_actions_count === $iteration_execution_budget );
+			$continue_scheduled_cleanup = $continue_scheduled_cleanup || ( $iteration_execution_budget === $fetched_actions_count );
 		}
 
 		if ( $is_scheduled_cleanup && $continue_scheduled_cleanup ) {
--- a/classes/ActionScheduler_QueueCleaner.php
+++ b/classes/ActionScheduler_QueueCleaner.php
@@ -171,6 +171,9 @@ class ActionScheduler_QueueCleaner {
 			try {
 				$this->store->delete_action( $action_id );
 				$deleted_actions[] = $action_id;
+				// Pause for 1ms to prevent excessive action and log deletion queries from flooding the replication log in clustered environments.
+				// This adds approximately 50–75ms per queue run at default batch sizes, which is negligible relative to the execution time budget.
+				usleep( 1000 );
 			} catch ( Exception $e ) {
 				/**
 				 * Notify 3rd party code of exceptions when deleting a completed action older than the retention period
--- a/classes/ActionScheduler_QueueCleaner.php
+++ b/classes/ActionScheduler_QueueCleaner.php
@@ -4,6 +4,13 @@
  * Class ActionScheduler_QueueCleaner
  */
 class ActionScheduler_QueueCleaner {
+	/**
+	 * The cleaner action hook is scheduled to run daily to initiate cleanup. It also triggers a series of non-recurring
+	 * actions that continue the cleanup in batches until all deletion actions are completed.
+	 *
+	 * @var string
+	 */
+	private const RUN_SCHEDULED_CLEANER_HOOK = 'action_scheduler_run_actions_cleanup_hook';
 
 	/**
 	 * The batch size.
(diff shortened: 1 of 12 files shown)


jorgeatorres (commented):
Hey @kalessil!

Thanks for all your work on this one. I think moving to an async job is definitely a move in the right direction and the fallback you added (for custom cleaners) addresses BC concerns cleanly. 💯.

Only thing I noticed is that there's a (perhaps small?) risk of the cleanup never running if the other tasks (which are normally of higher priority) keep the queue "occupied". Previously, cleanup _always_ ran as part of any queue run, instead of competing with the rest of the queue.

Im not sure how likely this is to happen, but on a site with a very busy queue of tasks of priority <= 10, it could in theory happen. A few options we could consider:

- Schedule the daily cleanup with higher priority (10 as other normal work)
- Add a time-based guard during a normal batch run: if the cleanup action exists and hasn't completed in N hours after its scheduled time, run it synchronously as part of the batch run (similar to how things behaved before this PR).

What do you think? Let me know if I've missed something.

kalessil:
Thanks for the feedback @jorgeatorres!

> Add a time-based guard during a normal batch run: if the cleanup action exists and hasn't completed within N hours of its scheduled time, run it synchronously as part of the batch run (similar to how things behaved before this PR).

I think this is not necessary:
- Ensuring the cleanup is scheduled is backed by A-S already (via scheduled `action_scheduler_run_recurring_actions_schedule_hook` daily task).
- If we demand harder guarantees here, we should probably execute `add_action( 'action_scheduler_init', array( $this, 'schedule_recurring_scheduler_hook' ) );` on plugin update to force the sync (needs verifying on the side-effects)

> Schedule the daily cleanup with higher priority (10 as other normal work)

Tricky one, similar to `action_scheduler_run_recurring_actions_schedule_hook`: there is a risk that if the queue is not draining by 3 AM (it's still a red flag for the store), both tasks will be pushed back further.

How about changing the priority to 0 or 5 for both of them? Both are fast and critical to A-S working correctly and performing well - interruption to queue processing, even in a fully stocked queue, will be minimal and likely not noticeable at 3 AM.

jorgeatorres (commented):
Thanks for making the updates @kalessil!

I spent a bit more time on this and I'm thinking we should limit the continuation/trailing jobs to just one (`unique = true`). I know this is explicitly called out in the current code, but thinking of busy stores with tons of failed old actions, this new cleanup could end up dominating the queue for a while, which isn't ideal (particularly due to the busy nature of the store in the first place).

Notice that `delete_old_actions()` can schedule 2 more continuation jobs, and assuming enough actions remain for cleanup, (> 250 in each run) those will in turn spawn 2 more, and so on.

Assuming the default queue of 25 jobs, cleanup jobs might end up taking all the spots eventually until the cleanup finalizes. Given prior to this PR, A-S never auto-deleted failed actions, it's not unreasonable to think there might be a backlog there.

To replicate the situation, you can artificially insert into the database, say, 10000 failed/completed/canceled actions with a date in the 2000s, for example.

After the initial run of the cleanup job, you can process the queue in discrete steps by running:

```
wp action-scheduler run \
    --hooks=action_scheduler_run_actions_cleanup_hook,action_scheduler_continue_actions_cleanup_hook \
    --batch-size=25 --batches=1
```

After each run you can confirm that the actions are getting deleted *but* also that scheduled cleanup jobs duplicate in size.

Let me know if you want me to share a script to set up things for that.

I've proposed a fix in #1325, which should merge cleanly into this branch, if you think it's reasonable. It makes the trailing jobs `unique = true`, so during the cleanup we at most occupy 1 spot in each queue. I think that should be enough and allows for other jobs to process normally. We could adjust the number of jobs later on if we think there's an opportunity, but for this version maybe we should be conservative as to not impact the ability for a site to process their normal workload. What do you think?

kalessil:
Good finding @jorgeatorres! I'll review/merge your PR today and ping you back.

jorgeatorres (approved):
LGTM. Thank you @kalessil!

PR #1313, opened 2026-03-12
Optimize get_claim_count() query for large stores

mcliwanow:
## Description

Optimizes `get_claim_count()` in `ActionScheduler_DBStore` for stores with large `actionscheduler_actions` tables.

The original query scans the (potentially huge) actions table:

```sql
SELECT COUNT(DISTINCT claim_id)
FROM wp_actionscheduler_actions
WHERE claim_id != 0 AND status IN ('pending', 'in-progress')
```

On a store with ~3M action rows, this took ~28 seconds to return 0.

The new query starts from the small actionscheduler_claims table and uses EXISTS to check the actions table:

```sql
SELECT COUNT(*)
FROM wp_actionscheduler_claims c
WHERE EXISTS (
   SELECT 1 FROM wp_actionscheduler_actions a
   WHERE a.claim_id = c.claim_id
   AND a.status IN ('pending', 'in-progress')
)
```

This returns in ~0.078 seconds (350x faster), because the claims table only holds active claims (typically a handful of rows), and EXISTS stops scanning at the first match.

If the claims-table query fails (e.g. the table doesn't exist yet during migration), the method falls back to the original query shape and logs the error.

Ported from WCCOM

### Test instructions

**Setup**

Run the unit tests to verify correctness:

```bash
./vendor/bin/phpunit tests/phpunit/jobstore/ActionScheduler_DBStore_Test.php \ -c tests/phpunit.xml.dist --filter test_get_claim_count
```

All 7 new test cases should pass (runs as 42 with timezone rotation).

**Manual verification**

**Prerequisites**: A local WooCommerce site. Copy the updated `ActionScheduler_DBStore.php` into `wp-content/plugins/woocommerce/packages/action-scheduler/classes/data-stores/`.

1. Add this to wp-config.php to log queries:

```php
define( 'SAVEQUERIES', true );
```

2. Then add this temporary mu-plugin to `wp-content/mu-plugins/log-claim-count.php`:

```php
<?php
add_action( 'my_test_hook', function( $i ) {
    global $wpdb;

    $store = ActionScheduler::store();
    $count = $store->get_claim_count();

    $claim_query = null;
    foreach ( $wpdb->queries as $q ) {
        if ( str_contains( $q[0], 'actionscheduler_claims' ) && str_contains( $q[0], 'COUNT' ) ) {
            $claim_query = $q;
        }
    }

    error_log( sprintf(
        'During action %d: get_claim_count() = %d | query time: %fs',
        $i,
        $count,
        $claim_query[1] ?? 0
    ) );
} );

 add_action( 'action_scheduler_after_process_queue', function() {
    $store = ActionScheduler::store();
    error_log( sprintf( 'After processing: get_claim_count() = %d', $store->get_claim_count() ) );
} );
```

3. Test actions
  - Schedule test actions via WP-CLI or a snippet:
  ```php
   for ( $i = 0; $i < 5; $i++ ) { as_enqueue_async_action( 'my_test_hook', array( $i ) ); }
   ```
  - Trigger processing — the log should show a non-zero count while actions run, then 0 after completion 
image
  - next processing should just show claims with 0

4. Cleanup
  - Remove the mu-plugin
  - Remove `SAVEQUERIES` from `wp-config.php`

### DIFF
--- a/classes/data-stores/ActionScheduler_DBStore.php
+++ b/classes/data-stores/ActionScheduler_DBStore.php
@@ -1073,10 +1073,34 @@ AND `group_id` = %d
 	public function get_claim_count() {
 		global $wpdb;
 
-		$sql = "SELECT COUNT(DISTINCT claim_id) FROM {$wpdb->actionscheduler_actions} WHERE claim_id != 0 AND status IN ( %s, %s)";
+		$sql = "
+			SELECT COUNT(*)
+			FROM {$wpdb->actionscheduler_claims} c
+			WHERE EXISTS (
+				SELECT 1
+				FROM {$wpdb->actionscheduler_actions} a
+				WHERE a.claim_id = c.claim_id
+				AND a.status IN ( %s, %s )
+			)
+		";
 		$sql = $wpdb->prepare( $sql, array( self::STATUS_PENDING, self::STATUS_RUNNING ) ); // phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared
 
-		return (int) $wpdb->get_var( $sql ); // phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared
+		$claim_count = $wpdb->get_var( $sql ); // phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared
+		if ( null === $claim_count ) {
+			$error = empty( $wpdb->last_error )
+				? _x( 'unknown', 'database error', 'woocommerce' )
+				: $wpdb->last_error;
+
+			// phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_error_log
+			error_log( sprintf( 'Action Scheduler get_claim_count() fast-path query failed, falling back to actions table query. Database error: %s', $error ) );
+
+			// Fall back to the historical query shape if the claims table cannot be used.
+			$fallback_sql = "SELECT COUNT(DISTINCT claim_id) FROM {$wpdb->actionscheduler_actions} WHERE claim_id > 0 AND status IN ( %s, %s)";
+			$fallback_sql = $wpdb->prepare( $fallback_sql, array( self::STATUS_PENDING, self::STATUS_RUNNING ) ); // phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared
+			$claim_count  = $wpdb->get_var( $fallback_sql ); // phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared
+		}
+
+		return (int) $claim_count;
 	}
 
 	/**
--- a/classes/data-stores/ActionScheduler_DBStore.php
+++ b/classes/data-stores/ActionScheduler_DBStore.php
@@ -1085,22 +1085,7 @@ AND `group_id` = %d
 		";
 		$sql = $wpdb->prepare( $sql, array( self::STATUS_PENDING, self::STATUS_RUNNING ) ); // phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared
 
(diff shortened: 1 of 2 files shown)


bor0 (approved):
Approving based on previous WCCOM review and neat results 🎉

mcliwanow:
Hey @jorgeatorres 👋  Please let me know if you're the best person to review this PR. I'd appreciate someone else taking a look at it but I'm also happy to reassign review to someone else. Thanks!

classes/data-stores/ActionScheduler_DBStore.php
```diff
@@ -1073,10 +1073,34 @@ private function db_supports_skip_locked() {
...
+				: $wpdb->last_error;
+
+			// phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_error_log
+			error_log( sprintf( 'Action Scheduler get_claim_count() fast-path query failed, falling back to actions table query. Database error: %s', $error ) );
+
+			// Fall back to the historical query shape if the claims table cannot be used.
+			$fallback_sql = "SELECT COUNT(DISTINCT claim_id) FROM {$wpdb->actionscheduler_actions} WHERE claim_id > 0 AND status IN ( %s, %s)";
+			$fallback_sql = $wpdb->prepare( $fallback_sql, array( self::STATUS_PENDING, self::STATUS_RUNNING ) ); // phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared
+			$claim_count  = $wpdb->get_var( $fallback_sql ); // phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared
+		}
```
jorgeatorres: Can you expand on why the fallback query? I'm not sure if we really need it. If the claims table doesn't exist, we wouldn't be able to claim the actions for running in the first place so unlikely we would end up here anyways.

As for being in the middle of a migration from the posts datastore to the db one, `HybridStore` [actually calls `stake_claim()` on the "primary" datastore](https://github.com/woocommerce/action-scheduler/blob/a95f351/classes/data-stores/ActionScheduler_HybridStore.php#L243-L253) so we would already be in trouble if the db doesn't have the claims table.

I might be missing something, of course, so happy to discuss.

jorgeatorres (commented):
@mcliwanow: thanks for the PR! This looks good to me and the improvements are quite nice. I left some feedback about the fallback query so we can discuss, but that's pretty much it. Thanks again!

classes/data-stores/ActionScheduler_DBStore.php
mcliwanow: I think I was probably overly cautious here (trying too eagerly not to break anything 😅 ). After another review, I do agree that this path isn't needed. I simplified it in https://github.com/woocommerce/action-scheduler/pull/1313/commits/b7c1d5d1f3025e0ea90a101c028d9761b10aa746.

Thanks for pointing this out!

jorgeatorres (approved):
LGTM. Ran a few tests with concurrent batches and also with actions marked completed without releasing the claim to confirm that the `get_claim_count()` matched the expected values. Thanks for the PR @mcliwanow! 🥇 

- - - -

> Copy the updated ActionScheduler_DBStore.php into wp content/plugins/woocommerce/packages/action-scheduler/classes/data-stores/ [...]

💡 For future reference, you can always clone this repo and add it to your test site as a normal WP plugin. While enabled, it'll override the version of A-S bundled in WC, so no need to modify WC itself (and no need to even have WC installed).

mcliwanow:
>  it'll override the version of A-S bundled in WC

Oh nice. That's good to know. Thanks!

Issue #980, opened 2023-09-05
10 x time limit in failure log entries


galbaras:
After setting the queue runner time limit to **600**, failed actions show the message "action marked as failed after **6000** seconds. Unknown error occurred. Check server, PHP and database error logs to diagnose cause."

This is caused by [Line 229 in ActionScheduler_Abstract_QueueRunner.php](https://github.com/woocommerce/action-scheduler/blob/4eb2fa9737a53e4d284dafcf3e0bf428b5f941bc/classes/abstracts/ActionScheduler_Abstract_QueueRunner.php#L229), which says

`$this->cleaner->clean( 10 * $this->get_time_limit() );`

This gives the impression that the action timed out after a much longer time than it likely did and should probably be adjusted to reflect the actual time limit used.

barryhughes:
Hi @galbaras,

It doesn't feel inaccurate to me (in these cases, the action has indeed been marked as failed after 6,000 seconds have elapsed) but perhaps we could tweak the language to make it less ambiguous. For example:

> This action was set to a status of in-progress at least 6000 seconds ago. There has been no further change and this suggests it has stopped running, therefore the status will now be updated to failed.

Alternatively, we could attempt to time the actual action duration and use PHP shutdown functions to handle this in situations where execution times out ... but, that is a larger effort and I'm a little unsure how valuable it would be to most users.

What do you think—and are there scenarios you can share where more accurate figures would be useful?

galbaras:
The cleaner doesn't seem to run after 10x the time. This figure is just hardcoded with no other reference I could find that would explain why. Since I sent the time limit to 600, wouldn't the task have died after 600 seconds already?

Basically, the log should help with troubleshooting as much as possible, and I'm seeing nothing relevant in my error logs, which is totally weird. Increasing the time limit seemed to help, although the log previous quoted "300", so I doubled it, but really multiplied by 20, because of this 10x factor in the code.

barryhughes:
> The cleaner doesn't seem to run after 10x the time.

Right, but it doesn't reflect when the cleaner itself runs: instead it is used to form the cutoff time when the cleaner [queries for stalled actions](https://github.com/woocommerce/action-scheduler/blob/3.6.2/classes/ActionScheduler_QueueCleaner.php#L162-L170).

[...]

PR #1323, opened 2026-06-09
fix: clarify timed-out action log message wording

thisismyurl:
## What

The log message for a timed-out action read:

> action marked as failed after 300 seconds

The value `300` comes from `10 × get_time_limit()` — it's the cleanup window, not the time the action actually ran. With a 30-second PHP time limit, the log implies the action ran for five minutes when in reality it ran for at most 30 seconds. This regularly trips up developers reading logs when diagnosing slow or hung actions.

**Before:**
```
action marked as failed after %s seconds. Unknown error occurred.
```

**After:**
```
action was in-progress for at least %s seconds without completing and has been marked as failed.
```

The new wording makes it clear that `%s` is the minimum elapsed time at which the cleaner *checked*, not a measurement of how long the action actually ran.

## Why not change the value passed?

Changing the multiplier or the method signature would affect the cleanup window logic, which is a separate concern. The wording fix is the minimal, safe change — it corrects the misleading implication without touching any behavior.

Related to #980

---
*Development and testing assisted by AI. All code reviewed and verified manually.*

### DIFF
--- a/classes/abstracts/ActionScheduler_Logger.php
+++ b/classes/abstracts/ActionScheduler_Logger.php
@@ -168,7 +168,7 @@ abstract class ActionScheduler_Logger {
 	 */
 	public function log_timed_out_action( $action_id, $timeout ) {
 		/* translators: %s: amount of time */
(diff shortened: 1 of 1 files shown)


jorgeatorres:
Closing as the commit is included in #1324.

PR #1324, opened 2026-06-09
fix: esc_html() hook name in admin notice markup

thisismyurl:
## What

In `ActionScheduler_ListTable::display_admin_notices()`, the action hook name was concatenated directly into HTML without escaping:

```php
// Before
$action_hook_html = '<strong><code>' . $action->get_hook() . '</code></strong>';

// After
$action_hook_html = '<strong><code>' . esc_html( $action->get_hook() ) . '</code></strong>';
```

Hook names are stored raw in the database. A hook name containing `<`, `>`, `&`, or quotes would render unescaped in the WooCommerce → Status → Scheduled Actions admin screen, producing broken markup.

## Why `wp_kses_post()` isn't sufficient

The composed string is later passed through `wp_kses_post()`, which strips *disallowed* tags — but it does **not** HTML-encode text content inside *allowed* tags like `<strong>` and `<code>`. A hook name of `my_hook<script>` would have its `<script>` stripped, but `my_hook&bad=<value>` would still render with a raw `&` and `<`. `esc_html()` is the correct tool for encoding text content in an HTML context.

---
*Development and testing assisted by AI. All code reviewed and verified manually.*

### DIFF
--- a/classes/ActionScheduler_ListTable.php
+++ b/classes/ActionScheduler_ListTable.php
@@ -413,7 +413,7 @@ class ActionScheduler_ListTable extends ActionScheduler_Abstract_ListTable {
 			delete_transient( 'action_scheduler_admin_notice' );
 
 			$action           = $this->store->fetch_action( $notification['action_id'] );
-			$action_hook_html = '<strong><code>' . $action->get_hook() . '</code></strong>';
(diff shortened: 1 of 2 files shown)


jorgeatorres (approved):
LGTM. Thank you @thisismyurl!

PR #1325, opened 2026-06-09
Turn cleanup continuations (trailing end) into an unique job

jorgeatorres:


kalessil:
From a reliability PoV, this change is brilliant!

kalessil (approved):

PR #1329, opened 2026-06-12
[Performance] Cache resolved group IDs to reduce the number of SQL queries

kalessil:
Lowers the number of SQL queries and increases throughput during bulk tasks processing and scheduling.

Testing:
- Automated: CI-checks shall pass
- Smoke test: build fresh zip and install it on the test site, perform products import and verify related actions scheduler jobs has been added to the queue.

### DIFF
--- a/classes/data-stores/ActionScheduler_DBStore.php
+++ b/classes/data-stores/ActionScheduler_DBStore.php
@@ -10,6 +10,16 @@
 class ActionScheduler_DBStore extends ActionScheduler_Store {
 
 	/**
+	 * WordPress object cache group for resolved group IDs.
+	 */
+	const GROUP_IDS_CACHE_GROUP = 'action_scheduler_groups';
+
+	/**
+	 * Cache key used for the default (empty-slug) group, since wp_cache_* rejects empty string keys.
+	 */
+	const GROUP_IDS_DEFAULT_CACHE_KEY = 'group:default';
+
+	/**
 	 * Used to share information about the before_date property of claims internally.
 	 *
 	 * This is used in preference to passing the same information as a method param
@@ -53,6 +63,23 @@ class ActionScheduler_DBStore extends ActionScheduler_Store {
 		$table_maker = new ActionScheduler_StoreSchema();
 		$table_maker->init();
 		$table_maker->register_tables();
+	}
+
+	/**
+	 * Flush all store caches.
+	 */
+	public function flush_caches() {
+		wp_cache_flush_group( self::GROUP_IDS_CACHE_GROUP );
+	}
+
+	/**
+	 * Translate a group slug into a valid (non-empty) object cache key.
+	 *
+	 * @param string $slug Group slug.
+	 * @return string
+	 */
+	private function get_group_cache_key( string $slug ): string {
+		return '' !== $slug ? $slug : self::GROUP_IDS_DEFAULT_CACHE_KEY;
 	}
 
 	/**
@@ -264,17 +291,17 @@ AND args = %s
 		}
 		return $this->hash_args( $encoded );
 	}
+
 	/**
 	 * Get a group's ID based on its name/slug.
 	 *
 	 * @param string|array $slugs                The string name of a group, or names for several groups.
 	 * @param bool         $create_if_not_exists Whether to create the group if it does not already exist. Default, true - create the group.
 	 *
-	 * @return array The group IDs, if they exist or were successfully created. May be empty.
+	 * @return int[] The group IDs, if they exist or were successfully created. May be empty.
 	 */
 	protected function get_group_ids( $slugs, $create_if_not_exists = true ) {
-		$slugs     = (array) $slugs;
-		$group_ids = array();
+		$slugs = (array) $slugs;
 
 		if ( empty( $slugs ) ) {
 			return array();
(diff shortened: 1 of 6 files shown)


classes/data-stores/ActionScheduler_DBStore.php
```diff
@@ -264,17 +278,17 @@ protected function get_args_for_query( $args ) {
...
 	 * @param string|array $slugs                The string name of a group, or names for several groups.
 	 * @param bool         $create_if_not_exists Whether to create the group if it does not already exist. Default, true - create the group.
 	 *
-	 * @return array The group IDs, if they exist or were successfully created. May be empty.
+	 * @return int[] The group IDs, if they exist or were successfully created. May be empty.
 	 */
 	protected function get_group_ids( $slugs, $create_if_not_exists = true ) {
-		$slugs     = (array) $slugs;
-		$group_ids = array();
+		$slugs = array_filter( (array) $slugs );
```
barryhughes: Previously, we didn't pass the `$slugs` through `array_filter()` and so the default group `(string) ''` would not be filtered out.

Imagine a new action is being saved: assuming `$create_if_not_exists === true` then (again, I'm talking about the historic behavior) we would actually create a new group for the empty string. If you have a long-lived local db you can probably see this in the `wp_actionscheduler_groups` table (obviously, the actual group ID will likely be different):

| group_id | slug |
| ---      | ---  |
| 123      |      |

By using `array_filter()` we stop that from happening, and I quite like that, it feels almost like an accidental bug fix (because, why would we want a group ID in this situation?) *but,* I think we should pause and make sure this doesn't introduce any problems.

Specifically, I'm thinking that for un-grouped actions we will end up with a split pre and post the introduction of this PR:

| action_id | hook    | group_id | note                              |
| ---       | ---     | ---      | ---                               |
| 1         | generic | 123      | Ungrouped, created before this PR |
| 2         | generic |          | Ungrouped, created after this PR  |

My hunch is that this will not present any problems, because 99.9% of the time both will be picked up by the general purpose queue runner ... but I still wanted to flag it in case there might be some other complication.

classes/data-stores/ActionScheduler_DBStore.php
```diff
@@ -19,6 +19,13 @@ class ActionScheduler_DBStore extends ActionScheduler_Store {
 	 */
 	private $claim_before_date = null;
 
+	/**
+	 * Slug to id resolution cache for groups.
+	 *
+	 * @var array<string,int>
+	 */
+	private $group_id_cache = array();
```
barryhughes: We can generally assume that group IDs are constant. That is, once a group is created, it will not (under normal circumstances) be deleted, not even during queue clean-up.

This makes me wonder if we could gain further benefits by using [wp_cache_set()](https://developer.wordpress.org/reference/functions/wp_cache_set/)|[wp_cache_get()](https://developer.wordpress.org/reference/functions/wp_cache_get/) ... so that, if a persistent caching layer is in place, we can drop our group query altogether in some cases (whereas, currently, they still happen at least once per request).

I think this would be relatively safe, especially with a fairly conservative cache expiry like `10 * MINUTE_IN_SECONDS`. That said, I'm partly asking because I'd be interested to learn if you already discarded this idea due to possible problems (and I think what we have already represents a step forward).

barryhughes (commented):
Left a few comments, but this looks nice!

Also, in relation to the slightly modified behavior when an action without a group is persisted, I wonder if we should cement that/describe it via a further unit test?

classes/data-stores/ActionScheduler_DBStore.php
kalessil: I dismissed the idea because of the concern about concurrent updates (caching a single cache entry with all resolved slugs).

But as some time passed, I think we can make it happen: one cache entry per slug will be reliable and feasible.

classes/data-stores/ActionScheduler_DBStore.php
kalessil: Sounds reasonable, I'll make the necessary adjustments.

kalessil: Reverted: `array_filter` being dropped.

classes/data-stores/ActionScheduler_DBStore.php
kalessil: Addressed: new cache group, one cache entry per slug.

kalessil:
@barryhughes: Ready for the next review round (force-pushing to clean commits history)

barryhughes (approved):
👍🏼 LGTM!

🔬 Ran through a set of tests with and without persistent caching (memcached in my case), and triggered queue runners where no groups were specified, where a single group was specified (initially uncached, then later cached), and where *multiple* groups were specified for exclusion (both uncached, one cached, both cached). Everything looked good.

🤔 Not a blocker, because setting/getting should be a very fast operation, but I'm also cognisant it often means an additional network request (in my case, to the memcached container). That makes me wonder if it would be better to have a single cached array vs individual cache keys within the same group ... but, the probable answer to this is _'it depends'_ and what we have is already have is a nice improvement.

📜 As in another recent PR, I think it would make sense (in terms of how our team generally works) to add a changelog entry as part of this PR. Again, not a blocker.

kalessil:
> Not a blocker, because setting/getting should be a very fast operation, but I'm also cognisant it often means an additional network request (in my case, to the memcached container). That makes me wonder if it would be better to have a single cached array vs individual cache keys within the same group ... but, the probable answer to this is 'it depends' and what we have is already have is a nice improvement.

As with any optimization work, there is a trade-off, which is okay if the changes make a net positive impact.
Here are two dimensions:
- we shift some of the DB load (might need to load data into memory) to the caching service (likely keep everything in memory): that's a fair balancing act, as the DB can spend the CPU cycles on processing actions; network usage is unlikely to see noticeable changes
- Group IDs in a single entry means potential concurrency issues, while for single entries we outsource concurrency handling to the caching service;  less maintenance for us

kalessil:
Added a new changelog entry in 08f5e7932a43336278e25c156a459c957295fa51.
Will merge once CI passes.

PR #1339, opened 2026-06-24
Consolidate dev scripts around Composer and other tooling fixes

jorgeatorres:
While working on the recent 4.0.0 release I noticed a few things with our scripts: particularly, that we were using Grunt just to be able to run something that could be a PHPCS sniff and as a wrapper for PHPCS itself.

This PR does a few things around tooling:

- Removes Grunt and replaces the old `grunt checktextdomain` with the `WordPress.WP.I18n` sniff.
- Removes Husky and pre-commit hook: this dependency was super outdated and was only activated by `npm install` which added an unnecessary dev dependency to what's mostly a Composer package. To replace that I added a few PHPCS commands and a linting CI workflow to enforce standards on PRs.
- Updates PHPCS sniffs to the latest versions.

A summary of new commands:

| Command | Scope |
|---|---|
| `composer phpcs` / `composer phpcbf` | Whole tree (check / fix) |
| `composer lint` | Unstaged changes |
| `composer lint-staged` | Staged changes |
| `composer lint-branch` | Changes vs base branch (same as CI) |

Now, the dev setup is just `composer install`. The `npm build` scripts remains unchanged as it's needed by woorelease.

### Testing instructions

1. Run `composer install` first so the dev tooling is present.
1. Confirm you can run PHPCS across the whole tree via Composer:
   ```sh
   composer phpcs                      # runs, scans the tree
   ```
1. Confirm the version of the WC sniffs in use:
   ```sh
   composer show woocommerce/woocommerce-sniffs
   composer show wp-coding-standards/wpcs
   ```
1. Confirm the WP i18n sniff catches wrong text domains (previously checked via Grunt):
   1. Create a test file `td_test.php` with the following contents:
      ```php
      <?php
      __( "hi", "wrong-domain" );
      ```
   1. Run `composer phpcs td_test.php`.
   1. Confirm it fails because of the `WordPress.WP.I18n.TextDomainMismatch` rule.
1. Test the various PHPCS linting scripts:
   1. Edit `classes/ActionScheduler_Versions.php` and add a new function `f()` like this:
      ```php
      function f() {
            $x = 1 ;
      }
      ```

      Note the extra space before the semicolon, though that function alone is going to trigger other violations as well.
   1. Run `composer lint` and note the violations are flagged.
   1. Stage the file via `git add classes/ActionScheduler_Versions.php` and confirm that `composer lint-staged`  flags the violations.
   1. Commit the change with `git commit` and run `composer lint-branch`. Confirm it flags the violation (on the changed lines).
1. Testing the new CI workflow is not really straightforward until it's merged into `trunk`, but you can see that it mostly just calls `composer lint-branch`, which is covered above and that it also runs on this branch.


Konamiman:
I think I might be missing some additional configuration:

> 3. Confirm the version of the WC sniffs in use:
> composer show woocommerce/woocommerce-sniffs wp-coding-standards/wpcs

This is throwing the following:

```
In VersionParser.php line 526:

Could not parse version constraint wp-coding-standards/wpcs: Invalid version string "wp-coding-standards/wpcs"
```

> 4. Create a test file td_test.php with the following contents (...)
> Run composer phpcs td_test.php.

This throws "ERROR: Referenced sniff "WooCommerce-Core" does not exist".

Also step 2, running `composer phpcs`, reports a ton of errors, not sure if that's expected.

Konamiman (approved):
Ok, with the updated testing instructions I can confirm that step 3 returns sniffs version 1.0.1 and wpcs version 3.3.0, and step 4 works. Approving.

PR #1342, opened 2026-07-08
[Performance] Use the WordPress caching layer for the option lock component.

kalessil:
Introduces the following changes to the `\ActionScheduler_OptionLock` component:

- Capture current timestamps once, where applicable, rather than using in-place calls (reliability).
- Use the WordPress cache for active locks to reduce SQL queries and decrease asynchronous dispatch latency during shutdown (micro-optimization).

Manual testing:
- Build the plugin and install it on a testing site as a standalone plugin
- Add the snippet below (AI-slop, but works), e.g., as a mu-plugin or to another place where it'll be executed
- Navigate to the admin area (with and without persistent object cache being active)
- Refresh the page to monitor and verify the lock acquisition

```php
add_action( 'admin_notices', function () {                                                                                                                                                                   
        $lock      = ActionScheduler::lock();                                                                                                                                                                  
        $lock_type = 'async-request-runner';                                                                                                                                                                   
        $is_locked  = $lock->is_locked( $lock_type );                                                                                                                                                          
        $expiration = $lock->get_expiration( $lock_type );
                                                                                                                                                                                                               
        $status = $is_locked                                                                                                                                                                                 
                ? sprintf(                                                                                                                                                                                     
                        '<strong>LOCKED</strong> — expires %s (in %ds)',                                                                                                                                     
                        esc_html( date( 'H:i:s', $expiration ) ),                                                                                                                                              
                        max( 0, $expiration - time() )                                                                                                                                                         
                )                                                                                                                                                                                              
                : '<strong>NOT LOCKED</strong>';                                                                                                                                                               
                                                                                                                                                                                                             
        printf(                                                                                                                                                                                                
                '<div class="notice notice-%s"><p>ActionScheduler lock <code>%s</code>: %s</p></div>',
                $is_locked ? 'warning' : 'info',                                                                                                                                                               
                esc_html( $lock_type ),                                                                                                                                                                      
                $status                                                                                                                                                                                        
        );                                                                                                                                                                                                     
  } );

```

### DIFF
--- a/classes/ActionScheduler_OptionLock.php
+++ b/classes/ActionScheduler_OptionLock.php
@@ -54,7 +54,7 @@ class ActionScheduler_OptionLock extends ActionScheduler_Lock {
 			return $inserted;
 		}
 
-		if ( $this->get_expiration_from( $existing_lock_value ) > $now ) {
+		if ( $this->get_expiration_from( $existing_lock_value ) >= $now ) {
 			return false;
 		}
 
--- a/classes/ActionScheduler_OptionLock.php
+++ b/classes/ActionScheduler_OptionLock.php
@@ -27,13 +27,14 @@ class ActionScheduler_OptionLock extends ActionScheduler_Lock {
 	public function set( $lock_type ) {
 		global $wpdb;
 
+		$now                 = time();
 		$lock_key            = $this->get_key( $lock_type );
-		$existing_lock_value = $this->get_existing_lock( $lock_type );
-		$new_lock_value      = $this->new_lock_value( $lock_type );
+		$existing_lock_value = $this->get_existing_lock( $lock_type, $now );
+		$new_lock_value      = $this->new_lock_value( $lock_type, $now );
 
 		// The lock may not exist yet, or may have been deleted.
 		if ( null === $existing_lock_value ) {
-			return (bool) $wpdb->insert(
+			$inserted = (bool) $wpdb->insert(
 				$wpdb->options,
 				array(
 					'option_name'  => $lock_key,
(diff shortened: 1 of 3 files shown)


classes/ActionScheduler_OptionLock.php
```diff
@@ -41,21 +42,37 @@ public function set( $lock_type ) {
 					'autoload'     => 'no',
 				)
 			);
+			$ttl = $this->get_expiration_from( $new_lock_value ) - $now;
+			if ( $inserted && $ttl > 0 ) {
+				wp_cache_set( $lock_key, $new_lock_value, 'action_scheduler_locks', $ttl );
+			}
+			return $inserted;
```
barryhughes: To make this work I think we need to update line 37, above:

```diff
-return (bool) $wpdb->insert(
+$inserted = (bool) $wpdb->insert(
```

Otherwise we don't reach this code.

barryhughes (commented):
This looks like a nice approach to me, and I think will add a small improvement even when there is no persistent caching layer (because `ActionScheduler_QueueRunner::maybe_dispatch_async_request()` effectively reads the lock twice in the same request).

I realize it's a draft, so this is likely planned, but definitely worth additional tests.

classes/ActionScheduler_OptionLock.php
kalessil: Yep, some merge conflict leftovers - I'll take care of those.

classes/ActionScheduler_OptionLock.php
```diff
@@ -27,35 +27,60 @@ class ActionScheduler_OptionLock extends ActionScheduler_Lock {
...
+				if ( $ttl > 0 ) {
+					wp_cache_set( $lock_key, $new_lock_value, 'action_scheduler_locks', $ttl );
+				}
+			}
+
+			return $inserted;
 		}
 
-		if ( $this->get_expiration_from( $existing_lock_value ) >= time() ) {
+		if ( $this->get_expiration_from( $existing_lock_value ) > $now ) {
```
barryhughes: I think that swapping `>=` for `>` is *probably* fine, but does it ever so slightly increase the chance of 'thrashing' (that may not be quite the right word) in the sense that we may try to acquire the lock (when `$existing_lock_expiration === $now`) when perhaps on occasion we cannot?

Not a blocker, as even if this does happen occasionally the effect would be very small and wouldn't cause a functional problem.

barryhughes (approved):
Nice!

<div align="center">Shows admin notices generated by Action Scheduler and by the testing instructions snippet</div>

I left a non-blocking comment but this looks good and tests well. I supported the testing instructions with a further snippet to create a small stream of past-due actions:

```php
add_action( 'action_scheduler_init', function () {
	rand( 0, 1 ) ? as_enqueue_async_action( 'test-pr-1342' ) : 0;
} );
```

These continued to be vacuumed up at the expected cadence (albeit there wasn't much contention in my local env, I didn't see cases where lock acquisition was notably delayed).

classes/ActionScheduler_OptionLock.php
kalessil: Good catch; it was an unintended one: restoring in 5c6fba79259992b3ca360e66517003d29e1dbf6a

PR #1344, opened 2026-07-11
Schedule deserialization hardening

### LINKED ISSUE #1318: Insecure Deserialization
Action Scheduler library contains an insecure deserialization vulnerability in ActionScheduler_DBStore::fetch_action() method. The
  unserialize() function is called without the allowed_classes parameter on user-controllable data from the database, allowing
  arbitrary PHP object instantiation and remote code execution via PHP Object Injection (POI) when chained with SQL injection or
  database write access.

How to reproduce:

  1. Access MySQL database with credentials (requires SQL injection or compromised admin account)
  2. Create malicious PHP object:
  class ExploitGadget {
      public $file_path = '/var/www/html/wp-content/exploit-test.txt';
      public $content = 'EXPLOITED';
      public function __destruct() {
          file_put_contents($this->file_path, $this->content);
      }
  }
  $payload = serialize(new ExploitGadget());

  3. Insert payload into Action Scheduler table:
  INSERT INTO wp_actionscheduler_actions (hook, status, scheduled_date_gmt, schedule, args, group_id)
  VALUES ('test_hook', 'pending', NOW(), '[SERIALIZED_PAYLOAD]', '[]', 1);

  4. Trigger deserialization via Action Scheduler cron or manual call to ActionScheduler_Store::instance()->fetch_action($action_id)
  5. Verify code execution: Check if /var/www/html/wp-content/exploit-test.txt was created

  Vulnerable code location:
  classes/data-stores/ActionScheduler_DBStore.php:397
  $schedule = unserialize( $data->schedule ); // No allowed_classes parameter

barryhughes:
Both of the default action stores persist schedule information as serialized PHP blobs. However, there are currently no protections during the deserialization process to prevent PHP object injection attacks.

- Fixes https://github.com/woocommerce/action-scheduler/issues/1318
- Builds on and potentially replaces https://github.com/woocommerce/action-scheduler/pull/1341 (but I've kept that alive in case we want to rework from that foundation)

The core concept here (originally introduced by #1341, thank you @vladolaru) is to handle deserialization in one or more passes. High-level (slightly simplified) summary of the algorithm in this PR:

- First pass: unserialize() with an `allowed_classes` list covering the library's own schedule classes and the small set of support classes they legitimately nest (the `CronExpression` family, and the built-in date/time value objects). **Normal data hydrates fully in this pass** — if the outermost object is a schedule and nothing decoded to `__PHP_Incomplete_Class`, we return it with no further work.
- If unexpected classes remain: the outermost object must itself be a schedule (a top-level non-schedule — e.g. an injected gadget — is rejected here, before we walk anything). Each remaining `__PHP_Incomplete_Class` is then vetted: acceptable if it's on the (filterable) support allow-list or implements `ActionScheduler_Schedule`. If all pass, we `unserialize()` once more restricted to the expanded set and return the result; if any fails, we return false (which the caller treats as corrupt).

The idea is to minimize any additional overhead when we are dealing with perfectly normal and expected schedule data, while still filtering out potentially dangerous class references.

---

### Testing

The test suite should provide reasonable coverage, but we can also perform some manual testing and exploration. For instance, you can try creating conventional scheduled actions of various sorts via WP CLI:

```sh
wp eval "as_schedule_cron_action( time(), '* */2 * * *', 'testcronaction' );"
```

Then, visit **Tools ‣ Scheduled Actions**. You should see the created actions in the queue of pending actions and, if you run them manually, they should successfully complete or fail as expected.

> [!TIP]
> When is it acceptable for a test action like `testcronaction`, created above, to fail? When we haven't registered a callback. You can workaround this by adding something like `add_action( 'testcronaction', fn () => 0 );` to a mu-plugin, or just look at the failed action log and verify it says something like, _'Scheduled action for testcronaction will not be executed as no callbacks are registered.'_ This is fine and expected.

Second, we can simulate a perfectly valid custom schedule that happens to contain an instance of an unknown class:

```php
class Problematic_Custom_Schedule implements ActionScheduler_Schedule {
	public $data;

	public function __construct() {
		$this->data = new Custom_Entity;
	}

	public function next(?DateTime $after = null) {
		return $after === null ? date_create( '+1 day' ) : $after->modify( '+1 day' );
	}

	public function is_recurring() {
		return false;
	}

	public function get_date() {
		return date_create();
	}
}

class Custom_Entity {
	public $data = 123;
}

ActionScheduler::store()->save_action( 
	new ActionScheduler_Action( 'testaction', [], new Problematic_Custom_Schedule )
);
```

Run the above with `wp eval-file /path/to/script.php`, then head over to **Tools ‣ Scheduled Actions**. Depending on how quickly you get there it will probably be sitting in either the **past-due** queue or else will have moved to **cancelled** (which is how these particularl cases are handled).

Shows an action cancelled because of corruption

Feel free to explore other permutations 🙂

---

### Notes/questions

- The above example of `Problematic_Custom_Schedule` illustrates how some third party or custom schedules, that until now were considered valid, will be handled. This is potentially problematic (by default, cancelled actions cannot be re-run and will be deleted after one month); so we need to decide if that risk is acceptable.
- Note here that I did not find any cases like this when scanning through a body of known WooCommerce extensions. Of course, that doesn't cover everything out there (still, it's a useful indicator).
- It may even be desirable to introduce a new status like `suspended` (right now I'm less interested in what we name it and more the concept as a whole) which would then give admins a chance to review and force run the actions. An existing status like `failed` might also work for this, but we would need to add the ability to forcibly re-run.
- Further, we provide a new `action_scheduler_allowed_nested_schedule_classes` filter to make it easier for custom schedule implementations to expand the allow-list.
- At site/project level, new hook `action_scheduler_enforce_schedule_allowed_classes` essentially allows for these new mechanics to be turned off (while still making it possible to observe any problems). Of course, this also requires the ability to add custom code. We again need to ask if that is acceptable by itself, or should be supported with some sort of settings UI.
- Last, it seems fine to ship this as part of a 4.1.0 release alongside other fixes already merged to `trunk`, but we could also separate it out.

### DIFF
--- a/classes/ActionScheduler_ScheduleDeserializer.php
+++ b/classes/ActionScheduler_ScheduleDeserializer.php
@@ -214,6 +214,10 @@ class ActionScheduler_ScheduleDeserializer {
 	 * @return object|false
 	 */
 	protected static function handle_rejection( $data, $offending_class, $top_class, array $nested_classes ) {
+		// Resolve enforcement once so the reported value and the branch taken cannot disagree, and the
+		// action_scheduler_enforce_schedule_allowed_classes filter runs a single time.
+		$enforced = self::is_enforced();
+
 		/**
 		 * Fires when a stored schedule blob references a class Action Scheduler did not expect.
 		 *
@@ -225,9 +229,9 @@ class ActionScheduler_ScheduleDeserializer {
 		 * @param bool     $enforced        Whether the blob was rejected (true) or allowed through
 		 *                                  in shadow mode (false).
 		 */
-		do_action( 'action_scheduler_unexpected_schedule_class', $offending_class, $top_class, $nested_classes, self::is_enforced() );
+		do_action( 'action_scheduler_unexpected_schedule_class', $offending_class, $top_class, $nested_classes, $enforced );
 
-		if ( self::is_enforced() ) {
+		if ( $enforced ) {
 			return false;
 		}
 
--- a/classes/ActionScheduler_ScheduleDeserializer.php
+++ b/classes/ActionScheduler_ScheduleDeserializer.php
@@ -1,41 +1,15 @@
 <?php
-
 /**
- * Safely turns a stored, serialized schedule blob back into a schedule object.
+ * Used to safely deserialize schedule information when retrieving a stored scheduled action.
  *
- * Scheduled actions persist their schedule as a native PHP-serialized blob. Reading that blob
- * back with a bare unserialize() lets an attacker who can influence the stored bytes have PHP
- * instantiate arbitrary classes, running their __wakeup()/__destruct() magic methods (PHP object
- * injection). @see https://github.com/woocommerce/action-scheduler/issues/1318
+ * This helps to guard against deserialization attacks, while maintaining backwards compatibility
+ * with older versions of Action Scheduler.
  *
- * This class moves Action Scheduler's existing "is this really an ActionScheduler_Schedule?" check
- * to *before* any class is instantiated, using a two-phase deserialization:
- *
- *   1. Parse the blob with `allowed_classes => false`, which instantiates nothing — every object
- *      in the payload becomes a harmless __PHP_Incomplete_Class placeholder. No constructor,
- *      __wakeup(), __destruct() or autoloading runs.
- *   2. Inspect the class names actually present. Only if the outermost class implements
- *      ActionScheduler_Schedule, and every nested class is on a small vetted allow-list, do we
- *      re-run unserialize() a second time restricted to exactly that verified set of classes.
- *
- * The allow-list is derived, not hard-coded: any loaded schedule class (including third-party
- * ones) is trusted automatically because it implements ActionScheduler_Schedule, so extenders need
- * to do nothing. A blob referencing anything else is rejected the same way a corrupt blob already
- * is today, or — in shadow mode — allowed through untouched while a warning is surfaced.
- *
- * As an optimization, deserialization first attempts a single restricted parse against the
- * statically-known-safe set (Action Scheduler's own schedule classes plus the vetted support
- * classes). When a blob references only those — the overwhelmingly common case — it is fully and
- * safely hydrated in one pass and the two-phase walk above is skipped. Restricting instantiation to
- * that set means nothing outside it can run, so if the result is a schedule with no placeholder left
- * behind, every class named in the blob was already trusted. The inert two-phase path is used only
- * for blobs that reference something outside the fast set: a third-party schedule class, a filtered
- * support class, or a tampered/gadget blob.
- *
- * @since 3.10.0
+ * @internal This implementation is subject to change without notice or back-compat measures.
+ * @see      https://github.com/woocommerce/action-scheduler/issues/1318
+ * @since    4.1.0
  */
 class ActionScheduler_ScheduleDeserializer {
-
 	/**
 	 * Maximum object-graph depth we will walk while vetting a blob.
 	 *
@@ -44,18 +18,15 @@ class ActionScheduler_ScheduleDeserializer {
 	 *
 	 * @var int
 	 */
-	const MAX_GRAPH_DEPTH = 100;
+	private const MAX_GRAPH_DEPTH = 100;
 
 	/**
-	 * Action Scheduler's own concrete schedule classes.
-	 *
-	 * Used to build the fast-path allow-list. This is a fixed, audited set of first-party classes; it
-	 * is deliberately not filterable. Third-party schedule classes are handled by the two-phase path,
-	 * which trusts any class implementing ActionScheduler_Schedule without needing to be listed here.
+	 * Action Scheduler's own concrete schedule classes. Along with any class that implements
+	 * ActionScheduler_Schedule, these are always trusted.
 	 *
 	 * @var string[]
 	 */
-	const KNOWN_SCHEDULE_CLASSES = array(
+	private const KNOWN_SCHEDULE_CLASSES = array(
 		ActionScheduler_SimpleSchedule::class,
 		ActionScheduler_IntervalSchedule::class,
 		ActionScheduler_CronSchedule::class,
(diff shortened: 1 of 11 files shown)


classes/data-stores/ActionScheduler_wpPostStore.php
```diff
@@ -238,7 +238,7 @@ protected function make_action_from_post( $post ) {
 		$args = json_decode( $post->post_content, true );
 		$this->validate_args( $args, $post->ID );
 
-		$schedule = get_post_meta( $post->ID, self::SCHEDULE_META_KEY, true );
+		$schedule = $this->get_schedule_from_post_meta( $post->ID );
```
barryhughes, line 241: ✍🏼 This means a slight performance hit. I think that's acceptable, as this is not the default store and will sit unused in most cases.

kalessil: Two places to verify additionally:
- `classes/data-stores/ActionScheduler_wpPostStore.php:167`: `update_post_meta( $post_id, self::SCHEDULE_META_KEY, $schedule )`
- `classes/data-stores/ActionScheduler_wpPostStore.php:241`: `$schedule = get_post_meta( $post->ID, self::SCHEDULE_META_KEY, true )`

Those also use the WordPress meta API and are susceptible to uncontrolled deserialization when WordPress primes the meta cache.

classes/ActionScheduler_ScheduleDeserializer.php
```diff
@@ -0,0 +1,412 @@
...
+	/**
+	 * Perform the two-pass deserialization for this instance's already-validated blob.
+	 *
+	 * @return ActionScheduler_Schedule|false
+	 */
+	private function deserialize() {
+		// Initial attempt to unserialize, specifying all trusted classes: for default actions using
+		// supplied schedule classes, this should succeed without any problems. Otherwise, the result
+		// may contain one or more __PHP_Incomplete_Class references that we will need to examine.
+		$all_trusted_classes = array_merge( $this->allowed_scheduler_classes, $this->allowed_nested_classes );
```
kalessil, line 172: There is potential to introduce request-level caching for accepted and rejected classes (an associative array with classes as keys, and using isset on those will be instant in terms of performance and less of a headache for deduplication). This will minimize the performance penalty of hardening security here.

classes/ActionScheduler_ScheduleDeserializer.php
```diff
@@ -0,0 +1,412 @@
...
+		 * The result of this filter is not memoized: this is deliberate, so that it can be changed
+		 * if the context (including the current blog, in a multisite network) alters during the
+		 * same request.
+		 *
+		 * @since 4.1.0
+		 *
+		 * @param string[] $default List of allowed nested class names.
+		 */
+		$allowed_nested = (array) apply_filters( 'action_scheduler_allowed_nested_schedule_classes', self::DEFAULT_NESTED_CLASSES );
+		$deserializer   = new self( self::KNOWN_SCHEDULE_CLASSES, $allowed_nested );
```
kalessil, line 125: There is potential to use the singleton pattern here - better for performance.

classes/ActionScheduler_ScheduleDeserializer.php
```diff
@@ -0,0 +1,412 @@
...
+	 *
+	 * Allowed if the class is loaded and implements ActionScheduler_Schedule. A class that is not
+	 * loaded cannot be validated (and, prior to this change, already produced a broken action), so
+	 * rejecting it here does not regress any case that previously worked.
+	 *
+	 * @param string $class_name Class name discovered in the blob.
+	 * @return bool
+	 */
+	private function is_schedule_implementation( string $class_name ): bool {
+		if ( '' === $class_name || ! class_exists( $class_name ) ) {
```
kalessil, line 328: Nit: I'd add a second parameter as `false`:
- Line ~328 — is_schedule_implementation(): class_exists( $class_name )                                                                                                                                      - Line ~332 — is_schedule_implementation(): class_implements( $class_name )
- Line ~359 — is_allowed_nested_class(): class_exists( $class_name )                                                                                                                                         
- Line ~360 — is_allowed_nested_class(): class_implements( $class_name )

Effectively, it's sealing the auto-loading attack channel once and for all (I know we have a prior array check, but extra hardening, if feasible, will not hurt).

This requires constructor patching to keep the hierarchy investigation working (it'll load the class definition):
```php
$this->allowed_nested_classes = array_filter( $allowed_nested_classes, 'class_exists' );
```

One thing to investigate, though, is when a custom class is not directly extending A-S classes (adds multiple inheritance layers).

kalessil (commented):
LGTM overall. Leaving some feedback in the comments below.

classes/ActionScheduler_ScheduleDeserializer.php
barryhughes: Here also replying to [this earlier comment](https://github.com/woocommerce/action-scheduler/pull/1344#discussion_r3584937284), I'm a *little* reluctant to implement these ideas (but may be overthinking it, so may as well talk it out) because of scenarios like this:

```php
foreach ( $blog_ids as $blog_id ) {
	switch_to_blog( $blog_id );

	if ( ! as_next_scheduled_action( 'custom_job' ) ) {
		warn_job_not_scheduled();
	}
}
```

Obviously the above snippet is a bit contrived, but I just mean to illustrate that it will (potentially) lead to schedule deserialization taking place multiple times within the same request, but across different and changing contexts (for which the range of acceptable/unacceptable classes may also change).

For example, instances of `Plugin_A_Schedule_Data` may be allowed on the first iteration, when we are inside Blog 1, but not on the next when we are inside Blog 2 and Plugin A is not even active. And/or, there may be plugin classes that are acceptable within Blog 2 but which were unknown when this first runs, within Blog 1.

This probably isn't very common, and I can think of supporting mechanics to work around it, but am basically feeling a little wary.

kalessil: Given your prior work on A-S, I'd trust your gut feeling.

"Just" drop in a comment there clarifying that the new instance is by design to accommodate context switching (as multisite and more) - so future people have this context and don't challenge this decision because of a lack of context.

classes/ActionScheduler_ScheduleDeserializer.php
kalessil: With more context from https://github.com/woocommerce/action-scheduler/pull/1344#discussion_r3585014778, this one comment is obsolete.

classes/ActionScheduler_ScheduleDeserializer.php
barryhughes: Re multiple inheritance layers, I think this is fine:

```php
class Baz implements ActionScheduler_Schedule {
	public function next( ?DateTime $after = null ) {}
	public function is_recurring() {}
}

class Bar extends Baz {}
class Foo extends Bar {}

var_dump( class_implements( Foo::class ) );
# array(1) {
# 	["ActionScheduler_Schedule"]=>
#   string(24) "ActionScheduler_Schedule"
# }
```

(...but let me know if I've misunderstood.)

barryhughes: > I'd add a second parameter as `false`:

The thing is that the current mechanism relies on this being `true`. 

This does indeed mean there is potential for autoloader side-effects (ie, when a file contains both a class definition and other top-level code, then the top-level code will be executed), but that's one of the trade-offs we're reckoning with here.

If `false`, then we will fail to load legitimate (third party) schedule implementations that rely on an autoloader for discovery.

I'm not closed to the idea, but it increases the risk of breakages (in the sense that up-until-now perfectly valid scheduled actions would risk being cancelled) in production.

classes/data-stores/ActionScheduler_wpPostStore.php
barryhughes: For the first of those, this leads to serialization but not deserialization, so far as I can tell. The second is the very instance we are updating (and replacing with our own mechanics).

classes/ActionScheduler_ScheduleDeserializer.php
barryhughes: That's a good call. There is an existing comment alluding to this, specifically in the sense of multisite network context changes (in the docblock for the `action_scheduler_allowed_nested_schedule_classes` hook), but I think it can be promoted to class level and made clearer. 

> I'd trust your gut feeling.

I'm definitely up for this work being challenged. It's a tricky line we're having to follow between full-on correctness and keeping existing sites working.

barryhughes:
✍🏼 This was discussed elsewhere, but it's possibly worth dropping a note in here, too, to make some of the background thinking a little clearer.

An obvious alternative or supplement to this approach would be to adopt JSON-based (de)serialization of scheduling information. The problem with that is that, unless we also  continue to write serialized blobs in their current form and current location — and we could, but it raises bigger performance questions that the current approach alone — then a version rollback would potentially render a large number of scheduled actions invalid (because older versions could not read that information).

That's a particular concern for this library, because multiple plugins may ship different (and sometimes quite old) versions of Action Scheduler, so a rollback can happen indirectly simply by a plugin that is no longer required being deactivated.

classes/ActionScheduler_ScheduleDeserializer.php
kalessil: I see, let's keep as it is. We can take this risk if security reports point to this attack vector again after this solution is merged. We'll have some options then if this scenario materializes.

classes/ActionScheduler_ScheduleDeserializer.php
kalessil: Weighing out complexity vs. maintainability aspects, `$deserializer   = new self( self::KNOWN_SCHEDULE_CLASSES, $allowed_nested );` is a clear winner. 

What I read from this thread is that my proposal is actually low ROI: micro-optimization with potential risk to multi-site (and likely more hidden scenarios). Let's ignore this one.

classes/data-stores/ActionScheduler_wpPostStore.php
kalessil: Nice, thanks for double-checking! Marking resolved.

PR #1345, opened 2026-07-11
(Optional) improvements for handling a subset of deserialization errors

barryhughes:
- Still a draft
- Potential enhancement for https://github.com/woocommerce/action-scheduler/pull/1344

### DIFF
--- a/classes/ActionScheduler_AdminView.php
+++ b/classes/ActionScheduler_AdminView.php
@@ -306,19 +306,20 @@ class ActionScheduler_AdminView extends ActionScheduler_AdminView_Deprecated {
 			'_asnonce'
 		);
 
-		// Markup is kept out of the translatable strings: each translatable piece is plain text, escaped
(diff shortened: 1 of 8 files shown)


barryhughes:
Closing. This is currently incomplete, and it's not really clear if we are going in the right direction: what we really need to do is have a little more discussion about when and how we notify site operators of problems. 

The specific issue we were concerned with here (cancellation due to unsafe scheduling data) is expected to occur only very rarely, and many sites will never encounter it at all. This begs the question, why go to special measures for this scenario when other action failures receive no such additional support (yet, something like a series of failing payment actions could be a serious matter for a merchant)? 

A partial answer is that by merging https://github.com/woocommerce/action-scheduler/pull/1344 we would be introducing a new reason for automatic action cancellation that did not previous exist, but the question still stands.

_Last note: some (at least partly related) prior art exists in the form of past-due notices and the past-due action view._

Issue #1107, opened 2024-10-02
"clean" command works incorrectly


slaFFik:
Here is the command I was using:

```
wp action-scheduler clean --before='31 days ago' --status=complete,failed,canceled
```

I did various tests, and it looks like:

1. it does not correctly report the number of cleaned (removed) actions - it consistently shows a higher number than it actually removed (for example, on a small DB 88 actions were removed from the `_actions` table, while CLI reported 101; same with bigger DB).

2. it does not remove actions that are satisfying the filtering criteria:
After running the command above I still see actions that are older than 31 days, and they have a `failed` status. According to my command - they should be deleted.

3. It does not remove all records of associated log entries from the `_logs` table.
So I see log entries referencing `action_id` that was removed.
I think the `clean` command should purge such records as well.

4. Old claims are not deleted too.

---

There are no custom filters that modify AS behavior to affect the results above.

PR #1346, opened 2026-07-12
Fix default 'before' date in WP-CLI clean command #1343

### LINKED ISSUE #1343: WP-CLI `clean` command ignores its documented 31-day default and deletes ALL completed/canceled actions when `--before` is omitted
## Summary

The `wp action-scheduler clean` command documents that `--before` "Defaults to 31 days", but the implementation defaults it to an empty string, which resolves to the **current time**. As a result, running the command without `--before` deletes **every** completed and canceled action regardless of age, instead of only those older than 31 days.

Real-world impact: we run this command from an hourly system cron. It silently wiped the entire Action Scheduler history + logs every hour, which cost us significant time while debugging unrelated issues — there was never any execution history to look at.

## Affected code

`classes/WP_CLI/ActionScheduler_WPCLI_Clean_Command.php`, method `clean()`.

The docblock promises (line 22):

```
[--before=<datestring>]
: Only delete actions with scheduled date older than this. Defaults to 31 days. e.g `--before='7 days ago'`, `--before='02-Feb-2020 20:20:20'`
```

But the implementation (lines 39, 45–49):

```php
$before  = \WP_CLI\Utils\get_flag_value( $assoc_args, 'before', '' );
...
try {
    $lifespan = as_get_datetime_object( $before );
} catch ( Exception $e ) {
    $lifespan = null;
}
```

Why this fails:

1. When `--before` is omitted, `$before` is `''` (empty string), not `null`.
2. `as_get_datetime_object( '' )` calls `new ActionScheduler_DateTime( '' )`, and PHP parses an empty date string as **"now"**. No exception is thrown, so the `catch` branch that would produce `null` is never reached.
3. `ActionScheduler_QueueCleaner::clean_actions()` does contain a proper 30-day fallback, but only for a `null` cutoff (line 100):

   ```php
   $cutoff = ! is_null( $cutoff_date ) ? $cutoff_date : as_get_datetime_object( $this->month_in_seconds . ' seconds ago' );
   ```

   The CLI always passes a real DateTime ("now"), so this fallback is unreachable from the CLI path. (It is also unreachable via the exception path, because the `$cutoff_date` parameter is type-hinted as non-nullable `DateTime`; passing `null` would raise a `TypeError`.)

Net effect: the cutoff is "everything modified before this instant" — i.e. no retention at all.

## Steps to reproduce

1. Have completed/canceled actions in the store that are **younger** than 31 days.
2. Run `wp action-scheduler clean` with no arguments.
3. All completed/canceled actions are deleted.

Verified reproduction on a store containing 68 completed actions younger than 30 days and 5 canceled actions older than 30 days: the command deleted all **73**. Expected: 5.

After patching the default (see below) and re-testing with one completed action aged 40 days and one aged 5 days: exactly the 40-day action was deleted and the 5-day action was retained.

## Expected behavior

With `--before` omitted, only actions older than 31 days are deleted, as documented.

## Actual behavior

All completed/canceled actions are deleted, regardless of age.

## Proposed fix

Make the default match the documentation:

```diff
--- a/classes/WP_CLI/ActionScheduler_WPCLI_Clean_Command.php
+++ b/classes/WP_CLI/ActionScheduler_WPCLI_Clean_Command.php
@@ method clean()
 		$status  = explode( ',', WP_CLI\Utils\get_flag_value( $assoc_args, 'status', '' ) );
 		$status  = array_filter( array_map( 'trim', $status ) );
-		$before  = \WP_CLI\Utils\get_flag_value( $assoc_args, 'before', '' );
+		$before  = \WP_CLI\Utils\get_flag_value( $assoc_args, 'before', '31 days ago' );
 		$sleep   = \WP_CLI\Utils\get_flag_value( $assoc_args, 'pause', 0 );
```

An alternative would be treating an empty value as "no explicit cutoff" and letting `ActionScheduler_QueueCleaner` apply its own retention default, but that requires making the `$cutoff_date` parameter nullable; the one-line default above is the minimal, documentation-matching fix. We have verified it in production-like conditions (see reproduction above).

## Environment

- Reproduced on Action Scheduler 3.9.3 (standalone and as bundled library); the identical code is present in the **4.0.0** release and current `trunk`, so all current versions are affected
- WordPress + WP-CLI, PHP 8.5
- Store: `ActionScheduler_HybridStore` / logger `ActionScheduler_DBLogger` (behavior is store-independent — the cutoff is wrong before the store is queried)

## Workaround

Always pass the cutoff explicitly: `wp action-scheduler clean --before='31 days ago'`.

## Related

- #1107 (`clean` reports incorrect deletion counts and leaves orphaned log entries) — different defect in the same command; possibly worth fixing together.

MagicGeny:
### Description
Closes #1343

This PR fixes an issue where running the `wp action-scheduler clean` command without specifying the `--before` flag incorrectly deletes **all** completed/failed/cancelled actions instead of keeping the last 31 days of data as stated in the documentation.

### Cause of the bug
In `ActionScheduler_WPCLI_Clean_Command::clean()`, the default value for the `before` argument was set to an empty string `''`. When passed further down into `strtotime()`, this empty string evaluates to "now", causing the cleaner to wipe out even newly created/completed actions.

### Solution
Changed the default value from `''` to `'31 days ago'` to align the default CLI behavior with the plugin's default retention period and documentation.

### How to test
1. Create some fresh completed actions: 
   `wp eval 'for($i=0; $i<5; $i++) { $id = as_schedule_single_action(time(), "test", [], "test"); ActionScheduler::store()->mark_complete($id); }'`
2. Run the clean command without flags: `wp action-scheduler clean`
3. **Before this fix:** All 5 fresh actions are deleted (`5 actions deleted.`).
4. **After this fix:** The fresh actions are safely preserved (`0 batches processed.`).

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -3,6 +3,7 @@
 = 4.1.0 - xxxx-xx-xx =
 * Use `wp_admin_notice()` to render admin notices, instead of custom HTML.
 * Fix an oversight in the lock implementation (used to rate-limit async request runners) that could result in a database error in unusual cases.
+* Fix the behavior of the `wp action-schedule clean` command so that the `--before` option correctly defaults to 31 days ago.
 * Dev - Updates the `wpPostStore` to capture any database errors that might occur while claiming actions.
 * Performance - Reduce the number of SQL queries on Action Scheduler admin page.
 * Performance - Cache resolved group IDs to reduce the number of SQL queries.
--- a/classes/WP_CLI/ActionScheduler_WPCLI_Clean_Command.php
+++ b/classes/WP_CLI/ActionScheduler_WPCLI_Clean_Command.php
@@ -36,7 +36,7 @@ class ActionScheduler_WPCLI_Clean_Command extends WP_CLI_Command {
 		$batches = absint( \WP_CLI\Utils\get_flag_value( $assoc_args, 'batches', 0 ) );
 		$status  = explode( ',', WP_CLI\Utils\get_flag_value( $assoc_args, 'status', '' ) );
 		$status  = array_filter( array_map( 'trim', $status ) );
-		$before  = \WP_CLI\Utils\get_flag_value( $assoc_args, 'before', '' );
+		$before  = \WP_CLI\Utils\get_flag_value( $assoc_args, 'before', '31 days ago' );
 		$sleep   = \WP_CLI\Utils\get_flag_value( $assoc_args, 'pause', 0 );
 
 		$batches_completed = 0;


MagicGeny:
Hi @barryhughes! 

I wanted to provide some context for this PR. This change resolves the default 'before' value issue for the cleanup process. 

It is also a crucial dependency for another bugfix I just submitted in #13480 (which addresses the incorrect 'before' filtering in the WP-CLI clean command, #1107). 

I have tested this locally, and all tests pass. Please let me know if you would like me to adjust anything here so we can move this forward. Thank you!

barryhughes:
Sorry for the delay here ... and good catch!

MagicGeny:
@barryhughes Barry! Thanks! I tried to figure out what was wrong with the tests, but it looks like GitHub encountered errors related to network timeouts/429 Too Many Requests during the cloning step, and for some reason, it failed to set up multiple environments. Could you re-run the error checking when you have a free minute?

barryhughes:
Yeah we hit this now and again. I'll re-run it periodically until it passes (or until we get a genuine failure, though I don't expect any given the change and given it runs perfectly well locally). All good!

barryhughes (approved):
LGTM, thank you for contributing 💪

To test, I used a variation of the testing instructions that created 24 completed actions, marked as having been completed at various points in time (some older than 31 days, some very new):

```php
function mark_completed_at( $action_id, $time ) {
	global $wpdb;

	$time = date( 'Y-m-d H:i:s', $time );

	$wpdb->update(
		$wpdb->prefix . ActionScheduler_StoreSchema::ACTIONS_TABLE, [
			'last_attempt_gmt'   => $time,
			'last_attempt_local' => $time, # We don't actually need to adjust between local/GMT for testing purposes
			'status'             => ActionScheduler_Store::STATUS_COMPLETE,
		], [
			'action_id' => $action_id,
		]
	);
}

for ( $i = 0; $i < 24; $i++ ) {
	mark_completed_at(
		as_schedule_single_action( time() + DAY_IN_SECONDS, 'test-pr-1346' ),
		time() - ( $i * DAY_IN_SECONDS * 2 )
	);
}
```

I then invoked the clean command with and without the `--before` flag. The counts were as expected:

```
% wp action-scheduler clean                      
Success: 8 actions deleted.
1 batch processed.

% wp action-scheduler clean --before="7 days ago"
Success: 12 actions deleted.
1 batch processed.

% wp action-scheduler clean --before="tomorrow"  
Success: 4 actions deleted.
1 batch processed.
```

MagicGeny:
Thank you for the thorough review and testing, Barry @barryhughes! Glad to help improve the tool

PR #1348, opened 2026-07-14
Fix/clean uses scheduled date 1107 (`wp action-scheduler clean` filtering by `last_attempt_gmt` instead of `scheduled_date_gmt`)

### LINKED ISSUE #1107: "clean" command works incorrectly

MagicGeny:
### Problem

The `wp action-scheduler clean` command was incorrectly filtering actions by their **last modified date** (`last_attempt_gmt`) instead of their **scheduled date** (`scheduled_date_gmt`). 

The `--before` parameter in the CLI command is documented as referring to the scheduled date — e.g. `--before='31 days ago'` should delete actions **scheduled** more than 31 days ago. However, the query was using `last_attempt_gmt`, so any action that had been retried or modified recently would **escape cleanup** even if it was scheduled far beyond the retention period.

**Example of the bug:**
- Action A: scheduled 60 days ago, last ran 60 days ago → correctly identified for deletion.
- Action B: scheduled 60 days ago, but retried 1 hour ago (recent `last_attempt_gmt`) → **incorrectly kept**, even though it's well past the retention period.

This meant that over time, a growing backlog of old actions would accumulate in the database if they had ever been retried or re-scheduled after their initial run.

### Root Cause

`ActionScheduler_QueueCleaner::clean_actions()` always queried actions using the `modified` field (`last_attempt_gmt`), regardless of who called it. Both the scheduled (cron) cleanup and the WP-CLI command used the same `modified`-based filtering.

The WP-CLI clean command should use the **scheduled date** (`scheduled_date_gmt`) since the `--before` parameter refers to when the action was scheduled to run, not when it last executed.

#### 1. `classes/ActionScheduler_QueueCleaner.php`
- Added `$date_type` parameter to `clean_actions()` with two modes:
  - `'modified'` (default) — filters by `last_attempt_gmt`. Used by scheduled cron cleanup.
  - `'date'` — filters by `scheduled_date_gmt`. Used by WP-CLI clean command.
- Scheduled cleanup (`delete_old_actions()`) **always forces** `$date_type = 'modified'`, preserving backward compatibility and the existing execution budget optimization.
- Updated PHPDoc for the new parameter.

#### 2. `classes/WP_CLI/ActionScheduler_WPCLI_Clean_Command.php`
- The `clean()` method now passes `'date'` as the 5th argument to `$cleaner->clean_actions()`, so the CLI correctly filters by scheduled date.

#### 3. `classes/data-stores/ActionScheduler_DBStore.php`
- When deleting an action, the store now checks the action's `claim_id` before deletion.
- After deletion, if no other actions reference that claim, the orphan claim is automatically deleted from `{$wpdb->actionscheduler_claims}`.
- This prevents orphaned claims from accumulating in the database.

#### 4. `classes/data-stores/ActionScheduler_DBLogger.php`
- In WP-CLI context (`defined('WP_CLI') && WP_CLI`), `clear_deleted_action_logs()` now performs a **direct DELETE** query instead of the batched approach using scheduled actions.
- This avoids the problem of scheduling follow-up actions during CLI cleanup, which could cause the CLI process to exit before logs are fully purged.

#### 5. `classes/data-stores/ActionScheduler_wpCommentLogger.php`
- Added `clear_deleted_action_logs()` method that cleans up both `commentmeta` and `comments` records for deleted actions.
- Registered via `action_scheduler_deleted_action` and `action_scheduler_canceled_corrupted_action` hooks.
- Previously, the comment-based logger never cleaned up logs when actions were deleted.

#### Automated Tests (4 new regression tests in `ActionScheduler_QueueCleaner_Test.php`)

```bash
composer run test -- --filter=ActionScheduler_QueueCleaner_Test


The new tests are:

1. __`test_clean_actions_uses_scheduled_date_when_date_type_is_date`__ — Verifies that with `$date_type = 'date'`, both actions (even one with recent `last_attempt_gmt`) are correctly deleted based on scheduled date.
2. __`test_clean_actions_with_modified_date_misses_recently_touched_actions`__ — Demonstrates the original bug: filtering by `modified` causes recently-retried actions to escape cleanup.
3. __`test_scheduled_cleanup_uses_modified_date`__ — Verifies scheduled cron cleanup still uses `modified` date (backward compatibility).
4. __`test_orphan_claim_cleaned_on_action_delete`__ — Verifies orphan claims are cleaned up when their last referencing action is deleted.

#### Manual Test

1. Schedule two actions 32 days ago and mark them complete:

```php
as_schedule_single_action( strtotime('-32 days'), 'my_hook', array(), 'test' );
as_schedule_single_action( strtotime('-32 days'), 'my_hook', array(), 'test' );
```

2. Manually update one action's `last_attempt_gmt` to simulate a recent retry:

```sql
SET last_attempt_gmt = UTC_TIMESTAMP(),
    last_attempt_local = NOW()
WHERE action_id = <second_action_id>;
```

3. Run the clean command:

```bash
wp action-scheduler clean --before='31 days ago' --batch-size=100
```

4. __Before fix__: Only the first action is deleted. The second action remains.
5. __After fix__: Both actions are deleted.

#### Regression Test

Verify that scheduled (cron) cleanup still works as expected:

- Create a completed action with a recent `last_attempt_gmt` (within the last hour).
- Run `wp cron event run action_scheduler_run_actions_cleanup_hook`.
- The action should __not__ be deleted (scheduled cleanup uses `modified` date).

### Related

- Fixes #1107
- Previous PR: https://github.com/woocommerce/action-scheduler/pull/1346

### DIFF
--- a/classes/ActionScheduler_QueueCleaner.php
+++ b/classes/ActionScheduler_QueueCleaner.php
@@ -188,7 +188,7 @@ class ActionScheduler_QueueCleaner {
 	 *
 	 * @return array Actions deleted.
 	 */
-	public function clean_actions( array $statuses_to_purge, DateTime $cutoff_date, $batch_size = null, $context = 'old' ) {
+	public function clean_actions( array $statuses_to_purge, DateTime $cutoff_date, $batch_size = null, $context = 'old', $date_type = 'modified' ) {
 		$batch_size        = ! is_null( $batch_size ) ? $batch_size : $this->batch_size;
 		$cutoff            = ! is_null( $cutoff_date ) ? $cutoff_date : as_get_datetime_object( $this->month_in_seconds . ' seconds ago' );
 		$lifespan          = time() - $cutoff->getTimestamp();
@@ -198,6 +198,9 @@ class ActionScheduler_QueueCleaner {
 		// For inline cleanup during a queue run, the batch size should remain unchanged to avoid increasing the process footprint.
 		$is_scheduled_cleanup = doing_action( self::RUN_SCHEDULED_CLEANER_HOOK )
 			|| doing_action( self::CONTINUE_SCHEDULED_CLEANER_HOOK );
+		if ( $is_scheduled_cleanup ) {
+			$date_type = 'modified';
+		}
 		// 250 balances replication safety, backlog clearance speed, and claim slot duration on high-volume stores.
 		$iteration_batch_size       = $is_scheduled_cleanup ? max( 250, $batch_size ) : $batch_size;
 		$iteration_unused_budget    = 0;
@@ -231,15 +234,19 @@ class ActionScheduler_QueueCleaner {
 		$deleted_actions = array();
 		foreach ( $statuses_to_purge as $status ) {
 			$iteration_execution_budget = $iteration_batch_size + $iteration_unused_budget;
-			$actions_to_delete          = $this->store->query_actions(
-				array(
-					'status'           => $status,
-					'modified'         => $cutoff,
-					'modified_compare' => '<=',
-					'per_page'         => $iteration_execution_budget,
-					'orderby'          => 'none',
-				)
+			$query                    = array(
+				'status'   => $status,
+				'per_page' => $iteration_execution_budget,
+				'orderby'  => 'none',
 			);
+			if ( 'date' === $date_type ) {
+				$query['date']         = $cutoff;
+				$query['date_compare'] = '<=';
+			} else {
+				$query['modified']         = $cutoff;
+				$query['modified_compare'] = '<=';
+			}
+			$actions_to_delete          = $this->store->query_actions( $query );
 			$deleted_actions[]          = $this->delete_actions( $actions_to_delete, $lifespan, $context );
 
 			$fetched_actions_count      = count( $actions_to_delete );
--- a/classes/WP_CLI/ActionScheduler_WPCLI_Clean_Command.php
+++ b/classes/WP_CLI/ActionScheduler_WPCLI_Clean_Command.php
@@ -36,7 +36,7 @@ class ActionScheduler_WPCLI_Clean_Command extends WP_CLI_Command {
 		$batches = absint( \WP_CLI\Utils\get_flag_value( $assoc_args, 'batches', 0 ) );
 		$status  = explode( ',', WP_CLI\Utils\get_flag_value( $assoc_args, 'status', '' ) );
 		$status  = array_filter( array_map( 'trim', $status ) );
-		$before  = \WP_CLI\Utils\get_flag_value( $assoc_args, 'before', '' );
+		$before  = \WP_CLI\Utils\get_flag_value( $assoc_args, 'before', '31 days ago' );
 		$sleep   = \WP_CLI\Utils\get_flag_value( $assoc_args, 'pause', 0 );
 
 		$batches_completed = 0;
--- a/classes/WP_CLI/ActionScheduler_WPCLI_Clean_Command.php
+++ b/classes/WP_CLI/ActionScheduler_WPCLI_Clean_Command.php
@@ -58,7 +58,7 @@ class ActionScheduler_WPCLI_Clean_Command extends WP_CLI_Command {
 					sleep( $sleep );
 				}
 
-				$deleted = count( $cleaner->clean_actions( $status, $lifespan, null, 'CLI' ) );
+				$deleted = count( $cleaner->clean_actions( $status, $lifespan, null, 'CLI', 'date' ) );
 				if ( $deleted <= 0 ) {
 					break;
 				}
--- a/classes/data-stores/ActionScheduler_DBLogger.php
+++ b/classes/data-stores/ActionScheduler_DBLogger.php
@@ -121,6 +121,11 @@ class ActionScheduler_DBLogger extends ActionScheduler_Logger {
 	 * @param int $action_id Action ID.
 	 */
 	public function clear_deleted_action_logs( $action_id ) {
+		if ( defined( 'WP_CLI' ) && WP_CLI ) {
+			global $wpdb;
+			$wpdb->delete( $wpdb->actionscheduler_logs, array( 'action_id' => $action_id ), array( '%d' ) );
+			return;
+		}
 		$this->clear_deleted_action_logs_single_batch( $action_id, -1, 1, 4000 );
 	}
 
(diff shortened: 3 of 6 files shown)


MagicGeny:
Hi @barryhughes! 

I have created this PR to address the issues with the `wp action-scheduler clean` command (#1107). 

Since this bugfix relies on the default value fix from my previous task (#1343 / #1346), I have branched this PR from that work to ensure everything functions correctly together. 

I have also written 4 new automated regression tests to cover these changes and ran them successfully in my local environment. 

Could you please take a look when you have a moment? I would really appreciate your feedback on whether this approach looks good to you, or if we should merge the previous default-value fix first.

Thanks!

barryhughes:
Thanks!

One thing I am unsure of is whether or not changing the current behavior is desirable in all cases.  Bear in mind that we can offer no real guarantees as to when an action is actually processed, so we might for instance have:

- An action scheduled for 2026-07-01
- That actually runs on 2026-08-01 (one month later)
- Under the current model, and assuming defaults, it will be retained for another month
- Under the proposed change, it would be removed almost on the very next queue cleaner invocation, providing little opportunity for debugging etc

My hypothetical timeline is a deliberate exaggeration (hopefully things don't normally look like this), but it's a useful way to think about the difference in behavior.

Re-framing a little: might it make more sense to see this as a bug in the documentation, rather than a functional bug?

MagicGeny:
Hi @barryhughes Barry! Thanks for highlighting this. You are absolutely right—in a delayed queue scenario (like the 2026-07-01 to 2026-08-01 example), relying strictly on scheduled_date_gmt would wipe the logs almost immediately after execution, destroying the audit trail.

However, treating this purely as a documentation bug leaves high-volume stores vulnerable to the core issue reported in #1107. When actions are frequently retried or rescheduled, their last_attempt_gmt lease renews indefinitely. This results in massive database bloat (millions of stale rows) because they never gracefully age out.

To balance the need for a reliable debugging window with database performance, I see two potential paths forward:

1. A Hybrid Default (Safety Net): We filter by scheduled_date_gmt but enforce a hard safety check on last_attempt_gmt (e.g., ensuring the action hasn't been touched in the last 7 or 14 days). This guarantees a debugging window while eventually clearing old schedule loops.

2. Opt-in CLI Flag (Safest for BC): We leave the default logic as last_attempt_gmt (and update the docs), but introduce a --by-scheduled-date flag for the CLI command. This allows store maintainers with massive database backlogs to explicitly clear out the queue based on the original timeline without affecting standard users.

I'd love to hear your thoughts on which direction feels safer and more aligned with the plugin's goals! I'd be happy to update the PR to implement either of these.

barryhughes:
> When actions are frequently retried or rescheduled, their last_attempt_gmt lease renews indefinitely. This results in massive database bloat (millions of stale rows) because they never gracefully age out.

In terms of clean-up, we are (by default) only interested in completed, cancelled or failed actions. Ordinarily they would not be retried.

There is also the slightly different case of a pending action which has been claimed but never processed: actions like these are periodically reset (they are released from the claim, so a future queue runner can pick them up) ... but that also does not seem to me like a good case for basing clean-up on the originally scheduled date.

Is the scenario you outlined, in which `last_attempt_gmt` is renewed indefinitely, something you are experiencing? Is any custom code involved?

MagicGeny:
Hi @barryhughes Barry! 

You make a very valid point. Thanks! In a standard setup, actions in terminal statuses (completed, failed, cancelled) shouldn't be retried or have their `last_attempt_gmt` renewed indefinitely. 
If a store experiences bloat, it's typically due to pure volume rather than an infinite retention lease loop. 
Changing the default behavior to `scheduled_date_gmt` indeed risks destroying the crucial debugging window for delayed queues. 


I completely agree with your perspective. This is better treated as a documentation bug rather than a functional one. 
This new functionality will be perceived by most users as a critical bug since they are already accustomed to the old behavior.

To resolve this properly, I will do the following:
1. Revert the changes to the date filtering logic (`$date_type` in `ActionScheduler_QueueCleaner` and the WP-CLI command arguments), keeping the reliable `last_attempt_gmt` approach.
2. Update the WP-CLI `--before` docblock/help text to clarify that it filters by the last activity/modified date (`last_attempt_gmt`) rather than the original scheduled date, aligning the documentation with the actual safe behavior.

However, IMHO, this PR also introduces fixes for the other independent database bloat issues reported in #1107:
* Cleaning up orphaned claims from the `actionscheduler_claims` table upon action deletion.
* Properly purging associated action log entries from both `_logs` and `comments` tables during CLI cleanup.

I'll update the branch to remove the controversial date-logic changes while keeping these highly beneficial log and claim cleanup optimizations. Let me know if that sounds good to you!

barryhughes:
Hey there, @MagicGeny.

> To resolve this properly, I will do the following:
>
> - Revert the changes to the date filtering logic (`$date_type` in `ActionScheduler_QueueCleaner` and the WP-CLI command arguments), keeping the reliable `last_attempt_gmt` approach.
> - Update the WP-CLI `--before` docblock/help text to clarify that it filters by the last activity/modified date (`last_attempt_gmt`) rather than the original scheduled date, aligning the documentation with the actual safe behavior.

That sounds good to me, and/or you could start from scratch with a fresh PR (whatever's easiest, we don't have a strong preference).

> However, IMHO, this PR also introduces fixes for the other independent database bloat issues reported in https://github.com/woocommerce/action-scheduler/issues/1107:
>
> - Cleaning up orphaned claims from the `actionscheduler_claims` table upon action deletion.
> - Properly purging associated action log entries from both `_logs` and `comments` tables during CLI cleanup.

I would be inclined to create either one or two fresh PRs for these matters, rather than bundling them with a documentation change, and then we can discuss them separately. Some (provisional/early) thoughts, fwiw:

- In relation to the proposed changes for claim cleanup, I have some concerns about querying and counting extant claims on each `delete_action()` call, and wonder if it would be better to perform this work as part of the queue cleaning process.
- In relation to the current log cleanup, and in particular WP Comment-based logging, since we already perform a force delete on the action post, I'm not immediately sold on the idea of us duplicating that work (and am not sure why this is better than letting WP take care of it during forced post deletion).

Thank you!

PR #1350, opened 2026-07-24
Add light colour scheme for GitHub Pages docs site

### LINKED ISSUE #1321: Is there way to toggle to light mode in the actionscheduler.org website? I can't read text on the website.
Is there way to toggle to light mode in the actionscheduler.org website? I can't read text on the website.
Or please implement light mode for thous who don't see in dark mode.

barryhughes:
The current documentation site only renders in 'dark mode'. This isn't ideal for visitors who prefer a light mode, as detailed in the linked issue. Note that this change does not introduce a toggle as suggested, but instead tries to respect the OS-level light/dark mode setting.

light-scheme

Closes #1321

---

- Testing this is more difficult than you might think (or, I'm missing a dimension about how to set it up) as the theme etc we see with the deployed site is not the result I get when previewing things locally/via Jekyll.
- However, we *can* simulate it by appending the CSS in the diff to the end of the current stylesheet (and then use browser dev tools to force light mode, depending on your OS configuration, etc).

### DIFF
--- a/docs/assets/css/style.scss
+++ b/docs/assets/css/style.scss
@@ -55,3 +55,60 @@ footer {
 		animation:octocat-wave 560ms ease-in-out
 	}
 }
+
+// Light colour scheme for visitors whose OS/browser prefers a light theme.
+// See https://github.com/woocommerce/action-scheduler/issues/1321.
+@media (prefers-color-scheme: light) {
+	body {
+		background: #ebf4fa;
+		color: #021b34;
+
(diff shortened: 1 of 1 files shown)


barryhughes (commented):
The hope here is that this gives us a straightforward solution we can continue to iterate on as needed (possibly at some point, we could simply adopt a more current theme from Jekyll).

marcusdevcode:
Thanks!

Issue #1243, opened 2025-03-06
Run continuously


szepeviktor:
Hello!
I'd like to use WP-CLI to run queued actions non-stop.
Should I write a never-ending shell script? Is there a better way?
```shell
while true; do
    wp action-scheduler run
    sleep 1
done
```
Thank you.

barryhughes:
I imagine that would work in practice. Some thoughts to factor in:

- This would effectively mean there is always at least one active batch. Depending on your configuration, this may prevent other queue runners (like the default async HTTP-triggered runner) from processing actions. In turn, the CLI queue runner may not run for the same reason *('too many concurrent batches').*
- It will keep consuming actions until there are no more pending actions left, so you could be prone to memory leaks and/or leaky state if the individual processes run for too long. The `--batches` option may help with this.
- A different common pattern is to invoke it once every minute via `cron.d` (and other system schedulers may offer increased granularity if once-per-minute is too coarse).

(I realize this was open for quite some time without a response, apologies!)

PR #1351, opened 2026-07-24
Add persistent WP-CLI queue worker

szepeviktor:
## Summary

Adds `wp action-scheduler work`, a persistent worker built on the existing WP-CLI queue runner.

Like Laravel's `queue:work`, it provides low-latency processing without repeated WordPress bootstraps, process-monitor integration, graceful shutdown, and configurable worker recycling. Existing Action Scheduler claims, concurrency limits, groups, hooks, and logging remain in use.

Inspired by @barryhughes in #1243


barryhughes:
Thanks for this! I don't have time to fully review this week, so am initially just leaving some thoughts/questions.

- Should this be folded into `wp action-scheduler run`?
  - Various arguments are common to/shared with the existing `run` sub-command
  - Various other arguments might be just as useful if applied to `run` (for example, `--memory-limit`)
- Could we take advantage of [`WP_CLI::runcommand()`](https://make.wordpress.org/cli/handbook/references/internal-api/wp-cli-runcommand/) (with `'launch' => true`):
  - Execution could be delegated to a sub-process
  - Once limits are met or exceeded, a fresh sub-process could be spawned after `--sleep` seconds
  - And/or if we don't do this, we may wish to check that `pcntl` is available to us
- Other stuff:
  - If the user `ctrl+c`'s (or just if we receive a signal that the process should stop), there is potentially a very long delay before this takes effect (depending on batch size, duration of individual actions, etc), because of where in the loop we perform this check
  - As with some other newly introduced commands, we should consider using namespaces (`Action_Scheduler\WP_CLI`)
  - On each loop where we trigger `$runner->setup()` we indirectly trigger the queue cleaner; depending on the circumstances we may end up doing this *a lot*

szepeviktor:
> Should this be folded into `wp action-scheduler run`?

Which one do you choose? --keep-alive or --continuous

barryhughes:
Of the two, I think I prefer `--continuous` ... but a further alternative might be `--poll-every=x`, where *x* is a number of millisecs (and in fact, naming it `--poll-every-ms=x` would be clearer yet). 

In that way, we roll at least two args into one, because there would be no need for `--sleep` (extra benefit: it eliminates the risk of `--sleep` and `--pause` being confused with one another).

This does mean we are requiring the caller to specify the delay between outer loop cycles ... but I feel that is more than reasonable in this situation.

szepeviktor:
All implemented 🍏

PR #1353, opened 2026-07-26
Fix #1107: Purge orphan claims during queue cleaning pass

### LINKED ISSUE #1107: "clean" command works incorrectly

MagicGeny:
### Description
Following up on the discussion in #1348 regarding issue #1107, this PR explicitly addresses the **accumulation of orphaned claim records** in the `actionscheduler_claims` table. 

As suggested by @barryhughes Barry, rather than checking and deleting claims dynamically during every single `delete_action()` call (which would introduce unwanted query overhead), this implementation handles the cleanup **batch-wise**. The orphan claims are now purged via a single optimized `LEFT JOIN` query at the very end of a queue cleaning pass in `ActionScheduler_QueueCleaner::clean_actions()`.

### Key Changes
1. **`ActionScheduler_Store` & `ActionScheduler_DBStore`**: Introduced `purge_orphan_claims()` which deletes rows from the claims table where no corresponding `action_id` exists.
2. **`ActionScheduler_QueueCleaner`**: Hooked the batch purge to run once per `clean_actions()` call execution loop.
3. **Test Suites**:
   - Added 5 new regression tests in `ActionScheduler_DBStore_Test` covering exact cleanup counts, referenced claim preservation, and no-op behavior.
   - Fixed database state leaking in `ActionScheduler_QueueCleaner_Test` via an explicit `: void` compatible `tearDown()` block ensuring clean runs.

### Related Issues
Fixes part of #1107 (Specifically point 4: "Old claims are not deleted too").

### How to test the changes
1. Run DBStore tests to ensure the core purging logic works seamlessly:
   ```bash
   vendor/bin/phpunit tests/phpunit/jobstore/ActionScheduler_DBStore_Test.php

### DIFF
--- a/classes/abstracts/ActionScheduler_Store.php
+++ b/classes/abstracts/ActionScheduler_Store.php
@@ -498,6 +498,15 @@ abstract class ActionScheduler_Store extends ActionScheduler_Store_Deprecated {
 	}
 
 	/**
+	 * Purge claim records that are no longer referenced by any action.
+	 *
+	 * @return int|false Number of orphan claims deleted, or false on failure.
+	 */
+	public function purge_orphan_claims() {
+		return 0;
+	}
+
+	/**
 	 * Get instance.
 	 *
 	 * @return ActionScheduler_Store
--- a/classes/data-stores/ActionScheduler_DBStore.php
+++ b/classes/data-stores/ActionScheduler_DBStore.php
@@ -1179,6 +1179,7 @@ AND args = %s
 		global $wpdb;
 
 		$sql = "DELETE c FROM {$wpdb->actionscheduler_claims} c LEFT JOIN {$wpdb->actionscheduler_actions} a ON a.claim_id = c.claim_id WHERE a.action_id IS NULL";
+		// phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared -- No user input, no placeholders needed.
 		$deleted = $wpdb->query( $sql );
 
 		return false === $deleted ? false : (int) $deleted;
--- a/classes/data-stores/ActionScheduler_DBStore.php
+++ b/classes/data-stores/ActionScheduler_DBStore.php
@@ -1166,6 +1166,25 @@ AND args = %s
 	}
 
 	/**
+	 * Purge claim records that are no longer referenced by any action.
+	 *
+	 * @return int|false Number of orphan claims deleted, or false on failure.
+	 */
+	public function purge_orphan_claims() {
+		/**
+		 * Global.
+		 *
+		 * @var \wpdb $wpdb
+		 */
+		global $wpdb;
+
+		$sql = "DELETE c FROM {$wpdb->actionscheduler_claims} c LEFT JOIN {$wpdb->actionscheduler_actions} a ON a.claim_id = c.claim_id WHERE a.action_id IS NULL";
+		$deleted = $wpdb->query( $sql );
+
+		return false === $deleted ? false : (int) $deleted;
+	}
+
+	/**
 	 * Return an action's claim ID, as stored in the claim_id column.
 	 *
 	 * @param string $action_id Action ID.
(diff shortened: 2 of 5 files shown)


classes/abstracts/ActionScheduler_Store.php
```diff
@@ -497,6 +497,15 @@ public function mark_migrated( $action_id ) {}
 	public function flush_caches() {
 	}
 
+	/**
+	 * Purge claim records that are no longer referenced by any action.
+	 *
+	 * @return int|false Number of orphan claims deleted, or false on failure.
+	 */
+	public function purge_orphan_claims() {
+		return 0;
```
barryhughes, line 506: Nit: does it make more sense to return `(bool) false` here, on the basis that the concrete store has not implemented the method? Whereas, to my mind, `(int) 0` suggests there was no problem and we simply found that there was nothing which needed to be purged.

classes/data-stores/ActionScheduler_DBStore.php
```diff
@@ -1165,6 +1165,26 @@ public function get_claim_count() {
...
+	 * Purge claim records that are no longer referenced by any action.
+	 *
+	 * @return int|false Number of orphan claims deleted, or false on failure.
+	 */
+	public function purge_orphan_claims() {
+		/**
+		 * Global.
+		 *
+		 * @var \wpdb $wpdb
+		 */
```
barryhughes, line 1178: I'd be inclined to drop this docblock, I think the role of `$wpdb` is widely understood.

classes/data-stores/ActionScheduler_DBStore.php
```diff
@@ -1165,6 +1165,26 @@ public function get_claim_count() {
...
+		global $wpdb;
+
+		$sql = "DELETE c FROM {$wpdb->actionscheduler_claims} c LEFT JOIN {$wpdb->actionscheduler_actions} a ON a.claim_id = c.claim_id WHERE a.action_id IS NULL";
```
barryhughes, line 1181: This would probably be a rare occurence, but I think we're looking at a race condition here.

Consider the case where this line runs in Process A and meantime over in Process B we begin the process of [claiming actions](https://github.com/woocommerce/action-scheduler/blob/4.0.0/classes/data-stores/ActionScheduler_DBStore.php#L864). Initially, a [new claim record](https://github.com/woocommerce/action-scheduler/blob/4.0.0/classes/data-stores/ActionScheduler_DBStore.php#L865) will be created, but we have [not yet assigned the claim ID to any actions](https://github.com/woocommerce/action-scheduler/blob/4.0.0/classes/data-stores/ActionScheduler_DBStore.php#L1014). If we execute this statement, we basically delete that newly minted claim.

Some possibilities:

- Examine `date_created_gmt`, introduce a buffer period.
- Make said buffer period filterable.

classes/data-stores/ActionScheduler_DBStore.php
```diff
@@ -1165,6 +1165,26 @@ public function get_claim_count() {
...
+		// phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared -- No user input, no placeholders needed.
+		$deleted = $wpdb->query( $sql );
```
barryhughes, line 1183: Further thought: we don't know how many stale claims are sitting in the database ... there could be none, or there could be millions. In the latter case I don't know that an unbounded `DELETE` operation is the best choice.

barryhughes (commented):
Thank you. Left a few thoughts and questions.

PR #1355, opened 2026-07-28
Make unique action inserts atomic

### LINKED ISSUE #1317: [Reliability] Revisit adding unique action solution for better concurrency handling
Currently, `\ActionScheduler_DBStore::build_insert_sql` adds unique actions using a select followed by an insert, which is not an atomic operation. In high-concurrency environments, this gap can result in duplicate unique actions.

We should consider implementing a versioned, nullable unique ID column to prevent duplicates at the atomic level.
Making the column nullable allows (?) us to leave it empty for non-unique actions, what is performance-friendly.

drewmt:
## What

Fixes #1317.

Unique actions currently use a lookup followed by an insert. Concurrent
requests can both pass that lookup and create duplicates.

This change:

- adds a nullable `unique_key` column with a database unique index;
- increments the store schema to version 9 so existing rows receive `NULL`;
- generates a versioned `v1:` identity from the hook, exact arguments, and
  group slug;
- retains the existing lookup so an active non-unique action still blocks a
  matching unique action;
- handles an atomic key conflict as the existing `0` result without masking
  unrelated database errors; and
- releases the key when an action completes, fails, or is canceled, while
  retaining it for pending and in-progress actions.

Schedule type, timestamp, and priority are intentionally excluded from the
identity, preserving the current hook + args + group uniqueness contract.
Using the group slug instead of its database ID also keeps the identity stable
if concurrent requests create the same previously unseen group.

## Review follow-up and compatibility

The existing daily cleanup now releases stale keys on completed, failed and
canceled actions in batches of 1,000. This recovers from custom terminal-status
transitions that bypass normal key release, without deleting action history
or clearing pending/running keys. A full batch reuses the existing cleanup
continuation; a running continuation does not prevent its successor.

Cleanup database errors are logged without throwing, so other hook handlers
can continue. When a unique insert returns the atomic conflict/no-op result,
the store clears only that identity's terminal key and retries once if a key
was released. Pending/running actions stay protected, including a competing
insert between recovery and retry. This path uses the affected-row result
because `ON DUPLICATE KEY UPDATE` does not produce a duplicate-key SQL error.

Existing method signatures, abstract store requirements and hook arguments
are unchanged. The new internal public DBStore callback,
`release_stale_unique_action_keys()`, is additive; subclasses already using
that name should be checked for declaration compatibility. Custom cleaners
that replace the existing scheduled hooks still need their own recovery.

Queries use the current registered site table. Single-site and multisite
bootstrap tests pass; explicit cross-blog switching and large-table/concurrent
runner stress testing were not performed. This follow-up adds no further
schema revision. Deadlock handling and recurring-action uniqueness remain
outside scope.

## Validation

Latest local validation against upstream `6f49ea8` on September 14:

- New stale-key regressions were written failing-first.
- Immediate recovery, interleaved competitor and cleanup-error hook coverage:
  30 tests, 48 assertions, in multisite; each also passed in single-site.
- DBStore and QueueCleaner: 492 tests, 536 assertions.
- Remaining suite excluding the recurring-scheduler baseline class:
  1,806 tests, 1,884 assertions, zero failures and ten existing PHPUnit warnings.
- Changed-line PHPCS, PHP syntax and `git diff --check` pass.
- Runtime: WordPress 7.0.1, PHP 8.3, PHPUnit 8.5.52, MySQL 9.6.0.

The unfiltered suite (1,824 tests, 1,894 assertions) has four failures in
`ActionScheduler_RecurringActionScheduler_Test`; the same failures reproduce
on clean upstream `d12d6ef` in a separate fresh database during the previous
validation. Upstream changes since that baseline affect only the admin Screen
Option and changelog, not the failing scheduler paths. Earlier validation of
the original insert change also confirmed the nullable unique index and
schema version 9 on MariaDB 10.4.

### DIFF
--- a/classes/data-stores/ActionScheduler_DBStore.php
+++ b/classes/data-stores/ActionScheduler_DBStore.php
@@ -142,14 +142,18 @@ class ActionScheduler_DBStore extends ActionScheduler_Store {
 				$data['extended_args'] = $args;
 			}
 
+			if ( $unique && self::STATUS_PENDING === $data['status'] ) {
+				$data['unique_key'] = $this->get_unique_action_key( $action );
+			}
+
 			$insert_sql = $this->build_insert_sql( $data, $unique );
 
 			// phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared -- $insert_sql should be already prepared.
-			$wpdb->query( $insert_sql );
-			$action_id = $wpdb->insert_id;
+			$query_result = $wpdb->query( $insert_sql );
+			$action_id    = $wpdb->insert_id;
 
-			if ( is_wp_error( $action_id ) ) {
-				throw new \RuntimeException( $action_id->get_error_message() );
+			if ( false === $query_result ) {
+				throw new \RuntimeException( $wpdb->last_error ? $wpdb->last_error : __( 'Database error.', 'action-scheduler' ) );
 			} elseif ( empty( $action_id ) ) {
 				if ( $unique ) {
 					return 0;
@@ -186,13 +190,15 @@ class ActionScheduler_DBStore extends ActionScheduler_Store {
 		$column_sql      = '`' . implode( '`, `', $columns ) . '`';
 		$placeholder_sql = implode( ', ', $placeholders );
 		$where_clause    = $this->build_where_clause_for_insert( $data, $table_name, $unique );
+		$conflict_clause = $unique ? 'ON DUPLICATE KEY UPDATE `unique_key` = VALUES(`unique_key`)' : '';
 
-		// phpcs:disable WordPress.DB.PreparedSQL.NotPrepared, WordPress.DB.PreparedSQL.InterpolatedNotPrepared, WordPress.DB.PreparedSQLPlaceholders.UnfinishedPrepare	 -- $column_sql and $where_clause are already prepared. $placeholder_sql is hardcoded.
+		// phpcs:disable WordPress.DB.PreparedSQL.NotPrepared, WordPress.DB.PreparedSQL.InterpolatedNotPrepared, WordPress.DB.PreparedSQLPlaceholders.UnfinishedPrepare	 -- $column_sql and $where_clause are already prepared. $placeholder_sql and $conflict_clause are hardcoded.
 		$insert_query = $wpdb->prepare(
 			"
 INSERT INTO $table_name ( $column_sql )
 SELECT $placeholder_sql FROM DUAL
-WHERE ( $where_clause ) IS NULL",
+WHERE ( $where_clause ) IS NULL
+$conflict_clause",
 			$values
 		);
 		// phpcs:enable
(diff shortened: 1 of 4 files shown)


vbelolapotkov (commented):
Hey @drewmt, thanks for your contribution and apologies that it took us so long to get back to you. 

Speaking of PR this is a genuinely good piece of work and a step forward! I ran a few tests and things work as expected. All the findings (like dead locks handling, ensuring recurring actions are unique) are pre-existing and fixing those goes beyond the scope of this PR.

I'll note that manual cleanup of unique keys introduced a risk of locking the queue for particular action. I.e. if some 3rd party code overrides methods which introduced unique_key cleanup doesn't do this, all further attempts to schedule an action will fail, even if there is no matching pending action. So I'd suggest extending a daily cleanup job or creating a tool resetting unique_keys for actions in terminal states. What do you think?

drewmt:
Thanks for testing this and flagging the cleanup gap. I have added bounded recovery to the existing daily cleanup, with coverage for stale terminal keys, active actions, continuation batches and database failures.

classes/data-stores/ActionScheduler_DBStore.php
```diff
@@ -63,6 +63,57 @@ public function init() {
...
+				ORDER BY action_id
+				LIMIT 1000",
+				self::STATUS_COMPLETE,
+				self::STATUS_FAILED,
+				self::STATUS_CANCELED
+			)
+		);
+
+		if ( false === $released ) {
+			throw new RuntimeException( 'Unable to release stale unique action keys: ' . esc_html( $wpdb->last_error ) );
```
vbelolapotkov, line 95: I don't think throwing here is the best approach, as it'll interrupt the execution of other hook handlers. I suppose the main intent here is to log the failure, so let's do so. In case of temp DB connection failure the next job should pick up cleanup, which I think is acceptable here.

Otherwise, we'd need to at least wrap error message into i18n function, as we do for other messages.

classes/data-stores/ActionScheduler_DBStore.php
```diff
@@ -142,14 +193,18 @@ private function save_action_to_db( ActionScheduler_Action $action, ?DateTime $d
...
 			// phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared -- $insert_sql should be already prepared.
-			$wpdb->query( $insert_sql );
-			$action_id = $wpdb->insert_id;
+			$query_result = $wpdb->query( $insert_sql );
+			$action_id    = $wpdb->insert_id;
 
-			if ( is_wp_error( $action_id ) ) {
-				throw new \RuntimeException( $action_id->get_error_message() );
+			if ( false === $query_result ) {
+				throw new \RuntimeException( $wpdb->last_error ? $wpdb->last_error : __( 'Database error.', 'action-scheduler' ) );
```
vbelolapotkov, line 207: Now that we have a cleanup method, it's probably a good idea to run particular cleanup in place if the last error matches duplicate key situation. This way if the job creation is blocked by a job in final state still holding unique key we can resolve it in place and self-recover without waiting for cleanup job to run.

vbelolapotkov (commented):
Thanks @drewmt , that's a good progress! There are a couple more things I'd suggest to address before merging this one, see my comments below.

drewmt:
Thanks for the suggestions. Cleanup failures are now logged without interrupting other handlers, and stale terminal keys recover immediately with one guarded insert retry. Added coverage for competing inserts and hook continuation, and resolved the changelog conflict. The affected suites pass in local testing.

vbelolapotkov (approved):
Thanks @drewmt ! I ran some tests locally and it works as expected. I'll go ahead and merge the PR.

Issue #1261, opened 2025-04-10
Enhancements to Scheduled Actions table in WP Admin


kaushikasomaiya:
Discussed in pcShBQ-3cR

We discussed it could be useful for troubleshooting if the table is made more flexible with the following improvements 

- Table lacks controlling number of entries per page 
- Filtering by specific groups could be useful
- Add Action ID info beneath the hook name 
- Sort by scheduled date but also with a possibility of excluding specific statuses

kaushikasomaiya:
We'll skip

> Filtering by specific groups could be useful
> Sort by scheduled date but also with a possibility of excluding specific statuses


This can be closed after the above PRs merge

PR #1359, opened 2026-08-04
Add action IDs to admin list and search

kaushikasomaiya:
## Summary
- Show the action ID below each hook name.
- Allow searching by action ID in both stores.

Addresses part of #1261.

### Before

image

### After

image

### DIFF
--- a/tests/phpunit/ActionScheduler_ListTable_Test.php
+++ b/tests/phpunit/ActionScheduler_ListTable_Test.php
@@ -16,7 +16,7 @@ class ActionScheduler_ListTable_Test extends ActionScheduler_UnitTestCase {
 			)
 		);
 
-		$this->assertStringContainsString( '&lt;script&gt;test&lt;/script&gt;', $output );
+		$this->assertStringContainsString( '&lt;script&gt;test&lt;/script&gt;', $output, 'HTML elements will be escaped' );
 		$this->assertStringContainsString( 'ID: 123', $output );
 		$this->assertStringNotContainsString( '<script>', $output );
 	}
--- a/tests/phpunit/ActionScheduler_ListTable_Test.php
+++ b/tests/phpunit/ActionScheduler_ListTable_Test.php
@@ -1,0 +1,23 @@
+<?php
+
+/**
+ * Tests for the Action Scheduler admin list table.
+ */
+class ActionScheduler_ListTable_Test extends ActionScheduler_UnitTestCase {
+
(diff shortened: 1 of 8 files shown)


tests/phpunit/jobstore/AbstractStoreTest.php
```diff
@@ -115,6 +115,18 @@ public function test_query_actions_by_array_status() {
...
+	public function test_query_actions_search_by_action_id() {
+		$store    = $this->get_store();
+		$schedule = new ActionScheduler_SimpleSchedule( as_get_datetime_object( 'tomorrow' ) );
+
+		$action_id       = $store->save_action( new ActionScheduler_Action( 'search_action_id', array( 'first' ), $schedule ) );
+		$other_action_id = $store->save_action( new ActionScheduler_Action( 'search_action_id', array( 'second' ), $schedule ) );
+		$results         = array_map( 'intval', $store->query_actions( array( 'search' => (string) $action_id ) ) );
+
+		$this->assertContains( $action_id, $results );
+		$this->assertNotContains( $other_action_id, $results );
```
barryhughes, line 127: From a perspective of tests-as-documentation: I wonder if, as part of this test or else a separate test in the same class, we should cover the potential for a **claim ID** to also be matched 🤔

tests/phpunit/ActionScheduler_ListTable_Test.php
```diff
@@ -0,0 +1,23 @@
...
+		$list_table = $reflection->newInstanceWithoutConstructor();
+		$output     = $list_table->column_hook(
+			array(
+				'ID'          => 123,
+				'hook'        => '<script>test</script>',
+				'status_name' => ActionScheduler_Store::STATUS_COMPLETE,
+			)
+		);
+
+		$this->assertStringContainsString( '&lt;script&gt;test&lt;/script&gt;', $output );
```
barryhughes:
```suggestion
		$this->assertStringContainsString( '&lt;script&gt;test&lt;/script&gt;', $output, 'HTML elements will be escaped' );
```

(Optional, just felt it might add additional clarity around the inclusion of the script tag in the assertion.)

barryhughes (commented):
Thanks, this will be useful. Notes:

- It's atypical to include post/entity IDs in WP admin list tables in this way, but given how Action Scheduler is used (and by whom) I think it is the right call.
- I left a couple of comments, but neither are blockers. If you wish to move ahead without making further changes that is fine.

Image

barryhughes (approved):

PR #1360, opened 2026-08-04
Fix screen options on WooCommerce status page

kaushikasomaiya:
## Summary
- Initialize the list table early on the WooCommerce Scheduled Actions tab.
- Show the existing per-page Screen Option there.

Addresses part of #1261.

### Before

image

### After

image

## Backward compatibility

Assessed, and considered acceptable.

* **New public method on an overridable class.** `ActionScheduler_AdminView::maybe_process_wc_admin_ui()` is new. Sites can replace that class through the `action_scheduler_admin_view_class` filter, so a third-party subclass could already declare a method of the same name — which would then silently become the override, or fatal on an incompatible signature. The name is specific enough that a collision is very unlikely, and the method is additive: no existing method, signature, or hook changed.
* **No new capability surface.** `load-woocommerce_page_wc-status` fires only after WooCommerce's own `manage_woocommerce` check for that page. The list table was already constructed on the same page by `woocommerce_admin_status_content_action-scheduler`, so this only moves the work earlier in the request.
* **Behavior change, and an improvement.** Row and bulk actions on the Scheduled Actions tab now run before output starts, so the `wp_safe_redirect()` in `ActionScheduler_Abstract_ListTable::process_actions()` works. Previously it ran after the headers were sent. Nonce checks are unchanged (`check_admin_referer()` for bulk actions, `wp_verify_nonce()` for row actions).
* **No change** to the public `as_*()` functions, to any hook we fire, to the stored data or schema, or to the Tools > Scheduled Actions screen.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 4.1.0 - xxxx-xx-xx =
+* Fix - Show the per-page Screen Option on WooCommerce's Scheduled Actions tab.
 * Use `wp_admin_notice()` to render admin notices, instead of custom HTML.
 * Fix an oversight in the lock implementation (used to rate-limit async request runners) that could result in a database error in unusual cases.
 * Add protections to guard against the risk of object-injection/deserialization attacks when retrieving stored schedule data.
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,8 @@
 *** Changelog ***
 
+= 4.2.0 - xxxx-xx-xx =
+* Fix - Show the per-page Screen Option on WooCommerce's Scheduled Actions tab.
+
 = 4.1.0 - 2026-08-05 =
 * Fix - Correct an oversight in the lock implementation (used to rate-limit async request runners) that could leave a lock permanently stuck and result in database errors in unusual cases.
 * Fix - Correct the behavior of the `wp action-scheduler clean` command so that the `--before` option defaults to 31 days ago.
--- a/classes/ActionScheduler_AdminView.php
+++ b/classes/ActionScheduler_AdminView.php
@@ -53,6 +53,7 @@ class ActionScheduler_AdminView extends ActionScheduler_AdminView_Deprecated {
 		if ( is_admin() && ( ! defined( 'DOING_AJAX' ) || ! DOING_AJAX ) ) {
 
 			if ( class_exists( 'WooCommerce' ) ) {
+				add_action( 'load-woocommerce_page_wc-status', array( $this, 'maybe_process_wc_admin_ui' ) );
 				add_action( 'woocommerce_admin_status_content_action-scheduler', array( $this, 'render_admin_ui' ) );
 				add_action( 'woocommerce_system_status_report', array( $this, 'system_status_report' ) );
 				add_filter( 'woocommerce_admin_status_tabs', array( $this, 'register_system_status_tab' ) );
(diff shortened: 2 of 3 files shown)


classes/ActionScheduler_AdminView.php
```diff
@@ -110,6 +111,17 @@ public function process_admin_ui() {
 		$this->get_list_table();
 	}
 
+	/**
+	 * Initialize the list table early on WooCommerce's Scheduled Actions tab.
+	 */
+	public function maybe_process_wc_admin_ui() {
+		$tab = isset( $_GET['tab'] ) ? sanitize_key( wp_unslash( $_GET['tab'] ) ) : ''; // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```
barryhughes, line 118: This works, but I'd still be inclined to swap `$_GET` with `$_REQUEST`, because that is what [WooCommerce uses in the same context](https://github.com/woocommerce/woocommerce/blob/11.0.0/plugins/woocommerce/includes/admin/views/html-admin-page-status.php#L10) and so, and I realize this is a bit hypothetical, if something like new ajax based navigation was introduced to the Woo UI (that leaned on sending `'tab'` via post data) we'd fall out of sync.

tests/phpunit/ActionScheduler_AdminView_Test.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+/**
+ * Tests for the Action Scheduler admin view.
+ */
+class ActionScheduler_AdminView_Test extends ActionScheduler_UnitTestCase {
```
barryhughes, line 6: - To make this run in CI, we need to add `<file>./phpunit/ActionScheduler_AdminView_Test.php</file>` to the [test suite definition](https://github.com/woocommerce/action-scheduler/blob/trunk/tests/phpunit.xml.dist#L24-L36) (or make some other suitable update)...
- ...That said, I'm not strongly sold on these tests and though it's a little unusual to decline new tests, I tend to think they are unnecessary in this case (so we could remove them).

changelog.txt
```diff
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 4.1.0 - xxxx-xx-xx =
+* Fix - Show the per-page Screen Option on WooCommerce's Scheduled Actions tab.
```
barryhughes, line 4: 4.1.0 shipped so let's move this to a new 4.2.0 block.

barryhughes (changes requested):
LGTM! Tests well:

- Existing screenshots match what I see when testing locally.
- No errors (in the browser, or in the error log) in either location.
- I confirmed the screen options take effect in both locations, and persist in a location-specific way (that is, removing the schedule column in one screen does not automatically remove it in the other).
- Left a couple of pretty minor notes/observations/things we should address before moving forward.

_Sidebar: it's well outside the scope of this PR, but I rather wish the special handling for WooCommerce was handled by WooCommerce._

kaushikasomaiya:
> it's well outside the scope of this PR, but I rather wish the special handling for WooCommerce was handled by WooCommerce.

I agree. Codex and I decided to add it here because there is already some Woo related handling done in AS. Perhaps all of that can be moved to Woo.

barryhughes (approved):
Thanks, I implemented some of those additional changes and confirmed things still look good in both locations (**Tools ‣ Scheduled Actions** and **WooCommerce ‣ Status ‣ Scheduled Actions** ... and of course in the latter case we get the UI improvement).

We can merge once checks pass.

PR #1367, opened 2026-08-12
Add/issue PAY-798 - AGENTS.md backward compatibility guardrail

rtio:

Adds an AGENTS.md Backward Compatibility section, adapted from [Core's guardrail](https://github.com/woocommerce/woocommerce/blob/trunk/AGENTS.md#backward-compatibility) to this library's surface: the public `as_*()` functions, the `ActionScheduler_Store`/`Logger`/`Lock`/`Abstract_QueueRunner` base classes and the `ActionScheduler_Schedule` interface, the `action_scheduler_*_class` swap filters, the stored tables and serialized schedules, and the multi-version embedding constraint. Docs only; no code or changelog.

Linear: PAY-798

Read the Backward Compatibility section in AGENTS.md and confirm the named functions, classes, filters, and tables are accurate for this repo.


rtio:
@barryhughes this adds a Backward Compatibility section to AGENTS.md as part of the Woo release-reliability guardrail rollout (Linear PAY-798). It's AI-drafted and grounded in the repo's real surface - the `as_*()` API, the store/logger/runner base classes, the `action_scheduler_*_class` swap filters, and the stored tables - but Chronos own Action Scheduler, so please review and take it from here. Adjust anything that reads wrong.

AGENTS.md
```diff
@@ -114,6 +114,48 @@ Multiple plugins, all embedding different versions of Action Scheduler, may be a
 
 Plugins can also load Action Scheduler in atypical ways, occasionally leaving an older-than-expected version initialized. We cannot control this, but should keep it in mind when introducing new APIs.
 
+## Backward Compatibility
+
+Any change to a **public or externally exposed** class, interface, function, or method signature is **high-risk** and **must state its backward-compatibility impact in the PR description**. Note that the parameters added to the published functions over time (`$unique`, `$priority`) are all optional and trailing: that is the shape a safe change to a published signature takes.
```
barryhughes: Adding:

- A note that optional trailing params added to top-level functions are only safe if we provide a means for consumers to test for them †
- A note about compatibility as it relates to named args ††

† Consider here the `$unique` param that most of our current `as_<schedule|enqueue>_*()` functions accept. If the version of Action Scheduler resolved at runtime does not support this, then a call like `as_enqueue_async_function( 'hook', [], '', true )` will not 'hard fail', but we'd get a different category of error (because unless other safeguards exist, which would anyway diminish the value of the `$unique` param, this could easily lead to an unexpected flood of actions, the exact opposite of what is desired).

†† Following nicely from that earlier point, if a caller takes advantage of named arg syntax and does `as_enqueue_async_function( 'hook', unique: true )` but, at runtime, an older version of Action Scheduler is loaded, then we would indeed encounter a hard failure.

```suggestion
Any change to a **public or externally exposed** class, interface, function, or method signature is **high-risk** and **must state its backward-compatibility impact in the PR description**. Note that the parameters added to the published functions over time (`$unique`, `$priority`) are all optional and trailing: that is considered to be a safe change, but should be supported by an update to `as_supports()` (and its list of supported features), as detailed elsewhere in the document.

Parameter names should never be changed for public methods or functions, as this will cause errors if consumers take advantage of named argument syntax.
```

AGENTS.md
```diff
@@ -114,6 +114,48 @@ Multiple plugins, all embedding different versions of Action Scheduler, may be a
...
+- The `action_scheduler_*_class` filters, which let a site swap any of those implementations for its own — so code that assumes the concrete class is one of ours breaks those sites.
+- The stored data: the custom tables (`{$wpdb->prefix}actionscheduler_actions`, `_claims`, `_groups`, `_logs`), the `ActionScheduler_Store::STATUS_*` values, the `serialize()`d schedule objects, and the legacy `scheduled-action` post type unmigrated sites still hold rows in.
+
+**Multiple versions coexist and the newest one wins.** Per the Bootstrapping section above, whichever embedded copy is most recent is the one that loads, so a change ships to host plugins that never adopted it and calls arrive from callers written against much older versions. What you add is not guaranteed to exist for a consumer whose site loaded an older copy, and `as_supports()` is the escape hatch for that. What you remove can still be called by an older host plugin, and there is no escape hatch for that at all, which makes removal effectively permanent.
```
barryhughes:
```suggestion
**Multiple versions coexist and the newest one wins.** Per the Bootstrapping section above, the general expectation is that whichever embedded copy is most recent is the one that loads, so a change ships to host plugins that never adopted it and calls arrive from callers written against much older versions. In exceptional and unusual cases, an older version of Action Scheduler may unexpectedly be loaded at runtime and so what you add is not always guaranteed to exist: `as_supports()` is the escape hatch for that. What you remove can still be called by an older host plugin, and there is no escape hatch for that at all, which makes removal effectively permanent.
```

AGENTS.md
```diff
@@ -114,6 +114,48 @@ Multiple plugins, all embedding different versions of Action Scheduler, may be a
...
+
+Adding a required method to a contract external code implements is backward-incompatible — a new `abstract public function` on one of those base classes fatals every third-party subclass on load, and `ActionScheduler_Schedule` carries the same risk as a true interface. Removing a required method breaks existing implementers too, leaving them carrying a now-dead method. Prefer a non-breaking alternative: add a concrete default implementation on the base class rather than an abstract declaration, or introduce a separate new interface.
```
barryhughes: Hopefully I'm not misunderstanding this one, but I'm provisionally thinking of some adjustments:

- "`ActionScheduler_Schedule` carries the same risk as a true interface" → well, that's because it **is** a [true interface](https://github.com/woocommerce/action-scheduler/blob/4.1.0/classes/schedules/ActionScheduler_Schedule.php#L6). Removing for clarity.
- I don't understand how _removing_ a required method (I assume abstract method) breaks existing concrete implementations. That is, if interface `foo` defines abstract function `bar`, but then `bar` is removed in a future release, that won't introduce a hard breakage for implementations that happen to still provide a method `bar`. There is some potential for soft/silent breakages, though, so I'll reword to acknowledge that.
- Also, re-ordering and breaking the two parts (addition and removal) into separate paragraphs for increased clarity. So, instead of _«Risk of adding» «Risk of removing» «Safe way to add»_ we have _«Risk of adding» «Safe way to add» (new para) «Risk of removing»._

```suggestion
Adding a required method to a contract external code implements is backward-incompatible — a new `abstract public function` on one of those base classes fatals every third-party subclass on load. Prefer a non-breaking alternative: add a concrete default implementation on the base class rather than an abstract declaration, or introduce a separate new interface. 

Removing a required method from an interface is potentially safe, but should be done with care as it may leave implementations carrying a now-dead method (and those implementations may rely on it being invoked in order to function in a satisfactory manner). .
```

AGENTS.md
```diff
@@ -114,6 +114,48 @@ Multiple plugins, all embedding different versions of Action Scheduler, may be a
...
+
+**Deprecate, don't rename.** Never rename or remove an existing public symbol (class, method, constant, hook, option key) in place. Mark the old one `@deprecated`, add the replacement alongside it, and keep both working through a deprecation window. ./deprecated/ and the `_deprecated_function()` calls that name the version they were introduced in are how this repo has always done it; extend that pattern rather than deleting.
```
barryhughes:
```suggestion
**Deprecate, don't rename.** Never rename or remove an existing public symbol (class, method, constant, hook, option key) in place. Mark the old one `@deprecated`, add the replacement alongside it, and keep both working through a deprecation window. Deprecated code can be relocated to ./deprecated/ (taking steps to ensure it is still loaded), and `_deprecated_function()` or appropriate equivalent such as `_deprecated_hook()` should be used.
```

AGENTS.md
```diff
@@ -114,6 +114,48 @@ Multiple plugins, all embedding different versions of Action Scheduler, may be a
...
+
+> This rule exists because WooCommerce 10.9.0 was reverted on WP Cloud: a required method added to a published interface fataled every older WooCommerce Stripe Gateway version that implemented it. The same failure mode applies to every contract this library publishes, with a wider blast radius — we do not know which plugins extend us.
```
barryhughes, line 140: It does strike me that this note (the blockquote that gives the context from WC 10.9) is a little superfluous, but perhaps it partly exists for human reviewers of this guidance?

AGENTS.md
```diff
@@ -114,6 +114,48 @@ Multiple plugins, all embedding different versions of Action Scheduler, may be a
...
+
+Treat a symbol as **externally exposed** when it is implemented or consumed outside this repository. When in doubt, assume it is exposed and state the BC impact — this is a library embedded in other people's plugins, so every consumer is code we cannot see, and a fatal introduced here fatals whatever host plugin bundled us.
+
+**As a producer of public API.** The surface third parties consume:
+- The top-level `as_*()` functions in ./functions.php — the declared public API.
+- The `ActionScheduler_*` base classes third parties extend (`ActionScheduler_Store`, `ActionScheduler_Logger`, `ActionScheduler_Lock`, `ActionScheduler_Abstract_QueueRunner`) and the `ActionScheduler_Schedule` interface.
+- The `action_scheduler_*_class` filters, which let a site swap any of those implementations for its own — so code that assumes the concrete class is one of ours breaks those sites.
+- The stored data: the custom tables (`{$wpdb->prefix}actionscheduler_actions`, `_claims`, `_groups`, `_logs`), the `ActionScheduler_Store::STATUS_*` values, the `serialize()`d schedule objects, and the legacy `scheduled-action` post type unmigrated sites still hold rows in.
```
barryhughes, line 129: Adding bullet re WP CLI commands (as part of the public API surface).

```suggestion
- The stored data: the custom tables (`{$wpdb->prefix}actionscheduler_actions`, `_claims`, `_groups`, `_logs`), the `ActionScheduler_Store::STATUS_*` values, the `serialize()`d schedule objects, and the legacy `scheduled-action` post type unmigrated sites still hold rows in.
- WP CLI commands and parameters. Individual sites may have shell scripts or similar which depend on the existence of these commands, or on the existence of various parameters. Such commands are not always invoked interactively but may run via the system scheduler or similar.
```

barryhughes (approved):
Great, thank you. Putting forward a couple of changes: I'll go ahead and merge/accept each of them, but of course don't hesitate to come back to me if you have doubts about any of them—we can easily revert and revise as needed.

PR #1374, opened 2026-08-31
Top up AGENTS.md backward-compatibility guardrails

rtio:

Adds the backward-compatibility blocks Core's AGENTS.md gained in woocommerce/woocommerce#67305 that this copy is missing: hook-data validation, overridable-method contracts, script handle contracts, database migrations. Docs only.

Read the diff. Confirm the named classes, handles, and migration code exist in this repo.

Refs ARC-1859, PAY-798


AGENTS.md
```diff
@@ -145,6 +145,14 @@ WordPress exposes more contracts than class and function signatures. The followi
...
 
+**Never trust data that flows through hooks.** Keep hook callback parameters untyped and validate or coerce the value before passing it to strictly typed code, since any callback can receive a value another one produced. And when firing a filter, validate the final return value before using it, since any callback in the chain can return the wrong thing.
+
+**Overridable classes are contracts too, including which internal methods get called.** The base classes listed above exist to be subclassed, and the `action_scheduler_*_class` filters mean a third-party subclass can be the live implementation on any site. Every public or protected method on them is a contract: changing its signature or removing it breaks those subclasses. Adding a fast path or skip that avoids calling an overridable method silently disables those overrides even though no signature changed: the subclass's code simply stops running. When optimizing such a class, ensure overridable methods are still invoked on every code path, or treat the change as breaking.
```
barryhughes: - I'm a little unsure what we mean by internal, it seems open to interpretation. If we mean methods marked `@internal`, then doesn't the opening statement of this paragraph undermine their use? If we mean `private`, I think we can be much clearer.
- Should we introduce guidance on the addition of new methods (since adding new methods also carries a risk, given that such a method can collide with a pre-existing method from third party subclasses, etc)? I'm not necessarily advocating for this, to be clear, but it seems worth consideration. A ruling or some guidance on this point would also cover cases involving visibility widening (ie, our ability to make a private method protected or public).

AGENTS.md
```diff
@@ -145,6 +145,14 @@ WordPress exposes more contracts than class and function signatures. The followi
...
+
+**Registered script and style handles are public contracts.** This library registers none today; keep it that way. The moment a handle ships, third-party code can enqueue it or list it as a dependency, and its name is frozen. To rename one later, register the legacy handle as an alias that depends on the new one (the same pattern WordPress core uses for `jquery` → `jquery-core`) rather than removing it; do not register the same file under both handles, or pages with mixed consumers will load it twice.
```
barryhughes: > This library registers none today; keep it that way.

Should this be softened? If we were to introduce a UI to let site operators tune performance characteristics (which is not implausible) we may wish to take advantage of JS and CSS, and we'd probably wish to follow WordPress conventions.

barryhughes: I wonder too if we should clearly recommend here that any handles be prefixed with something appropriate (like `as-`), to reduce the chances of a collision with a third party handle (which might be the sort of thing that would motivate a name change)?

AGENTS.md
```diff
@@ -145,6 +145,14 @@ WordPress exposes more contracts than class and function signatures. The followi
...
+
+**Schema updates are one-shot and must stay compatible one version back.** Table changes ship by incrementing `$schema_version` in the schema classes; each version's update runs once and is marked complete, so never amend a shipped schema version - increment it again. A host plugin can be downgraded, leaving an older embedded copy running against tables a newer copy already upgraded, so every schema change must stay readable one version back. Additive columns and keys survive that; dropping or repurposing a column fatals or corrupts under code that still expects it.
```
barryhughes: I wonder if _'one version back'_ is sufficient. Perhaps it should just be removed? Action Scheduler is pretty atypical (it can be installed as a regular plugin, but that's not how it is intended to be used), and we cannot control which parent plugins will be present or what version they ship (there's a greater chance of multi-version leaps forward and backwards than might be typical for most other plugins).

AGENTS.md
rtio: "Internal" meant call sites - the base class no longer calling one of its own overridable methods, so a subclass override silently stops running. It clearly doesn't carry that on its own, so I reworded the lead to "including which of their own methods they call" and dropped the word.

Added the new-method guidance. A name a subclass already declares fatals on an incompatible signature or visibility, and on a compatible one the subclass method silently becomes the override. Widening private to protected/public joins a previously invisible name to the contract, so the text treats it as adding a new method.

AGENTS.md
rtio: Softened. The rule I wanted is that the first handle opens a new compatibility surface, so it ships as a deliberate decision, not as a side effect of another change - the text now says that. Added the prefix recommendation too; I went with the full `action-scheduler-` since two characters is itself the kind of name that collides.

AGENTS.md
rtio: Removed the bound. "One version back" assumes an update cadence this library doesn't have - the live copy is whichever registered version wins after a host plugin is downgraded or deactivated, and that can jump years. The text now says there's no bounded window: older shipped code must stay able to read the tables, which in practice means additive-only.

barryhughes (approved):

PR #1377, opened 2026-09-10
Skip runtime scheduling side effects when bootstrapped from a plugin's uninstall.php

vbelolapotkov:
## Task

> Fix ACTSCH-109

[ACTSCH-109: Skip runtime scheduling side effects when Action Scheduler is bootstrapped from a plugin's uninstall.php](https://linear.app/a8c/issue/ACTSCH-109/skip-runtime-scheduling-side-effects-when-action-scheduler-is)

## Why

WordPress deactivates a plugin before running its `uninstall.php`, so `plugins_loaded` has already fired without it. A plugin that bundles Action Scheduler and wants to cancel its own actions on uninstall therefore has to bootstrap Action Scheduler itself from there (the theme-support branch at the bottom of `action-scheduler.php` makes a plain `require_once` sufficient). WooCommerce does exactly this in woocommerce/woocommerce#68527.

`ActionScheduler::init()` assumed every initialization is the start of a normal request, and created new state on a site that is about to lose Action Scheduler:

1. `ActionScheduler_RecurringActionScheduler::init()` hooks `schedule_recurring_scheduler_hook()` on `action_scheduler_init`, which `init()` fires synchronously, inserting a brand new pending `action_scheduler_run_recurring_actions_schedule_hook` action.
2. `ActionScheduler_QueueRunner::init()` calls `wp_schedule_event()` for `action_scheduler_run_queue` if it is missing.
3. `ActionScheduler_QueueRunner::init()` hooks the async request dispatcher on `shutdown`. This one is worse than a leftover row: WordPress deletes the host plugin's files later in the same request, so by `shutdown` the class files are gone and the autoloader fails with `Uncaught Error: Class "ActionScheduler_Lock" not found`.

The host plugin could only undo 1 by reaching into internals (the hook name is a `private const`), and works around 3 today by calling `ActionScheduler::runner()->unhook_dispatch_async_request()` right after bootstrapping.

## What changed

* `ActionScheduler::is_uninstalling()` reports whether this copy was initialized from a plugin's uninstall routine. `ActionScheduler_RecurringActionScheduler::init()` returns early when it is true; `ActionScheduler_QueueRunner::init()` registers the `cron_schedules` filter and then returns early, before the legacy-cron cleanup, `wp_schedule_event()`, the `action_scheduler_run_queue` callback, the async dispatcher and the cleaner hooks.
* `as_supports( 'uninstall_bootstrap' )` so host plugins can detect the behavior, plus `docs/api.md`, a new "Usage in `uninstall.php`" section in `docs/usage.md`, and a changelog entry.

The flag is **captured in `ActionScheduler::init()`**, not evaluated on demand. `defined( 'WP_UNINSTALL_PLUGIN' ) && did_action( 'plugins_loaded' ) && ! doing_action( 'plugins_loaded' )` is only specific to a late bootstrap at that moment: on a normal request the store and runner are initialized on `init`, where the same test also passes, so a test suite that defines the constant for the whole process (WooCommerce's PHPUnit bootstrap does) would lose its runtime. `ActionScheduler_Test::test_constant_defined_after_initialization_is_not_an_uninstall` pins that.

## Backward compatibility

**Public surface added, none changed or removed.**

* `ActionScheduler::is_uninstalling()` is a new public static method on `ActionScheduler`. Additive. The class is abstract with only static methods and is not designed to be subclassed, so the "an existing subclass may already declare this name" risk does not apply in practice; the name is also distinctive.
* `as_supports( 'uninstall_bootstrap' )` is a new entry in the existing feature list. Additive, and the documented way for consumers to detect this at runtime given older copies may win the `plugins_loaded` race.
* No hook was removed, renamed, reordered or retimed. Every hook `ActionScheduler_QueueRunner::init()` and `ActionScheduler_RecurringActionScheduler::init()` registered before is still registered on a normal request, including `apply_filters( 'action_scheduler_run_schedule', ... )` and the `cron_schedules` filter, and all overridable methods are still invoked on that path. On a normal request, `init()` registers everything in the same order as before.
* Behavior only diverges inside the uninstall bootstrap, which is a path that previously ended in a fatal. No schema or stored-data change.
* No global state is newly assumed: the check reads a constant and hook state only.
* Multisite: no site or network state is read or written by the change, so behavior is identical.

## Design decisions recorded

* **Detection is a proxy predicate.** Anything that defines `WP_UNINSTALL_PLUGIN` outside core's `uninstall_plugin()` combined with a late bootstrap (theme-support branch, a plugin requiring `action-scheduler.php` after `plugins_loaded`) would flip the flag on a site that keeps Action Scheduler, and the failure is a silently missing runtime for that request. Narrowing the check to "the plugin being uninstalled is the one this copy lives in" (comparing `dirname( self::$plugin_file )` with `WP_PLUGIN_DIR . '/' . dirname( WP_UNINSTALL_PLUGIN )`) was considered and rejected: it trades a failure mode that needs third-party misuse of a core-internal constant for one triggered by legitimate layouts (symlinked checkouts, case-insensitive filesystems), where the fatal would silently return.
* **The detection expression has no test for its TRUE branch.** The new tests force the flag by reflection, because `ActionScheduler::init()` runs once per process and the constant cannot be undefined. The existing test pins the snapshot semantics (defining the constant after initialization must not flip the flag); a true-branch test would need `@runInSeparateProcess` or a seam extracted purely for testing, and was left out on purpose.
* **The `run()` callback and the cleaner hooks are skipped during an uninstall.** A queue run spawned by WP Cron happens in a separate request, where the flag is false. In the uninstall request itself, a cron callback that deletes plugins can be followed by a due `action_scheduler_run_queue` event, and `run()` would then load classes from the deleted plugin folder.
* **The pre-existing `action_scheduler_run_queue` cron event is not cleared.** Clearing it would act destructively on a predicate that can, in the configurations above, be a false positive. Left as is.

## Out of scope

WordPress defines `WP_UNINSTALL_PLUGIN` only for the `uninstall.php` branch of `uninstall_plugin()`. A plugin that cleans up from a `register_uninstall_hook()` callback instead has its main plugin file `include_once`'d with no constant and no other signal available at `init()` time, so that path cannot be detected. `docs/usage.md` says so and recommends `uninstall.php`.

## Testing

* `WP_TESTS_DIR=… ./vendor/bin/phpunit -c tests/phpunit.xml.dist` on the branch rebased onto current trunk: 1866 tests, 0 failures (10 pre-existing PHPUnit deprecation warnings). Same result for the directory-scan form used by `composer run test`.
* `composer run lint` / `composer run lint-branch` - clean.
* Verified red/green: reverting either guard or registering `run()` above the guard fails the new uninstall tests, and making `is_uninstalling()` lazy fails the snapshot test.
* **Verified on a Jurassic Ninja site** (WP 7.1, no other Action Scheduler provider) with a test plugin that bundles Action Scheduler and cancels its own action from `uninstall.php`, deleted through wp-admin's Plugins screen (non-AJAX `delete-selected` path), with the cron event, housekeeping row, transient and async lock cleared beforehand:
  * **trunk:** `PHP Fatal error: Uncaught Error: Class "ActionScheduler_Lock" not found … do_action('shutdown')` (the trace from the issue), the cron event recreated, a new pending housekeeping action inserted.
  * **this branch:** "The selected plugin has been deleted", `debug.log` with 0 fatal lines and `is_uninstalling=true` logged from `uninstall.php`, no cron event, no housekeeping row, no async lock, the plugin's own action canceled. On a normal admin request with the plugin active, `is_uninstalling()` is false and the shutdown dispatcher and cron event are set up as before.
* Not tested under multisite locally; CI covers it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01243f39i6CNmQjwKHmZBHPf


vbelolapotkov:
@barryhughes I need your approval to merge this. The solution was tested on JN site, and it works as expected.

classes/ActionScheduler_QueueRunner.php
```diff
@@ -78,6 +78,23 @@ public function __construct( ?ActionScheduler_Store $store = null, ?ActionSchedu
...
+		}
+
+		// The registrations above are kept during an uninstall: they create no state, and if WP Cron still
+		// spawns a queue run for this site, it should be served. Neither a cron event nor an async request
+		// may be left behind, though: WordPress deletes the host plugin's files later in the same request,
+		// so the shutdown callback would fatal on the missing classes and the cron event would outlive the
+		// code that serves it.
+		if ( ActionScheduler::is_uninstalling() ) {
+			return;
+		}
```
barryhughes: I'm not sure about this, though I think we're *partly* looking at a pre-existing problem.
  
- The comment seems to be a little misleading: if WP Cron *spawns* a queue run, then that is going to run in a separate request, not this one, and at that point `is_uninstalling()` will be false anyway.
- On the other hand, suppose we are *inside* the spawned WP Cron request, and some code runs that deletes plugins, including whichever one hosts Action Scheduler. Since we've moved `add_action( self::WP_CRON_HOOK, array( self::instance(), 'run' ) )` above this check, that callback will still run and we could imagine fatals as now non-existent class files are called.
  
Trunk registers the same callback, so this isn't a regression. But since this guard now exists, should the `WP_CRON_HOOK` registration move back below it?

docs/usage.md
```diff
@@ -96,6 +96,37 @@ Action Scheduler will later initialize itself on `'init'` with priority `1`.  Ac
 
 When using Action Scheduler in themes, it's important to note that if Action Scheduler has been registered by a plugin, then the latest version registered by a plugin will be used, rather than the version included in the theme. This is because of the version dependency handling code using `'plugins_loaded'` since version 1.0.
 
+### Usage in `uninstall.php`
+
+WordPress deactivates a plugin before running its `uninstall.php`, so `'plugins_loaded'` has already fired without it. A plugin that needs to cancel its own scheduled actions on uninstall therefore has to load Action Scheduler itself, which a plain `require_once` is enough to do:
+
+```php
+<?php
+// Another still-active plugin may already have loaded Action Scheduler for this request.
```
barryhughes: A bit of a nit, but is the following more accurate?

```suggestion
// Another still-active plugin may already have loaded Action Scheduler for this request,
// or it may have been bootstrapped by an inactive plugin that was also being uninstalled.
```

Or just:

```suggestion
// Action Scheduler may already have been loaded.
```

(On the basis that there could be >1 inactive plugins being uninstalled, where >1 from that set both include Action Scheduler.)

docs/usage.md
```diff
@@ -96,6 +96,37 @@ Action Scheduler will later initialize itself on `'init'` with priority `1`.  Ac
...
+$as_already_loaded = class_exists( 'ActionScheduler', false );
+
+require_once plugin_dir_path( __FILE__ ) . '/libraries/action-scheduler/action-scheduler.php';
```
barryhughes, line 109: It conveys the right idea and it aligns with some of our other docs, but this is also a little atypical. 

```suggestion
require_once plugin_dir_path( __FILE__ ) . '/vendor/woocommerce/action-scheduler/action-scheduler.php';
```

The above is probably the more common formulation in practice ([here is a real life example](https://plugins.svn.wordpress.org/seo-by-rank-math/tags/1.0.278/vendor/woocommerce/action-scheduler/) of Action Scheduler's location within a host plugin). I think we're copying from WooCommerce with the current example, but that is the outlier.

Again, though, no need to change. Changing it might even cause confusion unless we update other parts of the docs.

barryhughes (commented):
Left some comments inc one minor nit, just musing about whether we should address that nit (re WP_CRON_HOOK) _now_ 🤔

classes/ActionScheduler_QueueRunner.php
barryhughes: Addressed via 7f45b03.

docs/usage.md
barryhughes: Addressed via d7bb30c.

docs/usage.md
barryhughes: No need to address now.

barryhughes (approved):

PR #1073, opened 2024-06-20
Use WP 6.4 Admin Notices 

fiechtor:
Resolves #1072 .

Updates all notice prints to `wp_admin_notice` calls. Also updates the WordPress minimum version to due the introduction version of the `wp_admin_notice` function

PR #1381, opened 2026-09-14
Fix placeholders and escaping in admin notices

barryhughes:
We are addressing problems introduced in #1073, around placeholders and escaping.

- **Past-due actions notice:** the `printf()` call was dropped, so the notice showed `%1$d` instead of the number of actions, and its link pointed to `%2$s`. The placeholders are now replaced again.
- **Row action notice** (after an action is run or canceled, or fails to process): the hook name was no longer passed through `esc_html()`. It is escaped again.

Each fix has a regression test, and each test fails without its fix.


barryhughes (commented):
Used the following snippet to trigger and then confirm that the past-due notice renders correctly (ie, placeholders are replaced):

```php
add_filter( 'action_scheduler_check_pastdue_actions', fn () => true );
add_filter( 'pre_transient_action_scheduler_last_pastdue_actions_check', fn () => [] );
add_action( 'action_scheduler_init', fn () => as_schedule_single_action( time() - DAY_IN_SECONDS, 'past-due' ) );
```

Image

The above screnshot also shows a row action notice rendering as expected (we added back an `esc_html()` wrapper around the hook name).

